PR #100, opened 2012-10-11
Added  --files option to cli 

dignifiedquire:
and the ability to pass adapters via `"#ADAPTER"`.
For example you can do now

``` bash
$ testacular start --files "#JASMINE,file1.js,file2.js" testacular.conf.js
```


rpkilby:
Was there any discussion on why this was reverted in 1c172ea94aa943161a73cc084027a36eb3dcc90b? I don't see a relevant PR or other comments.

PR #380, opened 2013-03-08
autoStart option

### LINKED ISSUE #375: Add hook to delay `start()`
For Ext.js, ember, and perhaps other applications, the test runner could use a direct API for delaying the test suite from running.  Here's my current work-around:

``` js
window.__testacular__.loaded = function(){};
someApp.on('init', function() {
  window.__testacular__.start();
});
```

And then add it to the files array in the test config. Haven't had a chance to think about the API for that yet.

ryanflorence:
closes #375

Can prevent tests from being run automatically
for scenarios that require an application to
do some initialization work before starting tests

### DIFF
--- a/docs/config/01-configuration-file.md
+++ b/docs/config/01-configuration-file.md
@@ -92,6 +92,15 @@ it again. After three attempts to capture it, Testacular will give up.
 See [config/files] for more information.
 
 
+## autoStart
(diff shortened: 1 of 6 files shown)


Dignifiedquire:
Closing this for now. If there is still a need and interest for this feature please open a new issue.

PR #778, opened 2013-10-06
Feature proposal/discussion

yhwh:
While I was running my tests on browserstack the browsers were throwing an error randomly

    Opera 15.0.1147 (Windows Vista) ERROR
	    Uncaught SyntaxError: Unexpected end of input
	    at /home/travis/build/yhwh/stimuli/dist/stimuli.js:4075

It was really weird because it could happen randomly on any browser but always at the same line number, then I discovered that the issue was coming from the connect gzip compression, I don't think it's a bug it's just too slow. In fact the connection was aborting without any error after one http chunk was sent over the network. So it means that serving big js file can cause random issues.

I think this options should be set to true by default.

But I would even go further, in fact I don't think it's a good idea to let node handling the gzip compression since it's single threaded which can potentially blow the performances.
I have done some experiment on my own branch https://github.com/yhwh/karma/commit/f83faf4cea245b1ab7176111685fd198dd45d387
I removed the the hostname/port options couple and created 4 others instead 
* clientHost the host used by the browser (and the runner)
* clientPort the port used by the browser (and the runner)
* serverPort the host the server will listen to
* serverHost the port the server will listen to
By doing this I was able to put nginx between my browserstack tunnel and my karma server (on travis).
https://github.com/yhwh/stimuli/blob/master/.nginx/nginx.conf
I decided to serve the static files as well through and handle the gzip compression through nginx.
I just had to switch to xhr-polling only, because even if nginx support websockets it doesn't work with old webkits implementation so phantomjs and safari 5 were not able to connect. It seems to have improved the overall performances (but it could be due to the resultBuffer of xhr-polling).
What I know for sure is that using nginx as several advantage:
* it should scale better when finely tuned
* it's extremely flexible you can create dynamic rules for proxying etc... 

What do you think ?

### DIFF
--- a/docs/config/01-configuration-file.md
+++ b/docs/config/01-configuration-file.md
@@ -112,6 +112,14 @@ it again and, after three attempts to capture it, Karma will give up.
 **Description:**  Enable or disable colors in the output (reporters and logs).
 
 
+## disableCompression
+**Type:** Boolean
+
+**Default:**   `false`
+
+**Description:**  Enable or disable gzip compression on the server.
+
+
 ## exclude
 **Type:** Array
 
--- a/lib/config.js
+++ b/lib/config.js
@@ -279,6 +279,7 @@ var Config = function() {
     args: []
   };
   this.browserDisconnectTimeout = 2000;
+  this.disableCompression = false;
 
   // TODO(vojta): remove in 0.10
   this.junitReporter = {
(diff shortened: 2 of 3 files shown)


Dignifiedquire:
Closing this for now as this seems to be resolved by the commit @vojtajina mentioned

PR #834, opened 2013-11-20
feat(preprocessors): allow multiple file patterns matching

revolunet:
allows to specify multiple file patterns in the preprocessors config so you can add exclusions rules. useful if you want for example coverage to match all src/*_/_.js files except those ending with *.spec.js.

Should help for https://github.com/karma-runner/karma-coverage/issues/13

BREAKING CHANGE: the preprocessors config order has been reversed.

Before :

``` js
preprocessors: {
  '**/*.coffee': ['coffee'],
  '**/*.js': ['coverage']
}
```

After :

``` js
preprocessors: {
  'coffee': '**/*.coffee',
  'coverage': ['**/*.js', '!**/*.spec.js']
}
```

### DIFF
--- a/docs/config/04-preprocessors.md
+++ b/docs/config/04-preprocessors.md
@@ -4,9 +4,8 @@ of the configuration file:
 
 ```javascript
 preprocessors: {
-  '**/*.coffee': ['coffee'],
-  '**/*.tea': ['coffee'],
-  '**/*.html': ['html2js']
+  'coffee': '**/*.coffee',
+  'html2js': '**/*.html'
 },
 ```
 Note the multiple expressions referencing the "coffee" preprocessor, as a preprocessor can be listed more than once,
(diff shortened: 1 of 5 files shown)


johnpapa:
I think changing this to a glob pattern would make it tons simpler.

lrodziewicz:
Unfortunately this ("src/**/!(*.spec).js") glob only works on sh/bash shells and fail on zsh. Round brackets are treated as attributes. See http://zsh.sourceforge.net/FAQ/zshfaq02.html
I'm not sure if it's an minmach issue or does it open again a subject of changing the syntax of preprocessors configuration.

mprentic-harris:
Something needs to be addressed similar to this. At minimum the current behavior with multiple matching via minimatch should be documented in the README.md.

allenhwkim:
+1 for more advanced inclusion/exclusion

e-globa:
Helps for me:

` preprocessors: {
      '<%= tmp_dir %>/src/**/*.js': ['babel'],
      '<%= tmp_dir %>/src/**/!(*spec).js': ['coverage']
    },`

rmorrise:
+1 for this feature. Putting the '!' syntax in the middle of a pattern string to specify exclusions is not intuitive at all.

PR #948, opened 2014-03-12
feat: allow frameworks to add preprocessor files

taylorhakes:
This is necessary for the commonjs plugin I am using. I need to parse a npm module from the files array and add it to the preprocessors object. Since the keys are saved, it is not possible to add preprocessor patterns currently.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -27,10 +27,6 @@ var log = logger.create();
 var start = function(injector, config, launcher, globalEmitter, preprocess, fileList, webServer,
     capturedBrowsers, socketServer, executor, done) {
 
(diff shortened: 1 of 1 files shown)


vojtajina:
@taylorhakes Agree that modifying the config (eg. `files` property) by plugins and hoping that it will happen before somebody reads it is bad. We should probably have a clear API for plugins to alter the config file.

The order of preprocessors is clearly defined by the user, in the config file. I don't see any problem with that.

vojtajina:
#merge

vojtajina:
It's in canary, in NPM as `0.12.28-beta-f65c864`.

PR #1053, opened 2014-05-01
feat(server): wait for frameworks to load via q.all() before starting server

### LINKED ISSUE #851: Support for asynchronous plugins
**Use case:**

Looking to test an AngularJS module (angular-ui/ui-router) against several versions of AngularJS, with and without specific plugins (angular-animate).

So, rather than making a big ugly grunt task to manage all of this, I thought it would be nice to write a karma plugin to simplify this.

The goal is this: interface with the bower API to install modules before a test suite is run (by injecting config.bower into the framework factory)

The problem I run into is that I can't seem to find a way to have a plugin treated as asynchronous, so that it can wait for installation to complete.

---

This would be a super-awesome plugin (I think), and it would depend on a super-awesome feature of Karma.

I could look at implementing this, but I am not aware of all of the finer points of the karma-runner, so it would be really nice if someone could give this some consideration

caitp:
The API is to simply return a promise from a factory function, or use a promise as a value. The server will wait for the collection of frameworks to resolve before carrying on starting the server.

This is useful for when setup requires some asynchronous startup.

Currently, any Promise api which is interoperable with q may be used.

In addition to this feature, this CL also adds a set of callbacks to the server, via an object for the 3rd parameter. This is useful for test purposes, and enables the test case to run faster, by not waiting for the server to shut down via the second parameter.

---

There are some design decisions here which you might not like, such as using `mockery` rather than creating test fixtures outside of the test, or adding an object of callbacks to the server opener to help with testing it (it might be nice to have for other reasons, like bonus logging points or something). And of course, the main changes might be restructured in different ways too. So let me know if there are any issues with it @vojtajina, I suspect you were thinking the CL would be a bit smaller and change fewer things, but testing it seemed like a good thing to do.

Closes #851

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -8,6 +8,7 @@ JSHINT_BROWSER =
 JSHINT_NODE =
   node: true,
   strict: false
+  latedef: "nofunc"
 
 module.exports = (grunt) ->
 
--- a/lib/server.js
+++ b/lib/server.js
@@ -23,230 +23,237 @@ var processWrapper = new EmitterWrapper(process);
 
 var log = logger.create();
 
+var q = require('q');
+
+function noop() {}
 
 var start = function(injector, config, launcher, globalEmitter, preprocess, fileList, webServer,
-    capturedBrowsers, socketServer, executor, done) {
+    capturedBrowsers, socketServer, executor, done, callbacks) {
 
-  config.frameworks.forEach(function(framework) {
-    injector.get('framework:' + framework);
-  });
+  q.all(config.frameworks.map(function(framework) {
+    return injector.get('framework:' + framework);
+  })).then(startServer, startServer);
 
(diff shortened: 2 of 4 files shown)


PxlBuzzard:
Did you have enough time to review @vojtajina? I kid, but it would be nice to look into this again.

caitp:
I don't think @vojtajina is going to be reviewing this --- and I can't actually even remember writing this. What was this for again?

PxlBuzzard:
@caitp Your [original issue](https://github.com/karma-runner/karma/issues/851) was asking for async plugin support, and the discussion ended with a workaround and this PR.

Dignifiedquire:
@caitp @PxlBuzzard is this still a feature that is needed? 

I'm going to close this PR for now, as doesn't merge cleanly anymore. If you want to update it feel free to reopen it and I will review it in a more timely fashion.

PR #1056, opened 2014-05-03
Only include original line info in stack trace when using source map

devongovett:
I don't think it's useful at all to list the compiled line number/column in stack traces, especially since it's displayed first before the original line number/column info from the source map.  Especially when using something like browserify, you get these horribly useless paths to look at.  For example:

```
at /var/folders/_p/kll4xqzn0hn9znq6607nr14m0000gn/T/a5fdf7b6ec86e5a5a2a6f44ed6234a6791e3a679.browserify:203:0 <- /Users/devongovett/projects/editor/node_modules/chai/lib/chai/assertion.js:110:0
```

This commit removes the compiled line information from stack traces and only displays the original line information.  The output of the above example would just be:

```
at /Users/devongovett/projects/editor/node_modules/chai/lib/chai/assertion.js:110:0
```

### DIFF
--- a/lib/reporter.js
+++ b/lib/reporter.js
@@ -48,8 +48,7 @@ var createErrorFormatter = function(basePath, emitter, SourceMapConsumer) {
         var smc = new SourceMapConsumer(file.sourceMap);
         var original = smc.originalPositionFor({line: line, column: column});
 
-        return util.format('%s:%d:%d <- %s:%d:%d', path, line, column, original.source,
-            original.line, original.column);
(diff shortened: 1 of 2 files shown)


Dignifiedquire:
@devongovett I think the suggestion @aymericbeaumet made is a good one, would you be up to update this PR to the latest master and add a debug log message for the full details? That way people who need debugging details can see it and everyone else doesn't need to care about it. (Yes I know this is a late response :( )

Dignifiedquire:
Closing for now. Feel free to open a new PR updated to the latest master.

PR #1084, opened 2014-06-02
replace simple reassign with  _.merge

huangyingjie:
The following code will override some properties that is an Object, like config.client.
```js
this.set = function(newConfig) {
    Object.keys(newConfig).forEach(function(key) {
      config[key] = newConfig[key];
    });
};
```
If someone want to  pass an custom property to client, He has to also pass all the properties:useIframe,args,and captureConsole, or they will be lost.
So I suggest replacing this with following:
```js
   this.set = function(newConfig) {
     config = _.merge(config, newConfig);
   };
```


Dignifiedquire:
Closing due to inactivity.

Issue #974, opened 2014-03-22
Improve watching


vojtajina:
There are still many issues with file watching (both using events and polling).
See https://github.com/paulmillr/chokidar/issues/105, https://github.com/paulmillr/chokidar/issues/9, https://github.com/paulmillr/chokidar/issues/35, https://github.com/karma-runner/karma/issues/960 and many others...

I looked into chokidar and I think we might workaround some of these issues by never watching a file, but rather parent directory.

THE CHANGE:
Right now, for patterns such as `src/*.js` we have to watch the `src/` directory and so we do (see https://github.com/karma-runner/karma/blob/65b5f1b37fedfb5bd1c43c900c7520b5d0c26017/lib/watcher.js#L16-L39). But for patterns such as `src/main.js`, we watch that file. Instead, we should watch the parent directory (`src/`).

ISSUES (when watching a single file):
- if file does not exist, it's not watched (eg. when later added, it's ignored) (https://github.com/paulmillr/chokidar/issues/105)
- with OS events, it even throws ENOENT error
- if file is removed, chokidar won't notice on OSX (https://github.com/paulmillr/chokidar/issues/9) and when later added, chokidar won't notice either

This workaround will fix these issues. At least on Mac OSX. We need to test this on Linux and Windows, too. Both with `usePolling: true` and `usePolling: false`.

It still won't work if the parent directory does not exist (we should show a warning), but I think it would be a significant improvement. This has to be fixed in chokidar, but I think the change is not trivial.

@paulmillr what do you think?

Dignifiedquire:
Closing as most of the work was done in chokdiar and we have upgraded

PR #1123, opened 2014-07-02
refactor(watcher): Use gaze for FS watching

schmod:
There are several issues surrounding file-watching (see #974), and it makes increasingly little sense to try to address all tricky edge cases within Karma, when there are existing libraries that tackle the "tough issues" for us.

This PR replaces chokidar with Gaze, which is a purpose-built globbing file watcher.  [Gaze](https://github.com/shama/gaze) has some of its own issues (particularly related to ST3), but largely seems to be more stable and reliable than Karma's old approach.  Gaze is fairly widely used, and notbaly provides the underlying functionality for [grunt-contrib-watch](https://github.com/gruntjs/grunt-contrib-watch).

Additionally, we now watch files according to their relative paths (rather than the absolute path that config.js computes for us).  This seems to prevent Gaze from inadvertently traversing filesystem boundaries (which hurts performance, and breaks the detection of new files).

Most unit tests in watcher.js were removed, because Gaze implements its own test suite, which doesn't need to be replicated here.  If there's any functionality that you think should be tested, please leave a comment.

When this PR is merged, it may make sense to start cleaning up file_list.js, and removing globbing hacks wherever we can, and delegate that functionality out to an external library.

### DIFF
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -1,107 +1,71 @@
-var chokidar = require('chokidar');
-var mm = require('minimatch');
-
+var Gaze = require('gaze').Gaze;
+var globule = require('globule');
 var helper = require('./helper');
 var log = require('./logger').create('watcher');
-
-var DIR_SEP = require('path').sep;
-
-// Get parent folder, that be watched (does not contain any special globbing character)
-var baseDirFromPattern = function(pattern) {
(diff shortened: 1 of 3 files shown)


Dignifiedquire:
chokidar has come a long way and with the upgrade to 1.0 I think it's in a good place for now, so unless big issues come up I don't think we will switch filewatchers any time soon

PR #1138, opened 2014-07-18
fix(client): Fix stringify serializing objects

jakzale:
Stringify uses the name of the constructor to serialize objects.
When the object was created using Object.create(null) or has a property
constructor this may throw an Error


vojtajina:
#merge

vojtajina:
Thank you @jakzale

PR #1146, opened 2014-07-29
fix(server): launch browsers when file_list is ready

jpommerening:
After weeks of intermittent timeouts starting PhantomJS I finally found out what's wrong–

We're having a lot of files (well, not *that* many, maybe a hundred) and we're using `grunt-karma` to trigger single-run tests with PhantomJS in a continuous integration environment.

Correct me if I'm wrong, but it looks like the web server is not ready to reply to the browser's connect request until the file list is finished. Since PhantomJS' startup is really quick, it sometimes tries to connect before `fileList.refresh()` completed.

After applying this patch, the problem seems to disappear… what do you think?

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -34,7 +34,7 @@ var start = function(injector, config, launcher, globalEmitter, preprocess, file
   var filesPromise = fileList.refresh();
 
   if (config.autoWatch) {
-    filesPromise.then(function() {
+    filesPromise = filesPromise.then(function() {
       injector.invoke(watcher.watch);
     }, function() {
       injector.invoke(watcher.watch);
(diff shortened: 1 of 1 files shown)


vojtajina:
Short story: I agree we need to fix this issue, but I don't think this PR is the right solution. I'm not sure what is the cause of the issue is.

Long story: Here is how Karma works:
It starts the webserver, once the webserver is listening on given port, start browsers.
A browser opens `/` url, the webserver serves `static/client.html` for that url, which establishes socket.io connection - the browser is captured. Responding to this request does not wait for resolving the files.

When a test run starts, all browsers refresh an iframe with actual tests. This iframe points to `/context.html`. Webserver does delay serving of `/context.html`, until the `file_list` promise is resolved (it needs updated list of files in order to generate the `/context.html` response).

**Thus, capturing of the browser is not delayed until the files are resolved.**

I'm not sure what the actual cause of your problem is. Sounds like the Karma process is too busy resolving all the files (and possibly preprocessing them) and the webserver does not get scheduled to handle the requests from browser. Can you try it with some GUI browser, eg. Chrome and check out the timeline - how long does it take to serve the initial `/` request, including its resources (such as `static/karma.js`, socket.io, etc)? Then the browser sends a message (through socket.io), "register", how long does that message take? Also, how many files do you have (I would like to reproduce the issue? 
 
Let's figure out what the cause is and then we can fix it properly.

Can you open the timeline in the browser
This PR will make capturing browsers much slower (waiting for all files to be resolved and then starting browsers; it takes significant time for the browser to actually open the url). So I don't think this PR is a good solution.

jpommerening:
Hey Vojta,

Thanks for that excellent summary. I think I understand the whole process a lot better now!

I'll follow your suggestions and post the results next week. Also, I'll try to build a small repo to trigger the problem.

Have a nice weekend!

Dignifiedquire:
@jpommerening any updates on this?

Dignifiedquire:
Closing for now.

PR #1150, opened 2014-08-04
fix(server): exit process when browser is not captured in captureTimeout...

binyroo:
exit process when browser is not captured in captureTimeout


vojtajina:
@binyroo I'm not sure I understand the issue. Can you describe more? What is the problem you are solving?

Dignifiedquire:
Closing due to inactivity.

PR #1160, opened 2014-08-15
fix(web-server) Fix 581 https support for karma server

JayBerlin:
This PR implements support for running the karma proxy server as https so that karma is able to proxy a secure site and handle secure cookies.  It really just some formalization of [danielabar's hack](https://github.com/danielabar/hack-karma-https) as I need this to support some internal e2e testing using karma.  

To us this, add this to the karma.conf.js file:

    protocol: 'https'

By default, this will create a self-signed certificate that lasts 1 day to be used by the karma server.  If you want to supply your own certificate, add the following property to the karma.conf.js file:

    cert: {
       key: 'name_of_key.pem_file',
       cert: 'name_of_cert.pem_file'
    }

Caveats:

* If you are using self-signed certificates, you need to modify the karma-chrome-launcher module by adding these additional options when launching:

    '--ignore-certificate-errors',
    '--ignore-urlfetcher-cert-requests',

Otherwise, the browser will hang waiting for a confirmation that it is okay to visit a site with a self-signed certificate.  I have not found a way to do this for Firefox or other browsers.

* I found that I had to use version 0.8.3 of the modified [node-http-proxy](https://github.com/jayv/node-http-proxy) in place of the default http-proxy in order to run e2e tests with embedded frames.  This version properly handles location header rewriting so that access to proxied content within an iframe works properly.  This was an issue for us when redirects were occurring.

Not sure if this is worth pulling into karma for everyone as there are some pretty big caveats, but wanted to get a PR in place as we are using this now.


Dignifiedquire:
First, sorry for the late feedback. This looks like a good idea to me and it seems there is more need for this, see the newer #1400 and #1401. So I would really like to get this in, could you rebase onto the latest master and squash the latest commit?

Dignifiedquire:
Closing due to inactivity

mystro-bryce:
@JayBerlin have you had any luck polishing off @Dignifiedquire's final requests? I'd love to see this in Karma proper ASAP.

JayBerlin:
I have been away from this code for the last several months.  I will see if I can resurrect it and get an environment set up to merge this in, but will not be able to get to it for several days.  Anyone else is more than welcome to do so in the interim.

PR #1168, opened 2014-08-24
feat(server): config.capturePort

bazilio91:
Had a problem tunnelling karma server through ngrok (ngrok http port is 80).


Dignifiedquire:
In general I think this is a valid use case and it's okay to add this option, as long as the default is sensible so most people don't have to think about it. 

Could you address the feedback @maksimr gave, add a test for this and rebase onto the latest master please?

Dignifiedquire:
Closing due to inactivity.

PR #1212, opened 2014-10-14
pass cliOptions to config file

madhums:
This would enable the user to have access to cli options. Currently there is no way you can send extra params to config, this would atleast help in setting custom configs based on params.


vojtajina:
You can parse CLI arguments yourself, in your config file. It's not as nice, I know. Unless there is a good use case (that many people neeed), I don't think it's worthy making the API more complex...

Please re-open if you think there is a good use case.

Issue #1186, opened 2014-09-11
Karma should not read IP address from the environment, or it should say it does


ghost:
Karma silently reads the IP address to connect from then environment variable named IP.

It should not rely on such a generic name, and output something in the console saying that the hostname has been overriden by an environment variable.

Is there an interest in submitting a PR for that ?

Anyhow, thanks for this wonderful project :-)

PR #1214, opened 2014-10-15
fix(server): use the config.hostname to open server binding

paulRbr:
__Hello!__

I've had the issue to not be able to bind the karma webserver on the local ip in a cloud instance. 
Realising that Karma was not listenning on the configurated hostname, I am submitting a PR for that.

Let me know if this is a constraint that comes from somewhere I didn't get.

Thanks for this wonderful project!

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -6,6 +6,7 @@ exports.VERSION = pkg.version;
 
 exports.DEFAULT_PORT = process.env.PORT || 9876;
 exports.DEFAULT_HOSTNAME = process.env.IP || 'localhost';
+exports.DEFAULT_LISTEN_IP = process.env.KARMA_IP || '0.0.0.0';
 
 // log levels
 exports.LOG_DISABLE = 'OFF';
--- a/lib/server.js
+++ b/lib/server.js
@@ -45,7 +45,7 @@ var start = function(injector, config, launcher, globalEmitter, preprocess, file
     if (e.code === 'EADDRINUSE') {
       log.warn('Port %d in use', config.port);
       config.port++;
-      webServer.listen(config.port);
+      webServer.listen(config.port, config.listen);
     } else {
       throw e;
     }
@@ -61,7 +61,7 @@ var start = function(injector, config, launcher, globalEmitter, preprocess, file
   // Some browsers did not get captured.
   var singleRunBrowserNotCaptured = false;
 
-  webServer.listen(config.port, function() {
+  webServer.listen(config.port, config.listen, function() {
     log.info('Karma v%s server started at http://%s:%s%s', constant.VERSION, config.hostname,
         config.port, config.urlRoot);
 
--- a/docs/config/01-configuration-file.md
+++ b/docs/config/01-configuration-file.md
@@ -213,6 +213,14 @@ Additional information can be found in [plugins].
 **Description:** Hostname to be used when capturing browsers.
 
 
+## listen
+**Type:** String
+
+**Default:** `0.0.0.0` or `KARMA_IP` env variable value if set
+
+**Description:** The address where the web server will be listening.
+
+
 ## logLevel
 **Type:** Constant
 
--- a/lib/config.js
+++ b/lib/config.js
@@ -205,6 +205,7 @@ var Config = function() {
   this.frameworks = [];
   this.port = constant.DEFAULT_PORT;
   this.hostname = constant.DEFAULT_HOSTNAME;
+  this.listen = constant.DEFAULT_LISTEN_IP;
   this.basePath = '';
   this.files = [];
   this.exclude = [];


vojtajina:
@popox why do you need to listen only on a specific interface? What is the use case for it?

@dlecan Yes, it is not a breaking change, but it does make configuration of Karma more complex so I wanna make sure it is worthy.

popox:
@vojtajina my use case is on a public paas instance (openshift online) where I _don't have any rights to bind a specific port on the local ip `0.0.0.0`_ but I have the rights to listen on the public IP available for my instance.

This helped me to be able to run an integration server on a cloud provider and be able to launch Karma to orchestrate my JS tests.

maksimr:
Good use case from https://github.com/karma-runner/karma/pull/1311

> Current karma version doesn't specify the address and in this case node binds to IPv4 address only. So there is no way to bind it to v6 interface. With added option it's now possible to specify v6 address.

maksimr:
What if we will not add new `setting` and simple adopt `hostname` for this purpose.
If hostname the same as default hostname then listen on `0.0.0.0`

/сс @zzo @aymericbeaumet @Dignifiedquire

zzo:
Sgtm is there a pr for just that change somewhere I missed?

maksimr:
@paulRbr what do you think about using `hostname` for this purprose?

@zzo no, only this PR

paulRbr:
@maksimr If we don't add a new configuration option, we'll be more likely to break someone's usage of Karma.

**I.e.** I am thinking of people configuring the `hostname` to let's say `whatever.example.com` which is pointing to the ip `88.77.66.55`. Without noticing they would limit the server listen host to that specific ip address. Therefore, in case they have scripts / dependencies connecting to karma through the local interface `127.0.0.1` where karma is running, they would break and receive a:
 `Failed to connect to 127.0.0.1 port 9876: Connection refused`

If we go for that, they would need to launch karma with the env variable `IP=88.77.66.55` to be feature equivalent.

/cc @zzo @vojtajina

maksimr:
@paulRbr Thanks!

Yeah...  `hostname` is not best choice.
Will be cool add information about `listen` to documentation

paulRbr:
@maksimr PR updated with doc and rebased

lib/server.js
```diff
@@ -45,7 +45,11 @@ var start = function(injector, config, launcher, globalEmitter, preprocess, file
     if (e.code === 'EADDRINUSE') {
       log.warn('Port %d in use', config.port);
       config.port++;
-      webServer.listen(config.port);
+      if (config.listen) {
+        webServer.listen(config.port, config.listen);
+      } else {
+        webServer.listen(config.ports);
```
just-boris: looks like there is should be `config.port`

paulRbr: good catch, thanks

lib/constants.js
```diff
@@ -6,6 +6,7 @@ exports.VERSION = pkg.version;
 
 exports.DEFAULT_PORT = process.env.PORT || 9876;
 exports.DEFAULT_HOSTNAME = process.env.IP || 'localhost';
+exports.DEFAULT_LISTEN_IP = process.env.IP || '0.0.0.0';
```
paulRbr: I don't really like the magical env variable as discussed in #1186, should we just leave `0.0.0.0` as default value and not read the `IP` env variable?

Dignifiedquire: How about we use `KARMA_IP`, I think env variables like that should always be scoped.

Dignifiedquire: Also we need documentation on that option if we add it i

Dignifiedquire:
Except for the inline comment I think this is a good solution, and as soon as this is addressed I think we can merge this.

paulRbr:
@Dignifiedquire done taking your inline comment into account.

Dignifiedquire:
That was quick :) looks good so far, the last thing I'm worried about is that there is no test for this, but the `server.spec.coffee` is a the moment pretty empty :( But if you feel up to it would be great if you add the first test into there

lib/server.js
```diff
@@ -45,7 +45,11 @@ var start = function(injector, config, launcher, globalEmitter, preprocess, file
     if (e.code === 'EADDRINUSE') {
       log.warn('Port %d in use', config.port);
       config.port++;
-      webServer.listen(config.port);
+      if (config.listen) {
```
maksimr: I think now we can remove `if (config.lister) {` and simple write `webServer.lister(config.port, config.lister)` and `webServer.lister(config.port,config.lister, startCallback)`

Thanks!

j0tunn:
Hey guys, what about this PR?

zzo:
Looks good to me - are we waiting for Maksim's comments to be addressed?

Dignifiedquire:
This needs rebasing and updating onto the latest master. Also there should be test for this, in the config.spec for reading the env variable and in the server.spec for it to be used ( there are mow tests in server.spec)

arikon:
@paulRbr Could you rebase and write some tests, please?

j0tunn:
Are there any chances to merge this PR?

Dignifiedquire:
Closing due to inactivity

PR #1232, opened 2014-11-03
feat(reporter): cache SourceMapConsumer

panrafal:
Error formatting is very slow on big files. This makes it a lot faster.


minznerjosh:
@vojtajina I'm not familiar with the Karma release process. Is there anywhere I can read up on how that works? Really, I'm trying to figure out when this will get released, as TDDing is starting to get pretty annoying for my team as our app grows (due to the long times it takes for karma to output errors.)

Thanks!

maksimr:
@minznerjosh now this changes allowed in `canary` branch/release. I think this changes should be moved to master branch because how said @vojtajina it's `refactoring` not new feature.

\cc @zzo

minznerjosh:
@maksimr Getting this released in master as soon as possible would be really helpful to my team. Thank you all!

\cc @vojtajina @zzo

PR #1234, opened 2014-11-04
feat(cli): Better CLI args validation

srawlins:
Start adding better CLI args validation (#603). Validate log-level to start. This doesn't close the issue.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -38,7 +38,16 @@ var processArgs = function (argv, options, fs, path) {
   }
 
   if (helper.isString(options.logLevel)) {
-    options.logLevel = constant['LOG_' + options.logLevel.toUpperCase()] || constant.LOG_DISABLE
+    var logConstant = constant['LOG_' + options.logLevel.toUpperCase()]
+    if (helper.isDefined(logConstant)) {
+      options.logLevel = logConstant
+    } else {
+      console.error('Log level must be one of disable, error, warn, info, or debug.')
+      process.exit(1)
+    }
+  } else if (helper.isDefined(options.logLevel)) {
+    console.error('Log level must be one of disable, error, warn, info, or debug.')
+    process.exit(1)
   }
 
   if (helper.isString(options.singleRun)) {
--- a/test/unit/cli.spec.coffee
+++ b/test/unit/cli.spec.coffee
@@ -8,6 +8,8 @@ describe 'cli', ->
   constant = require '../../lib/constants'
   path = require 'path'
   mocks = require 'mocks'
+  loadFile = mocks.loadFile
+  mockery = m = e = null
 
   fsMock = mocks.fs.create
     cwd:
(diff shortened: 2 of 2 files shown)


lib/cli.js
```diff
@@ -33,6 +33,9 @@ var processArgs = function(argv, options, fs, path) {
 
   if (helper.isString(options.logLevel)) {
     options.logLevel = constant['LOG_' + options.logLevel.toUpperCase()] || constant.LOG_DISABLE;
+  } else if (helper.isDefined(options.logLevel)) {
```
srawlins: @vojtajina It looks like the current behavior, on line 35, is to default unknown options to `LOG_DISABLE` so that if users currently specify `NO` or `OFF` or `FALSE` or `NONE`, etc. then it will just disable. You think we should instead fail if it is unknown?

vojtajina: yes.

I think it might be better to just show a warning when the user misconfigures `--log-level`. It won't affect Karma. This is different from failing to instantiate a preprocessor or launcher (we should fail in that case, because it might pass the tests - a browser that would fail didn't start).

What do you think?

Dignifiedquire:
Yes please. @srawlins could you please do what @vojtajina and just print the warning and not exit. Other than that I'm happy to get this in.

Dignifiedquire:
@srawlins ping

srawlins:
Eek, sorry. I'm on vacation for 2 more weeks. Back in late June.

srawlins:
OK changed behavior to exit on unrecognized log levels, and added a test. Ready for review again. Sorry for the absurd delay :worried:

lib/cli.js
```diff
@@ -38,7 +38,16 @@ var processArgs = function (argv, options, fs, path) {
...
+    var logConstant = constant['LOG_' + options.logLevel.toUpperCase()];
+    if (helper.isDefined(logConstant)) {
+      options.logLevel = logConstant;
+    } else {
+      console.error('Log level must be one of disable, error, warn, info, or debug.');
+      process.exit(1);
+    }
+  } else if (helper.isDefined(options.logLevel)) {
+    console.error('Log level must be one of disable, error, warn, info, or debug.');
+    process.exit(1);
```
Dignifiedquire: This could be simpler, I think something like

```js
var logLevel = options.logLevel

if (helper.isString(options.logLevel)) {
  logLevel = constant['LOG_' + options.logLevel.toUpperCase()]
}

if (!helper.isDefined(logLevel)) {
  console.error('Log level must be one of disable, error, warn, info, or debug.');
  process.exit(1);
}
```

srawlins: Actually, if `logLevel` is not defined (i.e. `options.logLevel` is not defined), this is valid. It just means the user did not pass a `--log-level` argument.

srawlins:
I fixed the lint warnings and there is still a Travis failure that I don't think is related to this PR because it's sporadic. Travis builds [#1600](https://travis-ci.org/karma-runner/karma/builds/71168245) and [#1601](https://travis-ci.org/karma-runner/karma/builds/71289473). It passes for me locally.

Dignifiedquire:
Thanks, looks good to me :octocat:

PR #1236, opened 2014-11-06
feat(logger): Log errors with console.error (defaults to stderr).

### LINKED ISSUE #1011: karma should write errors to stderr and return an exit code
When an error occurs, karma uses `log.error()` to print the error. However it writes to `stdout` instead of `stderr`. Moreover, no error exist code is returned.

srawlins:
Log errors, by default, to a new jog4js appender, the std_appender. This
new appender logs ERROR level events with console.error, and all other
events with console.log.

This std_appender is largely based on the console appender.

Closes #1011

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -1,6 +1,8 @@
 var fs = require('fs');
+var log4js = require('log4js');
 
 var pkg = JSON.parse(fs.readFileSync(__dirname + '/../package.json').toString());
+var stdAppender = __dirname + "/std_appender.js";
 
 exports.VERSION = pkg.version;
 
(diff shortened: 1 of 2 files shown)


srawlins:
This is so bizarre. I swear I wrote this PR because log4js had rejected a similar PR, but now I can't find any such thing :P I should open an issue with them, in case they _would_ accept a PR.

vojtajina:
When outputting to console, I would expect logging errors with `console.error`, right? I assume Node will write into stderr instead of stdout.

If we can't get it into log4js I'm happy to accept this PR, but sounds like adding lot of code into Karma that does not belong here. Can you send me a link to your PR/issue with log4js, I can try to help you arguing;-)

Dignifiedquire:
I'm going to close this, not because this functionality isn't needed, but I will review the log situation and probably move away from log4js anyway, so we don't need all this additional logic inside karma (as it should be in the logger in the first place). Ref #458

PR #1239, opened 2014-11-07
feat: Fail on launcher-, reporter-, or preprocessor-load errors.

### LINKED ISSUE #855: Throw an error on not found browsers
Currently, it looks like if a browser isn't able to be loaded, karma just warns you and happily proceeds with the browsers that could be loaded.

https://github.com/karma-runner/karma/blob/master/lib/launcher.js#L41-L46

In a CI environment, it would be really, really easy to overlook that all of the sudden one of your configured browsers isn't working anymore, but since no errors are thrown the build is still happy.

I'm proposing that instead of continuing on, encountering a browser that couldn't be loaded should be treated as a fatal error - any loaded browsers should be cleaned up, and the process should exit.

srawlins:
Fail out of karma if a launcher, reporter, or preprocessor fails to load, after attempting to load each of them, and reporting errors for each load error.

### Update

This PR is now ready, barring the issue with changing the signature of `createPreprocessor` that I highlight [below](https://github.com/karma-runner/karma/pull/1239#issuecomment-122085301). I did not add any tests, but would be happy to if someone could point me the way there.

### Example

I put the following into a config file:

```json
preprocessors: {
  "**/*.js": ["tea"]
},
reporters: ["progress", "regress"],
browsers: ["Chrome", "Crime"],
```

and started karma:

```
$ ../bin/karma start my.conf.js 
16 07 2015 13:14:36.172:ERROR [reporter]: Cannot load reporter "regress": it is not registered! Perhaps you are missing some plugin?
16 07 2015 13:14:36.182:ERROR [preprocess]: Cannot load preprocessor "tea": it is not registered! Perhaps you are missing some plugin?
16 07 2015 13:14:36.189:INFO [karma]: Karma v0.13.0 server started at http://localhost:9876/
16 07 2015 13:14:36.189:ERROR [launcher]: Cannot load browser "Chrome": it is not registered! Perhaps you are missing some plugin?
16 07 2015 13:14:36.189:ERROR [launcher]: Cannot load browser "Crime": it is not registered! Perhaps you are missing some plugin?
```

then karma promptly exits.

Closes #855

### DIFF
--- a/lib/preprocessor.js
+++ b/lib/preprocessor.js
@@ -18,8 +18,9 @@ extensions.forEach(function (extension) {
 })
 
 // TODO(vojta): instantiate preprocessors at the start to show warnings immediately
-var createPreprocessor = function (config, basePath, injector) {
-  var alreadyDisplayedWarnings = Object.create(null)
+var createPreprocessor = function(config, basePath, injector, emitter) {
+  var patterns = Object.keys(config);
+  var alreadyDisplayedErrors = Object.create(null);
 
   return function (file, done) {
     var patterns = Object.keys(config)
@@ -45,24 +46,25 @@ var createPreprocessor = function (config, basePath, injector) {
         return done()
       }
 
-      preprocessors.shift()(content, file, nextPreprocessor)
-    }
-    var instantiatePreprocessor = function (name) {
-      if (alreadyDisplayedWarnings[name]) {
-        return
+      preprocessors.shift()(content, file, nextPreprocessor);
+    };
+    var instantiatePreprocessor = function(name) {
+      if (alreadyDisplayedErrors[name]) {
+        return;
       }
 
       try {
         preprocessors.push(injector.get('preprocessor:' + name))
       } catch (e) {
         if (e.message.indexOf('No provider for "preprocessor:' + name + '"') !== -1) {
-          log.warn('Can not load "%s", it is not registered!\n  ' +
-            'Perhaps you are missing some plugin?', name)
+          log.error('Cannot load preprocessor "%s": it is not registered! ' +
+                    'Perhaps you are missing some plugin?', name)
         } else {
-          log.warn('Can not load "%s"!\n  ' + e.stack, name)
+          log.error('Cannot load preprocessor "%s"!\n  ' + e.stack, name)
         }
 
-        alreadyDisplayedWarnings[name] = true
+        emitter.emit('load_error', 'preprocessor', name);
+        alreadyDisplayedErrors[name] = true;
       }
     }
 
@@ -83,10 +85,11 @@ var createPreprocessor = function (config, basePath, injector) {
       if (err) {
         throw err
       }
-      nextPreprocessor(null, thisFileIsBinary ? buffer : buffer.toString())
-    })
-  }
-}
-createPreprocessor.$inject = ['config.preprocessors', 'config.basePath', 'injector']
+      file.sha = sha1(buffer);
+      nextPreprocessor(null, thisFileIsBinary ? buffer : buffer.toString());
+    });
+  };
+};
+createPreprocessor.$inject = ['config.preprocessors', 'config.basePath', 'injector', 'emitter'];
 
 exports.createPreprocessor = createPreprocessor
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -57,13 +57,14 @@ var Launcher = function (emitter, injector) {
         browser = injector.createChild([locals], ['launcher:' + name]).get('launcher:' + name)
       } catch (e) {
         if (e.message.indexOf('No provider for "launcher:' + name + '"') !== -1) {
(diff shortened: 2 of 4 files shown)


vojtajina:
Karma should fail if any plugin can't be instantiated. That is: launcher, preprocessor, reporter.

Btw, I like the idea of collecting all errors and then failing, so that user sees all the errors and not only the first one.

Here are all the places where Karma instantiate plugins and thus a fatal error can happen (in order in which it happens):
1. Loading plugins (loading the DI modules):
https://github.com/karma-runner/karma/blob/6d5330981b13a1df75f1037ded532429cabd77cb/lib/server.js#L310
2. Instantiate reporters
https://github.com/karma-runner/karma/blob/51f1b8832b86c4f556b58381fa183776212cfaa7/lib/reporter.js#L85
3. Instantiating frameworks (inside `start` function):
https://github.com/karma-runner/karma/blob/6d5330981b13a1df75f1037ded532429cabd77cb/lib/server.js#L30
4. Start browsers (after the webserver is ready listening):
https://github.com/karma-runner/karma/blob/6d5330981b13a1df75f1037ded532429cabd77cb/lib/server.js#L69
5. Instantiate preprocessors (this happens lazily, once the preprocessor is needed; we need to change that and instantiate all needed preprocessors ahead, inside `createPreprocessor` factory)
https://github.com/karma-runner/karma/blob/6d5330981b13a1df75f1037ded532429cabd77cb/lib/preprocessor.js#L88

Thus, I would add the check into the [webserver listener](https://github.com/karma-runner/karma/blob/6d5330981b13a1df75f1037ded532429cabd77cb/lib/server.js#L72), after launching browsers (), and fail Karma if any error occurred. At that point, all of the plugins were instantiated, except preprocessors. We need to change preprocessors to instantiate all needed preprocessors ahead (in the `createPreprocessor` factory), instead of lazily.


I'm sorry this is confusing. I think it would be good to refactor Karma to instantiate everything first. Right now, "instantiating things" is kind of tangled together with bootstrapping (this is probably more efficient, but harder to understand). I wish I did it differently;-)

@srawlins Let me know if I can help you more. This is very valuable work you are doing!

Dignifiedquire:
@srawlins any chance for you updating this PR with the suggestions from @vojtajina. I would really appreciate it and I promise a shorter feedback cycle!

srawlins:
Hi Friedel, i'm actually on vacation until June 20. Please feel free to
submit a different PR with my commits if you want this in soon. Otherwise I
can try to finish at the end of June.

Dignifiedquire:
Thanks @srawlins for the quick answer, I'll see where we're at until June the 20th and update this pr if I get to doing it myself before.

Dignifiedquire:
hey @srawlins any news on you getting back?

srawlins:
Hi @Dignifiedquire , I've updated this with @vojtajina's suggestions, and it's working great! There's one problem though: I've changed the signature of `createPreprocessor`, adding `emitter`:

```javascript
var createPreprocessor = function(config, basePath, injector, emitter) {
```

Travis revealed this as a problem, I think specifically with the [karma-browserify](https://github.com/nikku/karma-browserify) which calls `createPreprocessor`. I don't know karma's ecosystem well enough to know how big a problem this is. But I don't see another way around my implementation.

lib/preprocessor.js
```diff
@@ -18,8 +18,9 @@ extensions.forEach(function (extension) {
 })
 
 // TODO(vojta): instantiate preprocessors at the start to show warnings immediately
-var createPreprocessor = function (config, basePath, injector) {
-  var alreadyDisplayedWarnings = Object.create(null)
+var createPreprocessor = function(config, basePath, injector, emitter) {
+  var patterns = Object.keys(config);
+  var alreadyDisplayedErrors = Object.create(null);
```
Dignifiedquire: Please use `{}` instead

lib/preprocessor.js
```diff
@@ -45,24 +46,25 @@ var createPreprocessor = function (config, basePath, injector) {
         return done()
       }
 
-      preprocessors.shift()(content, file, nextPreprocessor)
-    }
-    var instantiatePreprocessor = function (name) {
-      if (alreadyDisplayedWarnings[name]) {
-        return
+      preprocessors.shift()(content, file, nextPreprocessor);
+    };
```
Dignifiedquire: No semicolons ;)

Dignifiedquire:
As far as I understand this di business, there should be no issue changing the signature, as it's only used as a factory function, which receives it's arguments through the injector

srawlins:
Yeah, I think you're right. Unfortunately, karma-browserify calls `createPreprocessor` [directly](https://github.com/nikku/karma-browserify/search?utf8=%E2%9C%93&q=createpreprocessor).

Dignifiedquire:
@srawlins any updates?

srawlins:
Closing to try to work this another way with #1931

Issue #488, opened 2013-04-11
Allow inserting source files in the HEAD / custom context.html


vojtajina:
Based on #430

**Let's allow custom `context.html` and `debug.html`.**

Add two config options:
- contextFile (default __dirname + '../static/context.html')
- debugContextFile (default __dirname + '../static/debug.html')

aiboy:
Added support to add custom context.html / debug.html - in "Support for custom context and/or debug html files"  #1360 

Unlike other pull request this one is little bit cleaner and doesn't break existing unit tests

UnsungHero97:
any update on when this change will be ready to use? I'd like to give this a try. I tried adding in the changes from the PR, but I'm running into problems; it seems like other places need to add in the extra parameter

aiboy:
Sorry, my bad. To busy to finish the job. I will try to make appropriate pull request asap

UnsungHero97:
@aiboy did you get a free moment to make that pull request?

T-vK:
Another good point for custom context.html files is that ExtJs 6 for example wants `<html manifest="">` instead of just `<html>`. It's also a lot of work to change the config of Karma every time you add a new script to your index.html and you can't really use internal script in your index.html file because you couldn't tell Karma how to get it for the context.html. In my opinion Karma should definitely inject into a copy of the index.html.  
...and not being able to set the DOCTYPE may be another thing that could cause problems.  

I'm really looking forward to this being implemented.

rosenfeld:
I believe there are many reasons why allowing custom HTML would be desirable. It gives other projects flexibility to do basically any kind of experimentation with Karma without having to fork the project.

In our specific case, we use oojspec to run unit and integration tests with a single runner. Most integration test runners out there are pretty slow, like Capybara, Selenium or anything using a real browser, headless like phantomjs or not.

OOJSpec doesn't enforce the way people should write code. If someone wants to be able to register events on DOM selectors which are never unregistered because it's not needed in the application, they shouldn't change the way they write their code just to satisfy their test runner limitations. So, one is free to register those events once in a `beforeAll` block and then run a set of tests, which may depend or not on previous tests (they are always run in the specified order by specification) which allows some integration scenarios to be tested more quickly than it would be if the tests had to be written in an independent way and also gives developers more flexibility on how they write code.

[...]

PR #1277, opened 2015-01-02
Support custom context file

krasimir:
Hello,

Karma is a wonderful spec runner. However, it is kinda design to work with AngularJS. I'm at the moment exploring various ways of testing client side code. Some frameworks (or apps) require having `<script>` or HTML tags on the page in order to work. I'm using TODOMVC as a example and I'm trying to test it by using different tools. This PR introduces one more property to the `client` configuration item.
```js
  config.set({

    // base path that will be used to resolve all patterns (eg. files, exclude)
    basePath: '',


    client: {
      contextFile: '/TestRunner.html'
    },

    ...
```
The content of the `TestRunner.html` is the same as the content of `static/context.html`. The only one difference is that I added what I needed there. In my case this was a Backbone app and I placed the templates along with the application's HTML. Here is how my test suite looks like:
```js
describe("Testing TODOMVC", function () {

    var setText = function(text, selector) {
        var input = $(selector || '#new-todo');
        var e = $.Event("keypress");
        e.which = e.keyCode = 13;
        return input.val(text).trigger(e);
    };

    before(function() {
        window.localStorage.removeItem('todos-backbone', '');
        app.todos.reset();
    });

    it("Adding new TODOs", function () {
        setText('TODO A');
        setText('TODO B');
        expect($('#todo-list li').length).to.be.equal(2);
    });

    it("Deleting TODO", function () {
        $('#todo-list li:first-child .destroy').click();
        expect($('#todo-list li').length).to.be.equal(1);
    });

    it("Edit and add TODOs", function () {
        setText('A new TODO');
        $('#todo-list li:first-child').addClass('editing');
        setText('A new TODO', '#todo-list li:first-child .edit').blur();
        expect($('#todo-list li').length).to.be.equal(2);
        expect($('#todo-list li label').eq(0).text()).to.be.equal($('#todo-list li label').eq(1).text())
    });

});
```

Ref: https://github.com/karma-runner/karma/issues/488

### DIFF
--- a/lib/middleware/karma.js
+++ b/lib/middleware/karma.js
@@ -59,7 +59,7 @@ var getXUACompatibleUrl = function(url) {
   return value;
 };
 
-var createKarmaMiddleware = function(filesPromise, serveStaticFile,
+var createKarmaMiddleware = function(filesPromise, serveStaticFile, serveCWDFile,
     /* config.basePath */ basePath,  /* config.urlRoot */ urlRoot, /* config.client */ client) {
 
   return function(request, response, next) {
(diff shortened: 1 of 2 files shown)


Dignifiedquire:
Closing due to inactivity

PR #1285, opened 2015-01-20
feat config: Allow for JSON/object exports in config files.

izaakschroeder:
Instead of only allowing a function in a javascript file, also accept `.json` and `.js` that export objects.

Makes it easier to manipulate karma configuration files with tooling (e.g. yeoman).

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -265,8 +265,8 @@ var parseConfig = function(configFilePath, cliOptions) {
       }
       return process.exit(1);
     }
-    if (!helper.isFunction(configModule)) {
-      log.error('Config file must export a function!\n' + CONFIG_SYNTAX_HELP);
+    if (!helper.isFunction(configModule) && !helper.isObject(configModule)) {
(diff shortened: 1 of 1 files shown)


maksimr:
@izaakschroeder Thanks!

Please write test on this behavior.
Also we should update documentation that karma can accept **json** file

Dignifiedquire:
ping @izaakschroeder

izaakschroeder:
@Dignifiedquire - I thought I tried looking into the tests for this when I made it and was unsure how to approach it. Direction welcome.

Dignifiedquire:
Sure @izaakschroeder take a look at this unit test for the config https://github.com/karma-runner/karma/blob/master/test/unit/config.spec.coffee you should be able to add a new test case in there where you pass in a config which contains a JSON object and make sure it parses as expected.

Dignifiedquire:
@izaakschroeder ping

Dignifiedquire:
@izaakschroeder any news?

Dignifiedquire:
Closing due to inactivity

PR #1287, opened 2015-01-21
chore(travis): Add io.js target

Dignifiedquire:
Now that nvm has got support for io.js we should add it as a target on all karma repositories, and fix any occurring issues. 
This PR is the start of this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: node_js
 node_js:
   - "0.10"
   - "0.8"
+  - "iojs-v1.1.0"
 
 env:
   global:


Dignifiedquire:
Looks like travis has not yet upgraded its installation of nvm https://github.com/travis-ci/travis-ci/issues/3108

Dignifiedquire:
Ref https://github.com/karma-runner/karma-firefox-launcher/issues/27

.travis.yml
```diff
@@ -2,6 +2,7 @@ language: node_js
 node_js:
   - "0.10"
   - "0.8"
+  - "io.js"
```
mzgol: It'll be `iojs`. But maybe it's better to specify concrete major+minor versions as in Node, i.e.:
```yaml
```
node_js:
  - "0.10"
  - "0.8"
  - "iojs-1.0"
```

mzgol: ok, 1.0 might not be the best idea considering that 1.1 is not a major but a minor bump. If any, `iojs-1` would be better.

mzgol:
> Ref karma-runner/karma-firefox-launcher#27

FYI: https://github.com/karma-runner/karma-firefox-launcher/issues/27#issuecomment-73347905

It works in io.js 1.1.0. :)

mzgol:
BTW, Travis has not io.js support so this should be fine!

Dignifiedquire:
Not needed anymore.

Issue #1078, opened 2014-05-28
Export binary


necolas:
The binary export was removed in this commit: c7d46270aca83ecfe78f69fa923bc574c0b5bfdc

But I think this has broken npm-scripts aliasing. Normally you can run the local install of a package from npm-scripts in package.json (because npm looks up the binary in `./node_modules/.bin/`:

```json
{
  "scripts": {
    "test": "karma start"
  }
}
```

But this doesn't work with Karma anymore.

nzakas:
Agree with @necolas here. This is pretty confusing and frustrating. If you are shipping the binary in the `karma` package, why not include an easy way to use it?

peter-mouland:
Hi,  just came across this as i was going to do the same PR.  In the docs it says you can use the full path.

https://karma-runner.github.io/0.12/intro/installation.html
 > # Run Karma:
 > $ ./node_modules/karma/bin/karma start

I dont get why i would want the cli when I would like to simply put `karma start xx.js` into my package.json.  Since the bin is in karma, and it works, could we add the bin into karma's package.json?

I think the route of asking people to also install the cli is making the process a bit convoluted.   The cli doesn't currently have tests and is only used for karma, so it isn't really a separate project. 

#1351 #1322 #1275 ... there are few issue open and closed about this!

just my 2cents.

donaldpipowitch:
Sorry, I don't get it? Just made a pull request myself. That is quite annoying to realize that I wasted my time with this even if it was just some minutes. Installing a different module (karma-cli) is just unexpected and in my experience not idiomatic for a node module.

nzakas:
@maksimr I'm trying to understand your objection, but I'm having a hard time. From our perspective, your stance is confusing because we are just asking for an easier way to run the binary that is already shipped in this package. As #1361 shows, it's basically a one-line change that would make our lives a lot easier. We aren't asking for any change in functionality, nor are we asking for additional files to be included. This is the logical equivalent of a symlink.

This package has 8x the downloads of `karma-cli`, which means the majority of users have no need for `karma-cli`. That, plus the number of issues and PRs related to this issue, seem to indicate a pain point for users. Is there anything we can do to change your mind?

maksimr:
Guys thanks for your feedback.

@donaldpipowitch your PR once again says that we have a *problem* we need to solve.

@nzakas 
[...]

PR #1289, opened 2015-01-23
Added path to binary

deiga:


maksimr:
@deiga Thakns,
But .bin was deleted in favor of karma-cli
duplicates https://github.com/karma-runner/karma/issues/1078

deiga:
Why is `karma-cli` the preferred solution? I wouldn't want to install 2 packages to be able to run `karma` and having `./node_modules/.bin` in my `$PATH` is a lot simpler

maksimr:
@deiga you can read about reason here https://github.com/karma-runner/karma/issues/1078

PR #1296, opened 2015-01-30
test(e2e): make sure config files do not collide

### LINKED ISSUE #1291: Race condition in our E2E tests
@juliemr and I spent a little while hunting this one down yesterday.  The short of it is that [asynchronously writing `karma.conf.js`](https://github.com/karma-runner/karma/blob/master/test/e2e/support/world.coffee#L38) causes the tests to be asynchronous, and as a result two tests can run at the same time.  Specifically, I can consistently cause the following to happen on my OSX and Ubuntu machines:

1. [`proxy.feature`](https://github.com/karma-runner/karma/blob/master/test/e2e/proxy.feature) reaches the `fs.writeFile` line
2. [`reconnecting.feature`](https://github.com/karma-runner/karma/blob/master/test/e2e/reconnecting.feature) reaches the `fs.writeFile` line
3. `reconnecting.feature`'s callback is called
4. `proxy.feature`'s callback is called

The result is that both `reconnecting.feature` and `proxy.feature` read from a corrupted `karma.conf.js`, causing both tests to fail dramatically.  I see three possible solutions:

1. Get cucumber.js to wait for one test to finish before starting the next
2. Make writing a file synchronous
3. Randomize the file name for what is currently `karma.conf.js` so that collisions are unlikely

Solution 1 the best of course.  I dug into how `grunt-cucumberjs` invokes the tests, and basically it's just running cucumber's [command-line script](https://github.com/cucumber/cucumber-js/blob/master/bin/cucumber.js).  I couldn't find a cucumber flag for waiting for asynchronous tests however (I even looked at parts of the [source code](https://github.com/cucumber/cucumber-js/blob/master/lib/cucumber/cli/argument_parser.js#L129)).  So this solution seems unfeasible.

Solution 3 is a terrible idea, since it exposes us to more race conditions in the future*.  So I vote that we make writing the config file synchronous.  It's a little inefficient, but it's the simplest, most future-proof solution.

*Julie ran into an error about a port being in use at one point in time.  It could have been that two tests were both trying to run `karma start` at the same time

sjelin:
Closes #1291

This fix just deals with the specific race condition I was running into, rather than solving the underlying problem that these tests are being executed in parallel where as they should be run in sequence (described [here](https://github.com/karma-runner/karma/issues/1291#issuecomment-71951739)).  I spent a good amount of time trying to fix that underlying problem, but didn't make much progress.

### DIFF
--- a/test/e2e/support/world.coffee
+++ b/test/e2e/support/world.coffee
@@ -1,6 +1,7 @@
 fs = require 'fs'
 vm = require 'vm'
 path = require 'path'
+hasher = require('crypto').createHash
 
 mkdirp = require 'mkdirp'
 _ = require 'lodash'
@@ -29,13 +30,15 @@ World = (callback) ->
   # Generate a configuration file and save it to path.
   @writeConfigFile = (dir, file, done) =>
     mkdirp dir, 0o0755, (err) =>
-      throw new Error(err) if err
+      return done err if err
 
       # Remove dirname from config again
       delete @configFile.__dirname
 
       content = @generateJS @configFile
-      fs.writeFile path.join(dir, file), content, done
+      hash = hasher('md5').update(content + Math.random()).digest 'hex';
+      fs.writeFile path.join(dir, hash + '.' + file), content, (err) ->
+        done err, hash
 
   @generateJS = (config) ->
     _.template @template, {content: JSON.stringify config}
--- a/test/e2e/steps/core_steps.coffee
+++ b/test/e2e/steps/core_steps.coffee
@@ -39,10 +39,10 @@ coreSteps = ->
     callback()
 
   @When /^I (run|start|init) Karma$/, (command, callback) ->
-    @writeConfigFile tmpDir, tmpConfigFile, (err) =>
-      callback.fail new Error(err) if err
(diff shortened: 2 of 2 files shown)


test/e2e/support/world.coffee
```diff
@@ -29,13 +30,15 @@ World = (callback) ->
   # Generate a configuration file and save it to path.
   @writeConfigFile = (dir, file, done) =>
     mkdirp dir, 0o0755, (err) =>
-      throw new Error(err) if err
+      return done err if err
```
juliemr: This looks fishy - what does it do?

sjelin: It calls the callback with the error message.  It's what should have been done the whole time.

juliemr:
LGTM.

juliemr:
#merge

freethejazz:
@juliemr This might need an extra push, I think the runnerbot is broken :(

juliemr:
Yup, looks like the bots are down. Humanity is still needed!

Merged as 794393138d4048dfe6fdc61a3bb8bd75dc939873

PR #1311, opened 2015-02-13
Add option to specify listen address

scf2k:
`listenAddr` (CLI: --listen-addr) option is added which is used in listen() call to bind to specified address. Default value is `0.0.0.0`.

Current karma version doesn't specify the address and in this case node binds to IPv4 address only. So there is no way to bind it to v6 interface. With added option it's now possible to specify v6 address.

### DIFF
--- a/docs/config/01-configuration-file.md
+++ b/docs/config/01-configuration-file.md
@@ -261,6 +261,15 @@ See [plugins] for more information.
 
 **Description:** The port where the web server will be listening.
 
+## listenAddr
+**Type:** String
+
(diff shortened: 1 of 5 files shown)


scf2k:
@maksimr @vojtajina please take a look

just-boris:
Ping, please.
We have IPv6 only-infrastructure and suffer from this caveat.
Now I am using this fork and it works fine.

maksimr:
Close in favor https://github.com/karma-runner/karma/pull/1214

Issue #1295, opened 2015-01-30
nsp reports potentially bad version of connect used by karma


pdehaan:
### Steps to reproduce:
```sh
﻿﻿$ git clone git@github.com:karma-runner/karma.git .
$ npm i
$ npm shrinkwrap --dev
$ nsp audit-shrinkwrap # [sudo] npm i nsp -g
```

### Actual results:
```sh
$ nsp audit-shrinkwrap

Name          Installed   Patched  Vulnerable Dependency
serve-static    1.6.4     >=1.7.2  karma > connect
qs              0.5.6     >= 1.x   karma > grunt-contrib-watch > tiny-lr-fork
qs              0.5.6     >= 1.x   karma > grunt-contrib-watch > tiny-lr-fork
syntax-error    0.0.1    >= 1.1.1  karma > grunt-cucumberjs > cucumber > browserify
```

I think this is the **serve-static** issue in question: https://nodesecurity.io/advisories/serve-static-open-redirect (fixed in serve-static@>=1.7.2) which seems to be fixed in newer versions of **connect@2.28.0** and later.

pdehaan:
And since I'm already posting, here's all the "outdated" modules, per `npm outdated` command:
```sh
$ npm outdated --depth 0 | sort
Package               Current     Wanted  Latest  Location
chai                    1.9.2      1.9.2  1.10.0  chai
chokidar               0.12.6  1.0.0-rc2  0.12.6  chokidar
coffee-script           1.7.1      1.7.1   1.9.0  coffee-script
colors                  0.6.2      0.6.2   1.0.3  colors
connect                2.26.6     2.26.6   3.3.4  connect
glob                   3.2.11     3.2.11   4.3.5  glob
graceful-fs             2.0.3      2.0.3   3.0.5  graceful-fs
grunt-browserify        2.1.4      2.1.4   3.3.0  grunt-browserify
grunt-contrib-jshint   0.10.0     0.10.0  0.11.0  grunt-contrib-jshint
grunt-jscs-checker      0.6.2      0.6.2   0.8.1  grunt-jscs-checker
http-proxy             0.10.4     0.10.4   1.8.1  http-proxy
karma-jasmine           0.1.5      0.1.5   0.3.5  karma-jasmine
LiveScript              1.2.0      1.2.0   1.3.1  LiveScript
load-grunt-tasks        0.6.0      0.6.0   3.1.0  load-grunt-tasks
[...]

PR #1312, opened 2015-02-13
(deps) connect@2.28.0 to pickup serve-static@~1.8.0

shawnzhu:
a fix for #1295 as fix for [serve-static Open Redirect(CVE-2015-1164)](https://nodesecurity.io/advisories/serve-static-open-redirect):

> When using serve-static middleware version < 1.7.2 and it's configured to mount at the root it creates an open redirect on the site.


Dignifiedquire:
Thanks a lot but we will upgrade connect to v3 soon, see #1410.

PR #1322, opened 2015-02-23
Include karma bin in package.json

dazld:


pkozlowski-opensource:
@dazld why? Didn't `karma-cli` work for you?

dazld:
hey, when you run `npm test`, for example, then npm automatically includes the `.bin` folder in its path for the script which is being run. this means you don't need either a global install of karma, nor do you have to reference the full path to the bin, and can do stuff like..

```
..
"scripts": {
    "test": "karma start"
  },
..
```
etc.

dazld:
does that make sense..? I know it's a simple change.

maksimr:
@dazld you can install `karma-cli` locally to the project and then you example above will work.

https://github.com/karma-runner/karma/issues/1078

dazld:
got it, thanks.

peter-mouland:
I was also going to do this PR.  Is there a reason why you want the extra step of having to install another dependency rather than just putting the path in to the package.json?

PR #1329, opened 2015-02-28
return config object when done

glan:
This allows plugins such as karma-coverage to pass data to downstream processes.

Might not be the best use of the config object. Better ideas for returning data are welcome.


glan:
plugins should pass data via some other mechanism.

Issue #1219, opened 2014-10-18
Dynamically add file available in testing inside preprocessor.


ouhouhsami:
Hello, 

In a preprocessor, I process a file (and return it, all works well), but also I create a new one that I need to have available during testing. How can I achieve this, inside a preprocessor.

Best

andersekdahl:
+1

I'm trying to get this working as well. What I'm doing is having both a framework and a preprocessor, and in the framework factory appending a file to the `config.files` list, and adding it to the `config.preprocessors` object as well. But for some reason it's not passed through my preprocessor. Is there some sort of timing issue where it's too late for the framework factory to add patterns to the `config.preprocessors` object? I can add them to `config.files` and they get included, just not processed.

Dignifiedquire:
Should be solved by #1341

PR #1341, opened 2015-03-07
fix(preprocessor): Lookup patterns once invoked

### LINKED ISSUE #1340: List of patterns to preprocess are saved in a closure too early
Currently the config value for which files to preprocess are saved in a closure when a preprocessor is created. Which means that if a pattern is removed or added after the preprocessor is created, the new pattern is missed because the patterns were saved in a closure.

This is an issue when you're creating a plugin and you want to add a file that you also want to preprocess. That is, in your framework factory function, you add an entry to `config.files` and add an entry to `config.preprocessors` for that file. Because of this bug, the preprocessor is never run for the added file.

andersekdahl:
Don't lookup the list of patterns of files to preprocess until the
preprocessor is called. This because patterns might be added after the
preprocessor is created, and those will be missed if the patterns are
looked up too early.

Closes #1340

### DIFF
--- a/lib/preprocessor.js
+++ b/lib/preprocessor.js
@@ -30,10 +30,10 @@ var isBinary = Object.create(null);
 
 // TODO(vojta): instantiate preprocessors at the start to show warnings immediately
 var createPreprocessor = function(config, basePath, injector) {
-  var patterns = Object.keys(config);
   var alreadyDisplayedWarnings = Object.create(null);
 
   return function(file, done) {
+    var patterns = Object.keys(config);
     var thisFileIsBinary = isBinary[path.extname(file.originalPath)];
     var preprocessors = [];
     var nextPreprocessor = function(error, content) {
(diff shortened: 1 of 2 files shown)


maksimr:
@andersekdahl Thanks

> This because **patterns might be added after the
preprocessor is created**, ...

Can you explain when such a case occurs

andersekdahl:
It's an issue when you're creating a plugin and you want to add a file that you also want to preprocess. That is, in your framework factory function, you add an entry to config.files and add an entry to config.preprocessors for that file. Because of this bug, the preprocessor is never run for the added file.

So this PR is perhaps not the best way forward for allowing this, but it's a very small, low-impact change. I'd much rather see some sort of stream-pipeline which you could push files through at any time, but I'm not familiar enough with the code base to propose such a change.

maksimr:
related issues 
https://github.com/karma-runner/karma/issues/1219

Dignifiedquire:
LGTM

PR #1345, opened 2015-03-11
fix: upgrade http-proxy module for bug fixes

terinjokes:
The http-proxy module was significantly outdated, and was missing many
bugfixes to support node newer than 0.10, as we as a refactor and
simplification of their codebase.

This commit upgrades the http-proxy dependency, and refactors the proxy
middleware to utilize the new API.

`parseProxyConfig` (an internal function) now takes the proxy
configuration object and converts it into an array of "ProxyRecord"
objects, sorted by path. A ProxyRecord object contains:

  * `host` The remote proxy host
  * `port` The remote proxy port
  * `baseUrl` The remote proxy path
  * `path` The local URL path (that triggers the rewrite)
  * `https` Boolean to determine if the remote connection will be made
    with SSL.
  * `proxy` The instance of http-proxy.

This change was necessitated by the removal of http-proxy's
RoutingProxy API, and the desire to only create one proxy instance per
configuration item.

The middleware simply determines if the current request matches a known
proxy path, and calls the proxy's `.web` and `.ws` methods with the
current request and response objects.

The relevant unit tests were rewritten in light of these changes.


megawac:
big :+1:

Dignifiedquire:
This is great, a big thanks first of all! 
There is only one question I have, what about the option `proxyValidateSSL`, why did you remove it?

package.json
```diff
@@ -167,7 +167,7 @@
     "chokidar": ">=0.8.2",
     "glob": "~3.2.7",
     "minimatch": "~0.2",
-    "http-proxy": "~0.10",
+    "http-proxy": "~1.8.1",
```
Dignifiedquire: Could you change this to `^1.11.1` please?

Dignifiedquire:
ping @terinjokes

terinjokes:
@Dignifiedquire sorry, i've been busy and didn't check in on this.

I no longer remember why I modified `proxyValidateSSL`, but I'll rebase this today and try to answer it for you. :+1:

Dignifiedquire:
Thanks

Dignifiedquire:
Merged as bcce56325eccc4758087a857d67be8a9366e91d6

PR #1352, opened 2015-03-17
chore(package.json): remove engines field

stevemao:
It warns on iojs and 0.12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: node_js
 node_js:
+  - "io.js"
+  - "0.12"
   - "0.10"
   - "0.8"
 


maksimr:
@stevemao Thanks!

We also should [fix documentation](https://github.com/karma-runner/karma/blob/master/docs/intro/01-installation.md) where we describe which version of node we supported.
I prefer explicitly add stable version of node which we supported, instead remove them.(IMHO)

/cc @karma-runner/contributors

aymericbeaumet:
I agree with @maksimr on this. However I find it confusing to mix the node.js and io.js engine versions in such case (`~0.8 || ~0.10 || ~0.12 || ~1`). So @stevemao's solution could actually be the best.

Dignifiedquire:
As we don't support `0.6` or lower we should at least do something like `>= 0.8`.

maksimr:
Can we write something like

```js
"engines": {
  "node":  "~0.8 || ~0.10 || ~0.12",
  "iojs": "~1"
}
```

or

```js
"engines": {
  "node":  ">=0.8 <=0.12",
  "iojs": "~1"
}
```

aymericbeaumet:
[This could be worth a try](https://docs.npmjs.com/files/package.json#engines).

stevemao:
Thanks, before I sent this I looked at many other node projects, many of them removed this field. Some using `>= 0.10` similar to what @Dignifiedquire pointed out. In fact not many support 0.8 any more. I personal like @maksimr 's solution :+1:

PR #1353, opened 2015-03-17
Engines

stevemao:
continue with #1352

### DIFF
--- a/docs/intro/01-installation.md
+++ b/docs/intro/01-installation.md
@@ -1,11 +1,11 @@
 Karma runs on [Node.js] and is available as an [NPM] package.
 
-## Installing Node.js
+## Installing Node.js or iojs
 
-There are [installers](http://nodejs.org/download/) for both Mac and Windows.
+There are [node.js](http://nodejs.org/download/) or [iojs](https://iojs.org/en/index.html) for both Mac and Windows.
 On Linux, we recommend using [NVM](https://github.com/creationix/nvm).
 
-Note: Karma works on the two latest stable versions. That is **0.8.x** and **0.10.x** at this point.
+Note: Karma works on the three latest stable versions of node. That is **0.8.x**, **0.10.x** and **0.12.x** at this point. Also works on iojs 1.x.x.
 
 
 ## Installing Karma and plugins
--- a/package.json
+++ b/package.json
@@ -229,7 +229,8 @@
   "main": "./lib/index",
   "bin": {},
   "engines": {
-    "node": "~0.8 || ~0.10"
+    "node":  ">=0.8 <=0.12",
+    "iojs": "~1"
   },
   "version": "0.12.31",
   "license": "MIT"


maksimr:
Oh... seems it [does not work well](https://travis-ci.org/karma-runner/karma/jobs/54700586#L648) with iojs. Possible for iojs this warning not so bad.

Thanks

**LGTM**

stevemao:
I don't understand why it says `"node":"1.5.1"` instead of `"iojs":"1.5.1"`. Should I remove iojs in the engine?

Dignifiedquire:
LGTM

mzgol:
Yes, this still warns:
```
npm WARN engine karma@0.12.31: wanted: {"node":">=0.8 <=0.12","iojs":"~1"} (current: {"node":"1.5.1","npm":"2.7.0"})
```

iojs says `node` because that's what it's been in Node and they don't want to encourage "engine-sniffing". See https://github.com/iojs/io.js/issues/269 for a discussion.

Generally, IMO it should just say:
```
"node":  "~0.8 || ~0.12 || ^1.0"
```
if anything needs to be specified at all. Otherwise io.js users will get warnings indefinitely, making them feel they're in an unsupported scenario.

stevemao:
What happens when node releases 1.5.1? Basically there is no way to tell if it's iojs 1.5.1 or nodejs 1.5.1 then?

mzgol:
> What happens when node releases 1.5.1? Basically there is no way to tell if it's iojs 1.5.1 or nodejs 1.5.1 then?

Yes, although the current thinking is that before Node.js reaches 1.0.0 the projects will reunite. If that doesn't happen then a different solution might be created but judging by the fact io.js seems reluctant to advertise itself as an "engine" it might not add the `iojs` field. Maybe @mikeal/@othiym23/@iarna would know more about plans.

That said, nothing (except maybe Heroku) understands the `iojs` field so adding it is kind of pointless.

othiym23:
@mzgol Your explanation is dead-on, as I understand things. I'm not 100% comfortable with there not being a way to distinguish between Node.js and io.js from the `engines` field, but for now, the only way to tell the difference between the two is to do V8 feature sniffing or to look at the version. I think we're all hoping that the two projects rejoin long before this becomes an issue in practice.

stevemao:
Thanks for explaining this guys. This is just bizarre. 
btw, `jsdom` 4 claims that they only work with `iojs` and they don't actually have an engine field in package.json.

iamstarkov:
iojs released 2.0 version, and I’m getting errors again. fixed it in pr #1394

mzgol:
The `iojs` field is ignored by io.js; I'd rather just remove the field and change `node` to `">=0.8 <=0.12 || >=1"`. @zzo, what do you think?

zzo:
unfortunate but sgtm thanks

iamstarkov:
you are right, iojs don’t care, but npm do

iamstarkov:
Here is a proof from my terminal

    ➜  karma-test git:(master) ✗ n io latest
    ➜  karma-test git:(master) ✗ node -v
    v2.0.0
    ➜  karma-test git:(master) ✗ npm update -g npm
    ➜  karma-test git:(master) ✗ npm -v
    2.9.0
    ➜  karma-test git:(master) ✗ less package.json | grep karma
        "karma": "^0.12.31",
    ➜  karma-test git:(master) ✗ npm i
    npm WARN engine karma-cli@0.0.4: wanted: {"node":"~0.8 || ~0.10"} (current: {"node":"2.0.0","npm":"2.9.0"})
    npm WARN engine karma@0.12.31: wanted: {"node":"~0.8 || ~0.10"} (current: {"node":"2.0.0","npm":"2.9.0"})
    ^C

mzgol:
@iamstarkov `npm` looks at the `node` field, not at the `iojs` field, it's clear from your terminal logs.

iamstarkov:
Wow, you are right, it wasn’t clear for me, and also been confused that karma used to have iojs field. Updated pr

PR #1357, opened 2015-03-19
fix(config): Default remaining client options if any are set

### LINKED ISSUE #961: client.captureConsole isn't defaulted correctly if client config is defined
The new client.captureConsole setting added in #850 (for #744) doesn't get defaulted correctly if you have any client options set. e.g:

    client: {
      args: [ /* some args */ ],
    },

In this case client.captureConsole is undefined, and you never see any console messages even though it's supposed to be `true` by default.

This possibly affects all default arguments that are nested.

firesock:
Closes #961

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -138,6 +138,12 @@ var normalizeConfig = function(config, configFilePath) {
     throw new Error('Invalid configuration: client.args must be an array of strings');
   }
 
+  var defaultClient = config.defaultClient || {};
+  Object.keys(defaultClient).forEach(function(key) {
+    var option = config.client[key];
+    config.client[key] = helper.isDefined(option) ? option : defaultClient[key];
+  });
+
   // normalize preprocessors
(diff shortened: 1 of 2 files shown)


firesock:
Not sure what broke 0.12 node build, any ideas?

nemoDreamer:
:panda_face:

firesock:
@maksimr Mainly wanted to keep the defaults in the same place in the file and access in a similar way to other defaults. Is it preferable to just use a simple top-level variable?

Dignifiedquire:
lgtm

maksimr:
@firesock all properties of instance `Config` user can change in `karma.config.js`, I dont think make `defaultConfig` as *public* property of `config` good idea.

Thanks!

Dignifiedquire:
#merge

firesock:
Sorry, was away for a while, couldn't pick up the comment. It was merged anyway, so I assume that's fine. Let me know if you still want the changes?

PR #1360, opened 2015-03-23
feat: Support for custom context and/or debug html files

aiboy:
Adding support for custom context.html and debug.html
```javascript
  config.set({

    client: {
      // paths must be relative to CWD ( current working directory )
      contextFile: '/context.html',
      debugFile: '/debug.html'
    },

    ...
```
Ref: #488

### DIFF
--- a/lib/middleware/karma.js
+++ b/lib/middleware/karma.js
@@ -59,7 +59,7 @@ var getXUACompatibleUrl = function(url) {
   return value;
 };
 
-var createKarmaMiddleware = function(filesPromise, serveStaticFile,
+var createKarmaMiddleware = function(filesPromise, serveStaticFile, serveFile,
     /* config.basePath */ basePath,  /* config.urlRoot */ urlRoot, /* config.client */ client) {
 
   return function(request, response, next) {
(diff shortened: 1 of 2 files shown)


leehambley:
Stumbled upon this from #488. I'm keen on this idea, as I'm struggling with QUnit and es6 module loaders, the problem is QUnit racing against the async module loaders, which *appears* to mandate something akin to the following as a solution (but as yet untested) https://github.com/jrburke/requirejs/wiki/Test-frameworks#qunit

jrolfs:
I'm trying to migrate our test suite (10,000+) from jasmine-rails to Karma and this would definitely help. A lot of our legacy tests insert into a specific div in our current runner html.

leehambley:
@jrolfs for what it's worth (and, that's "not much", I know) I "solved" my problem by using https://github.com/opudalo/generator-es6-lib which is a completely different stack of tools…

jrolfs:
@leehambley thanks for the suggestion (will be useful for other projects :smiley:), however unfortunately it's not really an option for the project in question.

Dignifiedquire:
Thanks, this looks mostly good, a few things to do please

* follow our [commit message convention](http://karma-runner.github.io/0.12/dev/git-commit-msg.html)
* update the docs so people can find these options
* these options shouldn't be nested in the client configs, they should be top level and default to the ones karma brings
* path normalization and setting defaults should be done in `lib/config.js`
* please rebase onto the `canary` branch and make the PR against `canary` as this will become the new stable version in the next days

aiboy:
Ok. I will try to fix all of those point as soon as possible. Is it ok if in the end I'll make a new pull request ?

Dignifiedquire:
> Is it ok if in the end I'll make a new pull request ?

Sure, github doesn't let you change the branch sadly

aiboy:
Ok. I will mention new pull request here..

aiboy:
I signed it!

Sanjo:
I have applied the changes to the 0.13.9 release here: https://github.com/Sanjo/karma/tree/0.13.9%2Bcustom-context-html.
I have signed the CLA in the past.

lib/middleware/karma.js
```diff
@@ -101,7 +101,19 @@ var createKarmaMiddleware = function(filesPromise, serveStaticFile,
     // or debug.html - execution context without channel to the server
     if (requestUrl === '/context.html' || requestUrl === '/debug.html') {
       return filesPromise.then(function(files) {
-        serveStaticFile(requestUrl, response, function(data) {
+
+        var fileServer;
+        if (client && client.contextFile) {
+            fileServer = serveFile;
```
UnsungHero97: it seems like the rest of the file follows 2-space indentations; you have 4-space

Dignifiedquire:
@Sanjo could you open a new pr with your changes please?

Dignifiedquire:
Closing due to inactivity

PR #1361, opened 2015-03-26
added karma to "bin"

donaldpipowitch:
Without installing Karma globally and only as a local devDependency, we can now write something like this into our `package.json` and call `$ npm run karma`:
````
{
  "scripts": {
    "karma": "karma start karma.conf.js"
  },
  "devDependencies": {
    "karma": "^0.12.0"
  }
}

````

### DIFF
--- a/package.json
+++ b/package.json
@@ -227,7 +227,9 @@
     "which": "~1.0"
   },
   "main": "./lib/index",
-  "bin": {},
+  "bin": {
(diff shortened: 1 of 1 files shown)


maksimr:
@donaldpipowitch Thanks

You can install `karma-cli` locally it should fix your problem

https://github.com/karma-runner/karma/issues/1078

PR #1368, opened 2015-04-03
feat(runner): provide error code on 'ECONNREFUSED' callback

Merott:
Add the error code as the second argument to the error callback. The information
can be useful in determining what to do on error when writing plugins against
the Karma API.

### DIFF
--- a/docs/dev/04-public-api.md
+++ b/docs/dev/04-public-api.md
@@ -29,3 +29,7 @@ runner.run({port: 9876}, function(exitCode) {
   process.exit(exitCode);
 });
 ```
+
+## Callback function notes
+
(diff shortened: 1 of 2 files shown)


Dignifiedquire:
Thanks, could you add this to the documentation as well?

Merott:
Not sure what format it's expected to be in, but no problem - I've updated the PR.

Dignifiedquire:
Thanks that's fine, just a note so people can actually find it at some point :)

Dignifiedquire:
LGTM

Dignifiedquire:
Merged into canary as 103ec6a5e04e17122d72e7fb9777cdf4e23e2690

PR #1381, opened 2015-04-23
chore: support wallabyjs for karma tests

maksimr:
kww

Thanks @ArtemGovorov for setting up [wallaby](http://wallabyjs.com) in karma.


Dignifiedquire:
@maksimr is this just so that we can use wallaby or is this actually a user face feature? (If so there should be some docs about this)

maksimr:
@Dignifiedquire just we can use wallaby for development `karma`

Dignifiedquire:
LGTM

maksimr:
#merge

Issue #1369, opened 2015-04-07
Provide API or event emitter example to cancel a running set of tests.


ajbogh:
In certain development cases there may be a large number of unit tests that take many seconds or minutes to run. In these cases the developer may be actively programming and saving, which could execute gulp watch events that force another unit test run. Currently when a set of unit tests is running Karma can execute another set asynchronously. 
 
It would be nice if the developer could develop a method to stop a running task before starting a new one. 

Example:

    if( karmaIsRunning ){
        karma.stopRun(); //stops actively running tasks before starting a new set of tests.
    }
    karma.start(config, function(){});

ajbogh:
Here's a PR which accomplishes the task. Please feel free to use it for its idea, not necessarily for the code.

https://github.com/ajbogh/karma/pull/1

Dignifiedquire:
Are you only interested in stopping? If so would what is talked about here #136 suffice for you? That is only talking about a command line option, but I would add an api option for it as well at the same time probably.

ajbogh:
I'm interested in stopping a running task programmatically because I'm not using the command line. In my particular case I was using Gulp with gulp-karma. I found that it was nearly impossible to kill Karma in a nice way, and as the developer of gulp-karma states "Karma. just. won't. stop." 

I made this PR to give advanced developers and plugin developers a way to stop Karma in a semi-nice manner.

Dignifiedquire:
Okay got it, I will probably use a different implementation, but I agree that we need to add this.

ajbogh:
Thanks!

schmuli:
I am working on an [Atom plugin](https://github.com/schmuli/atom-karma-runner) for running Karma, and just ran into this issue as well. Up till now, I have run Karma in the same process as Atom, without spawning, and this appears to be working.

However, when looking for a way to stop Karma, I ran into this issue. Even after looking at the recent API changes, including making the event emitter available, there still is not a way to stop Karma. Additionally, looking at the code in <code>lib/server.js</code>, it is still not easy to add this feature.

I should point out that I haven't found any clear documentation in Atom yet, but it may yet turn out to be a best practice to start such a plugin in a separate process, in which case I will just resort to killing the spawned process.

PR #1391, opened 2015-05-04
Add emittter to the server exports

ajbogh:
Adding the emitter to the exports allows a developer to stop a running set of tests using the below code:
    karma.emitter().emit('run_complete', [], {exitCode: 0});


Dignifiedquire:
Thanks, but will do a little bit different implementation. Progress will be tracked in #1369

PR #1394, opened 2015-05-07
update engine field because iojs 2.0

iamstarkov:
iojs just released 2.0 yesterday

### DIFF
--- a/package.json
+++ b/package.json
@@ -229,8 +229,7 @@
   "main": "./lib/index",
   "bin": {},
   "engines": {
-    "node":  ">=0.8 <=0.12",
-    "iojs": ">=1 <=2"
+    "node":  ">=0.8 <=0.12 || >=1 <=2"
   },
   "version": "0.12.31",
   "license": "MIT"
--- a/package.json
+++ b/package.json
@@ -230,7 +230,7 @@
   "bin": {},
   "engines": {
     "node":  ">=0.8 <=0.12",
-    "iojs": "~1"
+    "iojs": ">=1 <=2"
   },
   "version": "0.12.31",
   "license": "MIT"


iamstarkov:
Probably the same should be done for karma-cli project

mzgol:
The `iojs` field is ignored by io.js so this PR won't really change anything.

iamstarkov:
let’s continue in #1353 =)

mzgol:
I'll let @zzo decide but I'd even remove the `<=2` part, this is only causing problems and it's only advisory anyway.

zzo:
how is the <=2 causing problems?

mzgol:
> how is the <=2 causing problems?

I'm pretty sure io.js 3.0.0 comes soon enough that we'll have to change it again and all people that will be using an older Karma will see warning in a console while most likely everything will work; that's just confising & I don't see what's we're gaining for being so restrictive here. IMO adding an upper bound would be useful only if we knew the `karma` package has problems running on this newer version.

zzo:
Yes I agree with that however we upper-bound the nodejs version - but their releases are much slower than iojs I suppose.  Can we get rid of the upper bound for nodejs too?  Has anyone tried nodejs .13?

mzgol:
Well, the "node" field applies both to io.js & Node.js. There is no Node 0.13 yet and it will be a development line for the stable 0.14 so it doesn't matter too much.

zzo:
By your same logic we should get rid of the upper bound for the nodejs
version too.
I want to keep the upper limit on io.js in there as we're not testing
against 3.x so we don't know if it will work - think it's better than
having 3.x come out and a karma that doesn't work with it when our
package.json file says we do vs. it so happens to work with 3.x and we got
lucky.
We should also probably update .travis.yml to test against both io.js 1.x
and 2.x - wonder if nvm can do that?
I'm going to merge this as it is.
thanks!
  Mark

On Thu, May 7, 2015 at 8:56 AM, Michał Gołębiowski <notifications@github.com

zzo:
#merge

iamstarkov:
@zzo 

> and 2.x - wonder if nvm can do that?

nvm can do that, but problem that travis is not clear about defining iojs versions in it’s docs http://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Provided-Node.js-Versions

zzo:
bummer thanks

mzgol:
Travis docs are not very important here, under the hood it's just nvm which
is clear in supported version syntax. iojs-1 will do (IMO we should
additionally leave the iojs entry, though, to be sure we're always testing
against latest stable iojs).

-- 
Michał Gołębiowski

zzo:
Great i'll make a PR for it - also a sneaky way to test out the
mary-poppins server to auto-release karma on successful travis build.
thanks,
 Mark

zzo:
https://github.com/karma-runner/karma/pull/1395

PR #1395, opened 2015-05-07
Add iojs-1 to travis since we support it and 'stable' iojs is at 2.x

zzo:


mzgol:
For consistency we might also want to change io.js to iojs in that other place.

zzo:
sounds good.  Travis is lagging.

zzo:
#merge

PR #1401, opened 2015-05-12
fix(client): dynamic protocol for socket.io

### LINKED ISSUE #1400: Hardcoded http protocol for socket.io
`http` protocol is hardcoded in karma.js [here](https://github.com/karma-runner/karma/blob/master/client/main.js#L10) at the moment. This means that Karma does not work in an environment where `http` is not used (only `https`).

Karma can check `window.location.protocol` instead, which will allow to run it in `https`-only environments.

clexit:
Look up location.protocol for socket.io,
in order to use Karma in https-only enviournment

Closes #1400

### DIFF
--- a/client/main.js
+++ b/client/main.js
@@ -7,7 +7,7 @@ var KARMA_URL_ROOT = require('./constants').KARMA_URL_ROOT;
 
 // connect socket.io
 // https://github.com/LearnBoost/Socket.IO/wiki/Configuring-Socket.IO
-var socket = io.connect('http://' + location.host, {
+var socket = io.connect(location.protocol + '//' + location.host, {
   'reconnection delay': 500,
   'reconnection limit': 2000,
   'resource': KARMA_URL_ROOT.substr(1) + 'socket.io',


maksimr:
@clexit Thanks!

Are you use https proxy? 
Because karma run http server.

clexit:
Hi @maksimr!

Yes, it's https that proxies to the local Karma instance. Since there is only https, socket.io fails to connect because of the hardcoded protocol (it tries to connect from https to the inexistent http). While local http Karma is ok, we cannot configure http proxy due to security restrictions.

Can this change be merged? I've just noticed that the <scope> tag in the commit message might be incorrect. Does it matter for merge?

maksimr:
@clexit 

> Can this change be merged?
Yes, I think it's good change!

>  I've just noticed that the tag in the commit message might be incorrect. Does it matter for merge?
What tag? cli?

I think will be better change **scope** to `client`, because `cli` usual mean command line interface

`fix(client): ...`

Thanks

clexit:
Cool, thanks!

I changed the *scope* tag to `client`. The history should be OK now.

maksimr:
@clexit Thanks

**LGTM**

Dignifiedquire:
LGTM

maksimr:
#merge

maksimr:
#merge

zzo:
#merge

Dignifiedquire:
#merge

PR #1258, opened 2014-12-03
Socketio 1.1 new

dhimil:

PR #1404, opened 2015-05-15
Upgrade to Socket io 1.3

Dignifiedquire:
This is very much a work in progress, but from my tests neither the PR in #1257 nor #1258 does actually work. I hope to finish this in the next couple of days.

**Update1:** This only works when manually pulling in the forked & built version of `karma-jasmine@0.1.5` that's why travis will be complaining, until we updated the release there.

**Update2:** Finally all tests passing locally :)

**Update3:** Update 1 is now obsolete as I switched the client tests to mocha.

### DIFF
--- a/lib/launchers/process.js
+++ b/lib/launchers/process.js
@@ -64,11 +64,11 @@ var ProcessLauncher = function(spawn, tempDir, timer) {
 
     var errorOutput = '';
 
-    self._process.on('close', function(code) {
+    self._process.on('exit', function(code) {
       self._onProcessExit(code, errorOutput);
     });
 
-    self._process.on('error', function(err) {
+    self._process.on('karma_error', function(err) {
       if (err.code === 'ENOENT') {
         self._retryLimit = -1;
         errorOutput = 'Can not find the binary ' + cmd + '\n\t' +
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ before_script:
   - sh -e /etc/init.d/xvfb start
   - npm install -g grunt-cli
   - rm -rf node_modules/karma
+  - cd node_modules && ln -nsf ../ karma && cd ../
   - export $(openssl aes-256-cbc -pass env:CREDENTIALS_PASS -d -in credentials)
 
 script:
--- a/client/karma.js
+++ b/client/karma.js
@@ -13,9 +13,8 @@ var Karma = function(socket, iframe, opener, navigator, location) {
   var queryParams = util.parseQueryParams(location.search);
   var browserId = queryParams.id || util.generateId('manual-');
   var returnUrl = queryParams['return_url' + ''] || null;
-  var currentTransport;
 
-  var resultsBufferLimit = 1;
+  var resultsBufferLimit = 50;
   var resultsBuffer = [];
 
   this.VERSION = constant.VERSION;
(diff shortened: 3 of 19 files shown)


maksimr:
@Dignifiedquire Awesome!! :)

Dignifiedquire:
@maksimr and everyone else, basics are working for me, now I need some more eyes on this to test it out please.

maksimr:
@Dignifiedquire check in our projects all work fine with latest version of `karma-jasmine`, `karma-mocha`, `karma-webpack` ;)

Dignifiedquire:
@maksimr great. Also I was tired of jasmine so I switched the client tests to run on mocha with browserify :)

maksimr:
**LGTM**

Dignifiedquire:
@nikku @bendrucker do you have any ideas why this is failing? All is running fine on my local machine (osx + FF) but it seems travis doesn't like the setup :(

bendrucker:
karma-browserify has a hack to add a preprocessor for the bundle file: 

https://github.com/nikku/karma-browserify/blob/5d75d41879a32cc5a8b62b7c821b1ddbd796e870/lib/preprocessor.js#L3
https://github.com/nikku/karma-browserify/blob/5d75d41879a32cc5a8b62b7c821b1ddbd796e870/lib/preprocessor.js#L35

I'm betting the Karma under test !== the Karma that gets required on line 2. Try manually implementing the result from line 35 when you create your Karma config, adding: `'**/*': ['browserify-bundle']` to `config.preprocessors`.

Dignifiedquire:
Thanks a lot @bendrucker that fixed it :)

bendrucker:
Awesome. I'm guessing there's a reason npm doesn't handle that peer dep more cleverly. Couldn't find it though.

Dignifiedquire:
@bendrucker looks like that wasn't actually the issue, travis started failing after one successful run again. The issue is that we are deleting `node_modules/karma` but didn't symlink it, so `karma-browserify` couldn't find any installation of karma.

Dignifiedquire:
@maksimr need a new review, needed more fixes for tests including https://github.com/karma-runner/integration-tests/pull/4 to fix the integration tests.

lygstate:
it's better not use 1.3.5, use 1.3 should be better, cause only the master 
compat with iojs 2.1.0

lygstate:
Ping for review and pull

Dignifiedquire:
Merged as 79072aefee2a70d0fd53a47f159e455297b0bfdc

lib/launchers/process.js
```diff
@@ -64,11 +64,11 @@ var ProcessLauncher = function(spawn, tempDir, timer) {
 
     var errorOutput = '';
 
-    self._process.on('close', function(code) {
+    self._process.on('exit', function(code) {
       self._onProcessExit(code, errorOutput);
     });
 
-    self._process.on('error', function(err) {
+    self._process.on('karma_error', function(err) {
```
badcorporatelogo: This line does not look correct. The error event in this case is coming from the child process, indicating that spawn failed. It is different from the error event emitted by the socket, which was renamed from "error" to "karma_error".

Dignifiedquire: You are totally right, thanks a lot fixed in https://github.com/karma-runner/karma/commit/45a69221703bbd043cd71b8b0934c4be5339e111

PR #1429, opened 2015-05-28
fix(server): Start webserver and browsers after preprocessing completed

nmalaguti:
When loading many files, it is possible that the webserver and browsers
will start before file preprocessing has completed. This causes the
browser to attempt to connect and timeout and need to be relaunched.

This change will cause the server to wait to start the webserver and
browsers until after all files have been preprocessed. When a browser
connects to the webserver, the webserver will be able to start serving content
immediately and the browser will not experience any connection errors or
timeouts.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -31,15 +31,15 @@ var start = function(injector, config, launcher, globalEmitter, preprocess, file
     injector.get('framework:' + framework);
   });
 
-  var filesPromise = fileList.refresh();
+  // A map of launched browsers.
+  var singleRunDoneBrowsers = Object.create(null);
 
-  if (config.autoWatch) {
-    filesPromise.then(function() {
-      injector.invoke(watcher.watch);
-    }, function() {
-      injector.invoke(watcher.watch);
-    });
-  }
+  // Passing fake event emitter, so that it does not emit on the global,
+  // we don't care about these changes.
+  var singleRunBrowsers = new BrowserCollection(new EventEmitter());
+
+  // Some browsers did not get captured.
+  var singleRunBrowserNotCaptured = false;
 
   webServer.on('error', function(e) {
     if (e.code === 'EADDRINUSE') {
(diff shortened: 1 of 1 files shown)


nmalaguti:
@Dignifiedquire are you likely going to start accepting PRs this week?

Dignifiedquire:
@nmalaguti yes I will try to, just soo many repos, and so much backlog :/

nmalaguti:
@Dignifiedquire is there anything I can do to help out? I'd like to start using this fix as soon as possible.

Dignifiedquire:
I'll take a look tomorrow and will let you know

Dignifiedquire:
Okay, so I do like this, could you change the commit message to `fix(server)` instead of `feat` that way I can sneak it into the stable release, and I feel that this is more of a fix than a feature anyway.

nmalaguti:
@Dignifiedquire commit message changed.

Dignifiedquire:
Thanks, I just realized there is no test checking this behaviour. Do you think you can add one? 

PS. I'm gonna try to do a release tonight, so if you are quick I'll get it in ;)

nmalaguti:
@Dignifiedquire The specific behavior that the web server isn't started until after the files have been preprocessed isn't tested directly. There are many end to end tests that verify that the web server and browsers are still started, so I have confidence that this change hasn't caused any regressions.

I would like to have unit tests ensuring that this behavior doesn't regress in the future. This behavior is specific to `server.js` which is where I would like to test it. Unfortunately no unit tests currently exist for `server.js`, and figuring out all the mocks for it looks like a significant undertaking.

I'd be happy to start working on this, but I don't have high confidence that I'll have something working today.

Dignifiedquire:
Right, I remember this embarrassingly empty file :(
If you want to wirkin those tests that wiuld be great!
I'll pull this in without those in the meantime

Dignifiedquire:
Thanks a lot. Released in 0.12.35

PR #1437, opened 2015-06-07
chore: Move to standard JS style.

Dignifiedquire:
Make life as a developer a little easier. For more info see
https://github.com/feross/standard.

PR #1440, opened 2015-06-08
refactor: Replace q with bluebird

Dignifiedquire:
[Bluebird](https://github.com/petkaantonov/bluebird) is faster, more spec compliant and has a much better set of built in tooling.

Depends #1437


Dignifiedquire:
I signed it!

PR #1442, opened 2015-06-09
feat(runner): serve context in JSON format for JS-only environments

tailsu:
serve files to load at /context.json

### DIFF
--- a/lib/middleware/karma.js
+++ b/lib/middleware/karma.js
@@ -157,6 +157,16 @@ var createKarmaMiddleware = function (filesPromise, serveStaticFile,
             ' some errors:\\n  ' + errorFiles.join('\\n  ') + '");')
         })
       })
+    } else if (requestUrl === '/context.json') {
+      return filesPromise.then(function (files) {
+        common.setNoCacheHeaders(response)
+        response.writeHead(200)
+        response.end(JSON.stringify({
+          files: files.included.map(function (file) {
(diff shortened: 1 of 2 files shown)


Dignifiedquire:
Interesting, I expect this has the same reasoning behind as https://github.com/karma-runner/karma-jasmine/pull/93 ? 
Just being curious, how exactly do you actually use `context.json` with this change?

On another note, it would be nice to add a section to the documentation explaining that this is available, and how it could be used.

tailsu:
Yes. The prototype I'm working on uses the socket.io client through XHR polling (NS has XHR) to connect to the karma server. When an 'execute' is received, the runner downloads context.json (previously it downloaded and parsed context.html), downloads and `require`'s all script files. I ported the karma client code (`window.__karma__`) to NS, so the adapters can run with no modifications other than the occasional implicit expectation that it's running in an actual browser.

I'll see if I can share the work in progress with you if you're interested.

tailsu:
Uh, can you point me to a place in the documentation where this fits? This is an implementation detail useful only for trailblazers in the JS-only field. :worried:

tailsu:
https://github.com/NativeScript/karma-nativescript-launcher

Dignifiedquire:
Ah sorry for the delayed answer. You could add a new page in docs/plus for it

Dignifiedquire:
@tailsu I'm going to merge this for now, when you are ready just do a new PR with the documentation.

PR #1443, opened 2015-06-09
fix(file_list): follow symlinks

gavinaiken:
Changing to the latest version of the glob module means that
karma no longer follows symlinks to find files to serve.
Setting foillow:true in the glob options enables this
behaviour again.

### DIFF
--- a/lib/file_list.js
+++ b/lib/file_list.js
@@ -53,6 +53,7 @@ Url.prototype.toString = File.prototype.toString = function () {
 
 var GLOB_OPTS = {
   // globDebug: true,
+  follow: true,
   cwd: '/'
 }
 


Dignifiedquire:
Thanks!

Dignifiedquire:
Do you think you could add a test, so we don't get this regression again?

gavinaiken:
Ha! I was afraid you were going to ask that. I was wary of doing it because I wasn't sure if that might break tests on Windows, but I'll look into it tomorrow and see if I can add something to this PR.

Dignifiedquire:
We all love our Windows :wink: 
Thanks for looking into it, if you need any help I'm happy to help out

gavinaiken:
So it looks like the test/unit/file_list.spec.coffee tests all mock the glob function. Is that right? And if so, where would you expect to see this test, since it will need to use the real glob function, not a mock? OK to add it to the same file?

Dignifiedquire:
That's the problem when you have too many mocks in your tests.. **Adding todo note: refactor tests to use less mocks!**
Yeah I'm fine with adding your test into there.

Dignifiedquire:
Merging this for now as a temp fix for stable.

Issue #1436, opened 2015-06-07
TypeError: Cannot read property 'length' of undefined in changeFile


cassus:
Occurs when I save a file that is still being processed by karma-webpack preprocessor.

```
ERROR [karma]: [TypeError: Cannot read property 'length' of undefined]
TypeError: Cannot read property 'length' of undefined
    at changeFile (.../node_modules/karma/lib/file_list.js:345:33)
    at EventEmitter.<anonymous> (.../node_modules/karma/lib/watcher.js:90:17)
    at EventEmitter.emit (events.js:110:17)
    at EventEmitter.<anonymous> (.../node_modules/karma/node_modules/chokidar/index.js:136:15)
    at EventEmitter.FSWatcher._emit (.../node_modules/karma/node_modules/chokidar/index.js:149:5)
    at EventEmitter.<anonymous> (.../node_modules/karma/node_modules/chokidar/lib/nodefs-handler.js:268:12)
    at .../node_modules/karma/node_modules/chokidar/lib/nodefs-handler.js:176:13
    at Array.forEach (native)
    at StatWatcher.<anonymous> (.../node_modules/karma/node_modules/chokidar/lib/nodefs-handler.js:175:31)
    at StatWatcher.emit (events.js:110:17)
    at StatWatcher._handle.onchange (fs.js:1226:10)
```

code context:
```
    for (i = 0; i < buckets.length; i++) {
      for (j = 0; j < buckets[i].length /* Error here */; j++) {
        if (buckets[i][j].originalPath === path) {
          break outer;
        }
      }
    }
```

variables:
```
path == .../app/components/RegisterPage/__tests__/index.js
i == 5
buckets.length == 6
buckets[i] == undefined
buckets[i-1] == []
```

Dignifiedquire:
What version are you running?

cassus:
```
├─┬ karma@0.12.35
├─┬ karma-webpack@1.5.1
```

Dignifiedquire:
Thanks, I'll take a look

cassus:
I also have sometime "Some of your tests did a full page reload!" errors before this crash happens. This might be a helpful cue.

I'm still trying to isolate that error / make it somewhat reproducible.

Dignifiedquire:
Okay thanks, I just did a basic test with https://github.com/webpack/karma-webpack/blob/master/example/karma.conf.js and that worked fine while editing the files. 
[...]

PR #1447, opened 2015-06-12
Upgrade to latest glob.

zzo:
Use sync glob (much faster) plus get stat info from glob itself to avoid having to stat to get mtime and if this is a directory or not


zzo:
signed it

zzo:
@Dignifiedquire what do you think of this?  Want to get this into the next canary and then put it all into the angular builds

lib/file_list.js
```diff
@@ -71,6 +63,7 @@ var byPath = function (a, b) {
 var List = function (patterns, excludes, emitter, preprocess, batchInterval) {
   var self = this
   var pendingDeferred
+  var okToClearPendingDeferred = true
```
Dignifiedquire: What exactly is this variable supposed to be tracking?

Dignifiedquire:
Looks pretty good to me, that should make things faster :) 
The only thing I'm concerned about is windows..Have you tested this on windows machine if all the globbing is still working as expected there?

zzo:
Agreed I'm also concerned about Windows. 
Because I replaced an asynchronous method (the original glob) and got rid of another ( the stat) it's now possible for 'complete' to be called before 'refresh' actually returns - which nulls out the promise that refresh returns. So in that case we need to hold off before null'ing it out. 
The main thing wrong here (well 'not ideal' let's say!) is the 'pending' file counting that's going on to determine when the promise should be fulfilled. Votja has a todo in there about getting rid of that. 
That was more refactoring than I was willing to stomach for this right now. There is also the issue of the same promise being reused by all of the other methods in this file which maybe makes sense and maybe it doesn't. Again I didn't want to completely rip apart this file although I'm sure it's ripe for it someday.

Dignifiedquire:
There is also #1436 which is a bug in that file. I already promised to fix it, but that's only possible if I refactor this whole file..which is really badly needed.

zzo:
Interesting. Multiple refreshes must happen all of the time wonder why this
doesn't happen more often.

Dignifiedquire:
We don't want to push this into master, so closing this PR

PR #1448, opened 2015-06-12
fix (reporter): better handling of stack trace source map rewrite

maoueh:
In some cases, reported stack traces do not contain the url scheme and domain. This is the case when using the webpack preprocessor where stack entries look like: `at base/main.js?<sha1>:44977`. The previous url regexp was always assuming an url scheme and domain.

This commit make the url scheme and domain optional so more stack traces can be handled when trying to link from a transformed source code to the original location provided by the source map associated with the served file.

I might be fixing the wrong place. What I mean by this is that it's possible that the real bug is that when using webpack, the reported stack traces are not correct. Currently, there are reported like this `at base/test/main.js?1f50f4d71de2570a00bab944fc506d6900a3b1ed:44977` while from the current tests, they should have been written `at http://localhost:9876/test/main.js?1f50f4d71de2570a00bab944fc506d6900a3b1ed:44977`. If that's the case, will do the necessary research to track this elsewhere.

I built a trimmed down project using `karma` and `karma-webpack` if you would like to try.

Regards,
Matt

### DIFF
--- a/lib/reporter.js
+++ b/lib/reporter.js
@@ -22,7 +22,7 @@ var createErrorFormatter = function (basePath, emitter, SourceMapConsumer) {
     return null
   }
 
-  var URL_REGEXP = new RegExp('http:\\/\\/[^\\/]*\\/' +
+  var URL_REGEXP = new RegExp('(http:\\/\\/[^\\/]*\\/)?' +
     '(base|absolute)' + // prefix
     '((?:[A-z]\\:)?[^\\?\\s\\:]*)' + // path
     '(\\?\\w*)?' + // sha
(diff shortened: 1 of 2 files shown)


Dignifiedquire:
Thanks. I'm not a 100% sure right now if this should be fixed somewhere else as well, but in general I feel that this addition does makes sense. If you want to put some more research into it, the host part is missing I'd be happy to hear about it. 
For now could you please rebase onto the latest master?

maoueh:
Ok I just rebased on latest master.

When you say else where, do you mean at another location in `karma` codebase or in the `karma-webpack` project?

Do you have some hints about where to look exactly for the missing host part? Where is the stack trace coming from exactly? I'm willing to hunt this down further, but your wisdom is welcome for this journey :)

Dignifiedquire:
Sorry @maoueh for the late response, I'm not entirely sure where to look to be honest. I will try and see if I can find anything out in the next days

maoueh:
No problem, I will try to hunt it down starting from `karma-webpack` plugin and tracks where the problem could be.

Dignifiedquire:
@maoueh did you ever find anything more?

maoueh:
Nop, never try really hard to find the actual root cause that make it different. I searched a bit but did not come up with any conclusion.

dignifiedquire:
Going to close this for now. Sourcemaps need work I know, but I don't have the time and knowledge myself at the moment to this.

PR #1449, opened 2015-06-12
refactor: Replace q with bluebird.

Dignifiedquire:
Bluebird is faster, more spec compliant and has a much better set of built in tooling.

Replaces #1440

### DIFF
--- a/lib/events.js
+++ b/lib/events.js
@@ -1,6 +1,5 @@
 var events = require('events')
 var util = require('util')
-var Q = require('q')
 
 var helper = require('./helper')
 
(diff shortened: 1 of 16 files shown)


Dignifiedquire:
I signed it!

Dignifiedquire:
@zzo after I have merged this, I will start with the refactoring of `file_list.js` on the canary branch, that way we can hopefully ship it together with `0.13`

By the way I'm still missing write access to this repo since this morning :(

zzo:
lemmie now when the canary is pushed to npm :)

PR #1482, opened 2015-07-04
feat(server): expose server event lifecycle in public API

### LINKED ISSUE #1037: Public API does not provide hooks for server started, all browsers captured, etc
### Problem

The [public API](http://karma-runner.github.io/0.12/dev/public-api.html) does not provide enough insight into the state of the server.

### Why?

Because server tasks that are dependent on its state can be triggered programmatically by outside entities, it is necessary for said entities to have knowledge of the server's state.

For instance, imagine if you wanted to do the following, programmatically:

1. Start a Karma server (`server.start()`)
2. When all browsers are connected, run tests (`runner.run()`)
3. Trigger additional runs at arbitrary times (such as when files change) (`runner.run()`)

Currently, this is not possible. Yes, this is exactly what `autoWatch` does, but runs cannot being triggered programatically. Yes, `singleRun` will run tests when all browsers are connected, but it will kill the server and exit when tests are done running, which makes the third step impossible.

### Workaround

In order to work around this, implementors are forced watch `process.stdout` for messages to infer what's going on inside of Karma, which is incredibly brittle and hackish.

### Solution

I tried digging to find a way to get access to the event emitter that triggers some of these events, but it seems it's not currently possible via the public API.

If `server.start` returned and object with a reference the `globalEmitter` object, we'd be halfway there.

nikaspran:
Return the EventEmitter instance so that it's possible to into various lifecycle events. Add a `browsers_ready` event so that it's possible to know when the karma server has finished starting up.

Enables and partially resolves #1037

I am very interested in resolving #1037 (it's breaking our grunt and gulp builds), so if this is not quite the way you want it to be done, let me know and I'll do it differently! :)

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -76,10 +76,14 @@ var start = function (injector, config, launcher, globalEmitter, preprocess, fil
   globalEmitter.on('browser_register', function (browser) {
     launcher.markCaptured(browser.id)
 
-    // TODO(vojta): This is lame, browser can get captured and then crash (before other browsers get
-    // captured).
-    if (config.autoWatch && launcher.areAllCaptured()) {
-      executor.schedule()
+    if (launcher.areAllCaptured()) {
+      globalEmitter.emit('browsers_ready')
+
+      // TODO(vojta): This is lame, browser can get captured and then crash (before other browsers get
+      // captured).
+      if (config.autoWatch) {
+        executor.schedule()
+      }
     }
   })
 
@@ -247,6 +251,8 @@ var start = function (injector, config, launcher, globalEmitter, preprocess, fil
     log.error(error)
     disconnectBrowsers(1)
   })
+
+  return globalEmitter
 }
 start.$inject = ['injector', 'config', 'launcher', 'emitter', 'preprocess', 'fileList',
   'webServer', 'capturedBrowsers', 'socketServer', 'executor', 'done']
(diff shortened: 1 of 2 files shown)


nikaspran:
I signed it!

Dignifiedquire:
So in general I like the approach, though I would prefer not returning the `globalEmitter` from the `start` function. In my mind the code (not only yours) needs to be refactored in a way that the interacting api would look like

```js
var karma = require('karma')

karma.on('browsers_ready', function (event) {
  // ...
})

karma.start()
```

but I'm not a 100% sure yet what the best way is to introduce these changes into the code base.

nikaspran:
Thank you for taking a look at this :)

It seems quite a major undertaking to refactor everything but I agree it's the best idea longterm. I could take a stab if it's not in the near future for any of the regular maintainers.

One option to have this quickly and then still support easy migration/backwards compatibility later is to return a subset of the emitter, sort of an interface object. So something like this:

```js
{on: globalEmitter.on.bind(globalEmitter), ...}
```

Then, in the future, it could just start returning karma itself. I think there's a general need to expand the public API a little bit to enable more diverse workflows.

Dignifiedquire:
So it turned out, it wasn't that hard to do :) see #1485 and feel free to give it a test drive

nikaspran:
Oh, that's actually impressively quick! It seems like it will do the trick quite well. I think we're good to close this PR and the issue I mentioned if #1485 goes through. I'll play around with your branch tomorrow.

:+1:

Dignifiedquire:
Great

Issue #1467, opened 2015-06-24
Notification on webserver port binding


segrey:
Based on https://github.com/karma-runner/karma/issues/614

WebStorm Karma integration listens to webserver port binding event (i.e. when this line printed https://github.com/karma-runner/karma/blob/v1.0.0-0/lib/server.js#L61).

Now the only way to accomplish this is to parse standard output of Karma server and mine data from lines like:

```
INFO [karma]: Karma v0.12.37 server started at http://localhost:9876/
```

Such an approach seems to be not very safe, e.g. it'll fail if `logLevel: config.LOG_DISABLE` is set. Well, actually, there is a workaround for this case: `LOG_DISABLE` is not allowed https://github.com/karma-runner/karma-intellij/blob/master/lib/intellij.conf.js#L50. But still it doesn't feel right.

It'd be nice if plugins could use some API to listen to webserver port binding event. Thanks!

Dignifiedquire:
Which part of the api exactly are you referring to? The plugin api or the node api?

segrey:
@Dignifiedquire No particular preference for API.
I was thinking of something like this
```
globalEmitter.on('karma_server_port_bound', function(port) {
});
```

Didn't test, but looks like there is another way:
```
webServer.on('listening', function() {
  // "config.port" is bound
});
```
The only problem here is that it depends on `karma/lib/server.js` implementation, i.e. that "config.port" is increased by one after each unsuccessful binding attempt.

Dignifiedquire:
@segrey I've implemented a solution in #1485 for you, with that you can do

```js
var Server = require('karma').Server
var server = new Server(config, done)

server.once('browsers_ready', function () {
  var alwaysSafePort = server.get('config').port 
})

**Comment by segrey** [comment]
@Dignifiedquire Sorry for the late reply. Thanks for looking into this. Really appreciate it.
Unfortunately, `browsers_ready` event is not fired exactly after webserver port binding, there could be a delay until browsers are ready. Nevertheless, `browser_ready` is useful event for IDE: WebStorm guesses this very moment using 'browserConnected' events. Thanks a lot for introducing it.

About webserver port binding event: I'm thinking of using
```
webServer.on('listening', function() {
  var actuallyBoundPort = config.port;
});
```
What do you think about legitimating this way?
By legitimating I mean adding it to https://github.com/karma-runner/karma/blob/master/docs/dev/04-public-api.md.

Dignifiedquire:
[...]

PR #1485, opened 2015-07-08
feat(server): improve public api

### LINKED ISSUE #1037: Public API does not provide hooks for server started, all browsers captured, etc

Dignifiedquire:
This adds a slew of new api possibilities to the server. All main
events
from the `globalEmitter` are now emitted on the `server` instances and
publicly available.
For a list of available events see the docs file.

BREAKING CHANGE:

The public api interface has changed to a constructor form. To upgrade
change

```javascript
var server = require(‘karma’).server
server.start(config, done)
```

to

```javascript
var Server = require(‘karma’).Server
var server = new Server(config, done)
server.start()
```

Closes #1037, #1482, #1467

### DIFF
--- a/docs/dev/04-public-api.md
+++ b/docs/dev/04-public-api.md
@@ -2,19 +2,92 @@ Most of the time, you will be using Karma directly from the command line.
 You can, however, call Karma programmatically from your node module. Here is the public API.
 
 
-## karma.server
+## karma.Server(options, [callback=process.exit])
 
-### **server.start(options, [callback=process.exit])**
+### Constructor
+
+```javascript
+var Server = require('karma').Server
+var server = new Server({port: 9876}, function(exitCode) {
+  console.log('Karma has exited with ' + exitCode)
+  process.exit(exitCode)
+})
+```
+
+### **server.start()**
 
 Equivalent of `karma start`.
 
 ```javascript
-var server = require('karma').server;
-server.start({port: 9876}, function(exitCode) {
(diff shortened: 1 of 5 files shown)


Dignifiedquire:
@zzo @maksimr @pkozlowski-opensource if you have some time, please take a look and let me know what you thinkg

zzo:
like it.  Now would be the time to get in breaking changes.

PR #1487, opened 2015-07-09
fix(web-server): Correctly update filesPromise on files updated

KJTsanaktsidis:
When the `file_list_modified` event was being emitted, the filesDeferred
was simply being resolved again; this has no effect and subsequent
calls to `filesPromise.then` continued to be resolved with the old
files.

This had the effect that issuing `karma run` did not result in new test
files being served.

This has been fixed by reverting the changes to `web-server.js` made in
fabdbae.

### DIFF
--- a/lib/web-server.js
+++ b/lib/web-server.js
@@ -2,8 +2,7 @@ var fs = require('fs')
 var http = require('http')
 var path = require('path')
 var connect = require('connect')
-
-var helper = require('./helper')
+var Promise = require('bluebird')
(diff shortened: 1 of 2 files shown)


Dignifiedquire:
Thanks a lot, what an oversight from me :(

Dignifiedquire:
Published as `karma@0.13.0-rc.8`

KJTsanaktsidis:
Thanks!

PR #1501, opened 2015-07-18
karma@0.13 compatibility with karma-intellij (fixes #1500)

### LINKED ISSUE #1500: karma@0.13 compatibility with karma-intellij
WebStorm 10.0.4 integration fails to run tests using karma v0.13 (https://github.com/karma-runner/karma-intellij/issues/26).
This problem is already fixed in karma-intellij, but, unfortunately, all IDEs released prior to this fix still fail to run tests with karma v0.13, because karma-intellij is bundled with IDE.

A possible solution to make karma integration working in already released IDEs could be to restore backward compatibility in karma v0.13, like this
```
./lib/server.js

// Backward-compatibility with karma-intellij bundled with WebStorm
// to be removed in karma 0.14
Server.start = function (config) {
  new Server(config).start()
}
```

Restoring backward compatibility would be much appreciated. Thanks.

segrey:


segrey:
I signed it!

2015-07-18 11:43 GMT+03:00 googlebot <notifications@github.com>:

lib/server.js
```diff
@@ -92,6 +92,14 @@ Server.prototype.start = function () {
   this._injector.invoke(this._start, this)
 }
 
+/**
+ * Backward-compatibility with karma-intellij bundled with WebStorm
+ * @deprecated since version 0.13, to be removed in 0.14
+ */
+Server.start = function (config) {
```
Dignifiedquire: This will not get picked up, because of https://github.com/karma-runner/karma/blob/master/lib/index.js#L9-L18

Dignifiedquire:
Could you please change the commit message to follow our [convention](http://karma-runner.github.io/0.13/dev/git-commit-msg.html)

lib/server.js
segrey: Ah, I wasn't aware of it: `karma-intellij` requires `./lib/server.js` directly, so the fix is working for karma-intellij. Agreed, the fix should be more general. Will update the PR soon.

segrey:
Closing in favor of https://github.com/karma-runner/karma/pull/1503

PR #1503, opened 2015-07-18
fix: restore backward compatibility for karma@0.13

### LINKED ISSUE #1500: karma@0.13 compatibility with karma-intellij

segrey:
This adds `start` method back to the object exported by `lib/server.js`.

Closes #1500

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -7,13 +7,13 @@ var launcher = require('./launcher')
 
 // TODO: remove in 1.0
 var oldServer = {
-  start: function () {
-    throw new Error(
-      'The api interface has changed. Please use \n' +
+  start: function (cliOptions, done) {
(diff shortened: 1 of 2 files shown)


Dignifiedquire:
Thanks :octocat:

PR #1509, opened 2015-07-21
feat(preprocessor): Capital letters in binary files extenstions

### LINKED ISSUE #1508: Static binary assets are not served correctly when they have capital letters in extension
When serving static binary assets configured with:
```javascript
files: [
	...
	{pattern: './fixtures/images/*.jpg', included: false, served: true},
	{pattern: './fixtures/images/*.JPG', included: false, served: true},
	...
]
```
Only images with *.jpg extensions are served properly. Capital letters extension shows a 'broken image' icon. This seems to be due to not recognising said image as a binary file. Adding "JPG" to `lib/binary-extensions.json` or modifying `lib/preprocessor.js` condition like so:
```javascript
var thisFileIsBinary = isBinary[path.extname(file.originalPath).toLowerCase()]
```
Solves the problem.

Not sure how to write a failing test for this to submit a fix.

lonelyelk:
Use lower case file extension to detect if it is binary

Closes #1508

### DIFF
--- a/lib/preprocessor.js
+++ b/lib/preprocessor.js
@@ -47,7 +47,7 @@ var createPreprocessor = function (config, basePath, injector) {
 
   return function preprocess (file, done) {
     patterns = Object.keys(config)
-    var thisFileIsBinary = isBinary[path.extname(file.originalPath)]
+    var thisFileIsBinary = isBinary[path.extname(file.originalPath).toLowerCase()]
     var preprocessors = []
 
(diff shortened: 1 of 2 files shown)


lonelyelk:
I signed it!

Dignifiedquire:
Thanks :octocat:

Dignifiedquire:
Published in `0.13.3`

lonelyelk:
Yay! Even released! Thank you!

PR #1511, opened 2015-07-22
chore(release): update conventional-changelog and add conventional-gi…

### LINKED ISSUE #1510: conventional-changelog 0.1.x is release
Just wondering if you guys are cool to switch to the latest version. All issues with the CHANGELOG here or in any related repos are fixed. Because of the rewrite, the syntax is completely changed. A few people are using it and I've been using it but I haven't found any problems so far. 
Related modules such as conventional-github-releaser should also be good to use ~~but need more tests~~. Tests are comprehensive now.

There are grunt/gulp plugins

stevemao:
…thub-releaser

Fixes #1510


gruntfile.js
```diff
@@ -108,6 +108,30 @@ module.exports = function (grunt) {
...
+        },
+        src: 'CHANGELOG.md'
+      },
+    },
+    conventionalGithubReleaser: {
+      release: {
+        options: {
+          auth: {
+            type: 'oauth',
+            token: process.env.GH_TOKEN
```
Dignifiedquire: This means I have to set `GH_TOKEN` to a token that has write access to the karma repo correct?

gruntfile.js
```diff
@@ -134,7 +158,8 @@ module.exports = function (grunt) {
       'npm-contributors',
       'bump:' + (type || 'patch') + ':bump-only',
       'build',
-      'changelog',
+      'conventionalChangelog',
+      'conventionalGithubReleaser',
```
Dignifiedquire: This should probably run after `bump-commit`, as at this point the changelog isn't commited yet and no tag created

stevemao: This shouldn't matter if `allBlocks = true`. `conventionalGithubReleaser` should work exactly the same way as `conventionalChangelog`. `allBlocks = true` here because we want to regenerate all previous ones (once this has run once we can remove it). Once `allBlocks = true` is removed we actually wanna run it between changing the version in package.json, and commit & tag, just like `conventionalChangelog`

gruntfile.js
stevemao: Oh yeah. forgot to mention :D

stevemao:
I think we could remove `allBlocks = true` once you have tried to regenerate the previous ones (this also proves it works) by running `grunt conventionalGithubReleaser`.

gruntfile.js
stevemao: Oh I see your point here... I'll have to fix this

stevemao:
I have fixed the issue. Can you please try?

Dignifiedquire:
Thanks I will test it out later today

Dignifiedquire:
@stevemao worked great, only two things, now that all is published, what do I need to change so it's only publishing the latest, and the other, it seems to tag the wrong commit, instead of the `chore: release ` commit, it tagged the previous commit as release tag, any ideas why?

stevemao:
Thanks @Dignifiedquire , to only publish the latest, just remove `releaseCount: 0` (I can PR) or leave it as it won't do anything bad just a bit slower.

> it seems to tag the wrong commit, instead of the chore: release commit, it tagged the previous commit as release tag

Are you talking about `conventional-changelog` or `conventional-github-releaser`? They are a bit different. `conventional-github-releaser` won't have such a problem as it only releases on the tag. if the tag doesn't exist, it releases on the `HEAD`.

stevemao:
I can't see any problem with the CHANGELOG.md or releases. Can you confirm?

Dignifiedquire:
I fixed them up manually, that's why they look fine.

What happened was that I ran manually

```bash
$ grunt npm-contributors
$ grunt bump-only:patch
$ grunt build
$ grunt conventionalChangelog
$ grunt  bump-commit
# Tag v0.13.4 is locally generated but not pushed yet
$ grunt conventionalGithubReleaser
# Results in v0.13.4 release and tag being created on Github 909ff4d5872476b28a905d1b829652d0ab5bbafc
```

stevemao:
hmmm.. it must be pushed before `conventionalGithubReleaser`. I thought `grunt bump-commit` would push both commit and tag?

Dignifiedquire:
Not if you set `push: false` https://github.com/karma-runner/karma/blob/master/gruntfile.js#L132 ;)

Dignifiedquire:
But I can change that, and enable pushing before hand, if that's required for it to work, no problem

stevemao:
Yeah your right. We can't set `push: false` here. Did you always have to manually push before?
We have a similar setup https://github.com/btford/grunt-conventional-changelog/blob/master/Gruntfile.js

PR #1515, opened 2015-07-24
fix(runner): Wait for file list refresh to finish before running

KJTsanaktsidis:
The runner middleware was starting a file list refresh before executing
a test run, but was not waiting for it to complete. This meant that the
context.html generation would often not see the new file list, and so
stale test files would be served.

I haven't actually managed to successfully run the client tests on my machine, but I *think* this change shouldn't have any adverse impacts. Everything should be assuming that the code in the runner is all asynchronous anyway. I did have to update some of the runner unit tests to match this assumption though.


Dignifiedquire:
Thanks a lot :octocat:

Issue #1519, opened 2015-07-27
timer.setTimeout causes Illegal Invocation error in a limited execution environment


schmuli:
I am writing an Atom plugin for running Karma, which currently uses the Atom process to start the Karma server. I believe that Atom runs its plugins in a limited execution environment (I had a similar issue when loading the karma plugin, but found an Atom-specific workaround).

However, I keep getting an <b>Illegal Invocation</b> exception in <code>lib/launchers/capture_timeout.js</code> on line 15, when trying to set a timeout. It appears that the exception is caused by <code>setTimeout</code> being called as a method of the <code>timer</code> object.

Changing the timer definition in <code>lib/server.js</code> to delegate to the <code>setTimeout/clearTimeout</code> functions, rather than pass by reference, eliminates the exception.

Is this a valid use-case for making a change in Karma's code base?

Dignifiedquire:
The main reason why those timers are injected is to be able to easily mock them during testing of Karma itself. If that turns out to be an issue in this scenario this is definitely a bug. 
Would you be up to creating a PR?

schmuli:
I added a PR but one of the unit-tests are failing for some reason, which doesn't appear to have anything to do with the changes I made.

Dignifiedquire:
@schmuli thanks I will check it out later

PR #1522, opened 2015-07-28
Updated timers for limited execution environment

schmuli:
In limited execution environments the setTimeout/clearTimeout functions
need to be delegated to, rather than stored as references.

This is a follow up to #1519

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -3,6 +3,8 @@ var di = require('di')
 var util = require('util')
 var Promise = require('bluebird')
 
+var root = global || window || this
+
 var cfg = require('./config')
 var logger = require('./logger')
 var constant = require('./constants')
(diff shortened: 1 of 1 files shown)


schmuli:
I signed it!

lib/server.js
```diff
@@ -72,7 +74,10 @@ var Server = function (cliOptions, done) {
     reporter: ['factory', reporter.createReporters],
     capturedBrowsers: ['type', BrowserCollection],
     args: ['value', {}],
-    timer: ['value', {setTimeout: setTimeout, clearTimeout: clearTimeout}]
+    timer: ['value', {
+        setTimeout: function () { return setTimeout.apply(root, arguments)},
+        clearTimeout: function (timeoutId) { clearTimeout(timeoutId)}
```
Dignifiedquire: Shouldn't this be applied as well?

Dignifiedquire:
Thanks, can you please squash your commits and change the commit message to follow our [convention](http://karma-runner.github.io/0.13/dev/git-commit-msg.html).

lib/server.js
schmuli: setTimeout has variables args, whereas clearTimeout only ever receives a single arg. I can change it if you think this is less consistent.

Dignifiedquire:
Thanks :octocat:

PR #1525, opened 2015-07-29
Feature karma https webserver

KrekkieD:
Specifies 2 new config options: `protocol` and `httpsServerOptions` to allow Karma web server to run on https if desired. Default is still http.


lib/executor.js
```diff
@@ -10,7 +10,7 @@ var Executor = function (capturedBrowsers, config, emitter) {
     var nonReady = []
 
     if (!capturedBrowsers.length) {
-      log.warn('No captured browser, open http://%s:%s%s', config.hostname, config.port,
+      log.warn('No captured browser, open ' + (config.protocol || 'http:') + '//%s:%s%s', config.hostname, config.port,
```
Dignifiedquire: Please move the default setting into `lib/config.js`, so we don't have to add these checks everywhere

KrekkieD:
@Dignifiedquire I have made adjustment for deeper integration. The config value is not available in the unit tests as they test some classes directly without going through config. Changed implementation to use function arguments where possible, and updated unit tests accordingly. Feels more in line with how the app is set up.

KrekkieD:
Should [this line](https://github.com/KrekkieD/karma/blob/feature-karma-https-webserver/lib/reporter.js#L24) be updated to `'http[s]?:\\/\\/...'`? I am not sure what this function does, but it's the last hardcoded occurrence of `http:`.

lib/config.js
```diff
@@ -118,6 +118,13 @@ var normalizeConfig = function (config, configFilePath) {
   // normalize urlRoot
   config.urlRoot = normalizeUrlRoot(config.urlRoot)
 
+  config.protocol = config.protocol || 'http:'
+  if (['http:', 'https:'].indexOf(config.protocol) === -1) {
```
Dignifiedquire: Let's make the values `http` and `https` without the `:`, and strip it, if someone adds it by accident.

Dignifiedquire:
> Should this line be updated to 'http[s]?:\\/\\/...'? I am not sure what this function does, but it's the last hardcoded occurrence of http:.

Yes this needs to be updated as well

lib/config.js
KrekkieD: How about we turn that around; add the `:` when omitted? Reason being that [NodeJS's url.parse](https://nodejs.org/api/url.html#url_url) returns the protocol including `:`, and it is used the same way by `lib/middleware/proxy.js`. If the property includes the colon it is more consistent and allows further use of the `url` class where desired.

Dignifiedquire: yeah I'm okay with that

KrekkieD:
Changes made. The travis build seems a bit flaky but I don't think it's caused by my changes.

lib/config.js
```diff
@@ -118,6 +118,14 @@ var normalizeConfig = function (config, configFilePath) {
   // normalize urlRoot
   config.urlRoot = normalizeUrlRoot(config.urlRoot)
 
+  // force protocol to end with ':'
+  config.protocol = (config.protocol || 'http').split(':')[0] + ':'
+  if (config.protocol.match(/https?:/) === null) {
+    log.warn('"%s" is not a supported protocol, defaulting to "http:"',
+      config.protocol)
+    config.protocol = 'http:'
+  }
```
Dignifiedquire: Could you please add a test for the above logic in `test/unit/config.spec.js`

lib/web-server.js
```diff
@@ -61,7 +63,14 @@ var createWebServer = function (injector, emitter) {
...
 
-  var server = http.createServer(handler)
+  var serverClass = http
+  var serverArguments = [handler]
+
+  if (config.protocol === 'https:') {
+    serverClass = https
+    serverArguments.unshift(config.httpsServerOptions || {})
+  }
+  var server = serverClass.createServer.apply(null, serverArguments)
```
Dignifiedquire: Could you please add a test for this, making sure when `https` is set, it's actually used

Dignifiedquire:
Needs rebase onto latest master

KrekkieD:
Tests added and pulled in latest master.

Dignifiedquire:
Thanks @KrekkieD now squash it all together and we can get this shipped :)

KrekkieD:
I believe, after some Googling on how to do that properly after pulling in commits from master, it is done ;)

Dignifiedquire:
Thanks :octocat:

Dignifiedquire:
Released in [v0.13.5](https://github.com/karma-runner/karma/releases/tag/v0.13.5)

PR #1533, opened 2015-08-03
use lodash find() instead of Array.prototype.find

kketch:
When i upgraded to from Karma 0.12.x to 0.13.3, my karma setup stopped working.

file-list.js uses ES6 Array.prototype.find which of course fails with `TypeError: undefined is not a function` in normal node installations (without the harmony flag).

Replacing Array.prototype.find() with lodash.find() (which is already used in other places in file-list.js) seems to fix the issue.

### DIFF
--- a/lib/file-list.js
+++ b/lib/file-list.js
@@ -99,7 +99,7 @@ List.prototype._isExcluded = function (path) {
 //
 // Returns the match or `undefined` if none found.
 List.prototype._isIncluded = function (path) {
-  return this._patterns.find(function (pattern) {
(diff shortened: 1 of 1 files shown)


Dignifiedquire:
Thanks, could you change the commit message to follow our [convention](http://karma-runner.github.io/0.13/dev/git-commit-msg.html) please?

kketch:
fixed the commit message

Dignifiedquire:
Thanks :octocat:

Dignifiedquire:
Released in https://github.com/karma-runner/karma/releases/tag/v0.13.5

PR #1557, opened 2015-08-11
Throw error when browers option is not an array

nhunzaker:
I forgot that the `browsers` option must be an array. This commit 
catches this case and throws a more descriptive error instead of 
something a long the lines of `TypeError: undefined is not a function`


GitCop:

 * Commit: e92b99c6bd269d200de8b06628fb5e7f2a98e3fc

---

nhunzaker:
Yikes. Updating with checks.

PR #1558, opened 2015-08-11
fix(config): Error when browers option isn't array

nhunzaker:
I forgot that the `browsers` option must be an array. This commit catches this case and throws a more descriptive error instead of something a long the lines of `TypeError: undefined is not a function`

I *think* I've got everything formatted properly. Happy to make changes.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -149,6 +149,10 @@ var normalizeConfig = function (config, configFilePath) {
     throw new Error('Invalid configuration: client.args must be an array of strings')
   }
 
+  if (config.browsers && Array.isArray(config.browsers) === false) {
+    throw new TypeError('Invalid configuration: browsers option must be an array')
+  }
+
   var defaultClient = config.defaultClient || {}
   Object.keys(defaultClient).forEach(function (key) {
     var option = config.client[key]
--- a/test/unit/config.spec.js
+++ b/test/unit/config.spec.js
@@ -300,6 +300,16 @@ describe('config', () => {
       expect(config.preprocessors).not.to.have.property(resolveWinPath('*.coffee'))
       expect(config.preprocessors).to.have.property(resolveWinPath('/**/*.html'))
     })
+
+    it('should validate that the browser option is an array', () => {
+      var invalid = function () {
+        normalizeConfigWithDefaults({
+          browsers: 'Firefox'
+        })
+      }
+
+      expect(invalid).to.throw('Invalid configuration: browsers option must be an array')
+    })
   })
 
   describe('createPatternObject', () => {


Dignifiedquire:
Thanks, but checks like this should go into `lib/config.js`, specifically into this [function](https://github.com/karma-runner/karma/blob/master/lib/config.js#L80). Also if you add a check like this, please include a unit test for it :)

nhunzaker:
Rodger. Will do!

nhunzaker:
@Dignifiedquire Done. The unit test feels brittle. Chai's `expect().to.throw` is a bit goofy. Thoughts?

lib/config.js
```diff
@@ -149,6 +149,10 @@ var normalizeConfig = function (config, configFilePath) {
     throw new Error('Invalid configuration: client.args must be an array of strings')
   }
 
+  if ('browsers' in config && Array.isArray(config.browsers) === false) {
```
Dignifiedquire: Please use `if (config.browsers &&` instead, to be consistent with the rest of the code here.

test/unit/config.spec.js
```diff
@@ -300,6 +300,16 @@ describe('config', () => {
...
     })
+
+    it('should validate that the browser option is an array', () => {
+      var invalid = function() {
+        normalizeConfigWithDefaults({
+          browsers: ''
+        })
+      }
+
+      expect(invalid).to.throw(/Invalid configuration/)
```
Dignifiedquire: Try checking the full error message, to be sure it's the right error that is thrown:

```js
expect(invalid).to.throw('Invalid configuration: browsers option must be an array')
```

lib/config.js
nhunzaker: Done.

nhunzaker: The only drawback here is if `browsers` is provided as an empty string, but I wonder how often that'll be the case.

nhunzaker:
Updated with feedback and rebased from upstream.

Dignifiedquire:
Thanks, could you fix linting please?

nhunzaker:
Done. Thank you for your patience!

Dignifiedquire:
Thank you for staying with it :octocat:

DarthCharles:
hi, in which scenarios will the browsers option wont't be array?

DarthCharles:
Seems like a fancy solution, but i read that Array.isArray(); is actually a slow option. Wouldn't it be better to check it like this?  if (config.browsers && config.browsers.constructor !== Array) {};

maksimr:
@DarthCharles I dont think it's bottle neck :)

Thanks

PR #1568, opened 2015-08-27
add HackTimer.js to prevent slow tests when Chrome is in background

jamestalmage:
Some browsers (Chrome) drastically increase the `setTimeout` interval if the tab is not visible on screen. This creates [issues in karma][related] if you push Chrome to the background (i.e. while using `autoWatch`).

This inserts [HackTimer.js] before any other javascript on each html page (`client`, `debug`, `context`). That may have been overkill. I did not test to see if one page in particular was creating the problem. What I can say is that it *drastically* improves execution time for tests running in out-of-site Chrome tabs.

An [issue has been opened][chromium-issue] with the Chromium team to address this via a cli flag, but there is no movement on that yet. I have not tested to see if the problem exists in other browsers (I saw an SO posts that suggest it does).

At present, [HackTimer.js has not specified an open source license][no-license], so this PR is probably a no-go until that is resolved. There also appear to be no unit tests for the library, so that is a potential concern as well. Until those are resolved, this PR serves as a proof of concept.

An additional possibility would be making the inclusion of the scripts optional via the config. Ideally it would be configurable per browser, and have the ability to forcibly disable across all browsers via the karma `cli` (when running on CI servers for instance).

Interested parties can install this hotfix via:
```bash
 npm install --save-dev jamestalmage/karma#hack-timer-dist
```

[HackTimer.js]:https://github.com/turuslan/HackTimer
[related]:https://github.com/karma-runner/karma-chrome-launcher/issues/44
[no-license]:https://github.com/turuslan/HackTimer/issues/6
[chromium-issue]:https://code.google.com/p/chromium/issues/detail?id=485425

### DIFF
--- a/lib/middleware/karma.js
+++ b/lib/middleware/karma.js
@@ -97,7 +97,7 @@ var createKarmaMiddleware = function (filesPromise, serveStaticFile,
     }
 
     // serve the favicon
-    if (requestUrl === '/favicon.ico') {
+    if (requestUrl === '/favicon.ico' || requestUrl === '/HackTimer.js' || requestUrl === '/HackTimerWorker.js') {
       return serveStaticFile(requestUrl, response)
     }
 
--- a/static/HackTimer.js
+++ b/static/HackTimer.js
@@ -1,0 +1,137 @@
+//https://github.com/turuslan/HackTimer
+
+(function (workerScript) {
+  try {
+    var blob = new Blob (["\
+var fakeIdToId = {};\
+onmessage = function (event) {\
+	var data = event.data,\
+		name = data.name,\
+		fakeId = data.fakeId,\
+		time;\
+	if(data.hasOwnProperty('time')) {\
+		time = data.time;\
+	}\
+	switch (name) {\
+		case 'setInterval':\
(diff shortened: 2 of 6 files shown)


GitCop:

 * Commit: f07c6f3a8cd5ef071933f4e77a5f1f8295bbe407

---

jamestalmage:
Correction,

Install the hotfix via:
```
 npm install --save-dev jamestalmage/karma#hack-timer-dist
```

jamestalmage:
Hmmm.
After running this agains a larger production test suite, this still does not solve the problem. Not sure how I convinced myself it was working in earlier tests.

Closing, but somebody might still want to look into this.

jamestalmage:
@turuslan
I don't know if you are familiar with the karma project (it's a fairly popular test runner), but it's an ideal candidate for HackTimer. If you could look at this PR, I'd be interested to know if I was doing something wrong. I am pretty sure I was not supposed to explicitly include `HackTimerWorker.js`, but even after fixing that, things still don't work.

This PR is a bit outdated from everything I've actually tried. See this branch: https://github.com/karma-runner/karma/compare/master...jamestalmage:hack-timer-dist

PR #1579, opened 2015-09-09
Tweak tested/allowed Node.js/io.js versions

mzgol:
chore(*): Tweak supported Node.js versions, drop io.js

Current setup means karma won't be tested on future updates to the Node v4
line. Those updates are not allowed to introduce breaking changes and only
the newest one is officially supported so that's the one that should be tested.

In the same way, ythe config doesn't contain "0.10.0" or "0.12.0" but
"0.10" & "0.12" so it catches latest updates.

io.js testing has been dropped - io.js has no official support policy as
opposed to Node.js; also, it won't see any major releases.

Refs 6b8d30f234aa77c87ac8cd720114c90a66e7e88e

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,26 @@
 sudo: false
 language: node_js
 node_js:
-  - "iojs-v3.2.0"
-  - "0.12"
   - "0.10"
-  - "4.0.0"
+  - "0.12"
+  - "4"
+
+# Node.js 4 is supported but an optional socket.io dependency doesn't work yet, causing
+# a lot of error messages on Travis. Travis then terminates the build because of excessive
+# logging. All tests pass locally on Node.js 4.
+# TODO remove this exception when https://github.com/socketio/socket.io/issues/2228 is fixed.
+matrix:
+  allow_failures:
+    - node_js: "4"
 
 env:
   global:
   - SAUCE_USERNAME: vojtajina
   - secure: "bRVY+hYZwMf1SqVnMyZRJTLD0gN1hLx9/MwO8MM/qBiu3YNjXy49XElfMdzMKN6cZeKTmhcnjmZonbJuI1PQ2t+utGkyjnlVLJ/OlWptreKLzIlcbt4hrdPoTcjmUTwDWq9Ex9cVoYX8AzCasETttpczp3P+s3+vmOUj8z25JyU="
 
-# Make sure we have new NPM.
+# Make sure we have a new npm on Node 0.10.
 before_install:
-  - '[ "${TRAVIS_NODE_VERSION}" != "0.8"  ] || npm install -g npm@1.4.28'
-  - npm install -g npm
+  - '[ "${TRAVIS_NODE_VERSION}" != "0.10" ] || npm install -g npm'
   - npm config set loglevel warn
 
 before_script:
--- a/docs/intro/01-installation.md
+++ b/docs/intro/01-installation.md
@@ -1,11 +1,11 @@
 Karma runs on [Node.js] and is available as an [NPM] package.
 
-## Installing Node.js or iojs
+## Installing Node.js
 
-There are [node.js](http://nodejs.org/download/) or [iojs](https://iojs.org/en/index.html) for both Mac and Windows.
-On Linux, we recommend using [NVM](https://github.com/creationix/nvm).
+On Mac or Linux we recommend using [NVM](https://github.com/creationix/nvm). On Windows, download Node.js
+from [the official site](https://nodejs.org/).
 
-Note: Karma works on the two latest stable versions of node. That is **0.10.x** and **0.12.x** at this point. Also works on iojs 2.x.x.
+Note: Karma currently works on Node.js **0.10**, **0.12.x** and **4.x**. See [FAQ](04-faq.md#which-version-of-nodejs-does-karma-run-with-) for more info.
 
 ## Installing Karma and plugins
 
--- a/docs/intro/04-faq.md
+++ b/docs/intro/04-faq.md
@@ -25,7 +25,7 @@ The latest stable version from NPM (`npm install karma`). See [versioning] for m
 
 
 ### Which version of Node.js does Karma run with ?
-The three latest stable versions. That means `0.8`, `0.10` and `0.12` at this point. Also works on iojs `2.x.x`.
+Karma works on all LTS versions node in active maintenance state (see [LTS docs](https://github.com/nodejs/LTS/blob/master/README.md) for more info) as well as the latest stable version. That is **0.12.x** and **4.x** at this point. Additionally, Node **0.10** is currently supported.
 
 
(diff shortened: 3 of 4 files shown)


mzgol:
I can optionally remove io.js 3 from Travis. It won't be supported for much longer, there's no official support policy for io.js while Node.js 0.10/0.12 will officially be supported for another year and Node 4 for another 2.5 years.

mzgol:
Something is broken on master; my changes currently invoke tests on Node 4.0.0 since there is no newer version yet and yet it fails.

Dignifiedquire:
The integration tests fail on `node@4` currently. Need to investigate what's happening there.

Re iojs yes let's drop it, `4.0` has most of it anyway. Could you please also update the doc about what versions we support to mirror these changes please?

Dignifiedquire:
Damn, looks like because of all the optional dependencies travis is terminating the build, even though it succeeds for me locally.

See https://s3.amazonaws.com/archive.travis-ci.org/jobs/79468450/log.txt

mzgol:
> Re iojs yes let's drop it, 4.0 has most of it anyway. Could you please also update the doc about what versions we support to mirror these changes please?

Done. FAQ & installation docs had conflicting information so I made them consistent. There is duplicated info there, though, maybe on of them could be simplified?

I assumed we'll want to support all LTS releases in "active" support state, I didn't mention the "maintenance" support phase as it receives only security updates so it's mostly for apps already deployed and not modified anymore; new Karma releases shouldn't matter for that (read https://github.com/nodejs/LTS/blob/master/README.md for more info about the plan). Node 0.10 is still popular, though so I made an explicit exception for it. I also assumed we'll want to support the latest stable Node (every other stable release will become an LTS, new stable releases will appear each 6 months, not 6 weeks as it was with io.js).

The expected schedule would look as follows:
1. Node 5 is released - Karma then supports Node 0.12 (LTS, active phase), 4.x (LTS, active phase) and, as an exception, Node 0.10 (LTS, maintenance phase). It will also support Node 5 (stable).
2. Node 6 is released - Karma then supports Node 0.12 (LTS, active phase) and 4.x (LTS, active phase). It will also support Node 6 (stable).
3. Node 7 is released, Node 6 becomes an LTS - Karma then supports Node 4.x (LTS, active phase) and 6.x (LTS, active phase). It will also support Node 7 (stable).
4. Node 8 is released - Karma then supports Node 4.x (LTS, active phase) and 6.x (LTS, active phase). It will also support Node 8 (stable).

Does that sound OK?

mzgol:
I dropped an obsolete Node 0.8 entry from `.travis.yml` and made `npm` update only on 0.10 as it has npm@1 which has problems. npm in Node 0.12/4.0 is actively updated so it makes sense to test on the built-in version.

Dignifiedquire:
Sounds good, but it looks like travis is choking on your change for the `TRAVIS_NODE_VERSION` check

mzgol:
> Sounds good, but it looks like travis is choking on your change for the `TRAVIS_NODE_VERSION` check

Oops. Fixed.

mzgol:
Integration tests pass for me locally on Node 4.0.0 as well (OS X 10.11 public beta 6), although lots of optional deps fail to install in the process.

mzgol:
Isn't this the reason for the termination of the build?
```
The log length has exceeded the limit of 4 Megabytes (this usually means that test suite is raising the same exception over and over).

The build has been terminated.
```
It seems it's getting killed because it logs too much...

Dignifiedquire:
It's mostly one, all related to socket.io not upgrading their dependencies, see #1577 
The build would pass fine, but npm spits out so much log data about the optional deps failing that travis kills it, so we probably have to live with a failing build on travis for now :(

mzgol:
> The build would pass fine, but npm spits out so much log data about the optional deps failing that travis kills it, so we probably have to live with a failing build on travis for now :(

We can add v4 to accepted failures temporarily so that build is green. This should be fixed ASAP, though to make sure we don't regress on v4.

Let's hope https://github.com/socketio/socket.io/issues/2228 is resolved soon.

mzgol:
Do you want me to add:
```yaml
matrix:
  allow_failures:
    - node_js: "4"
```
to `.travis.yml` for now?

Dignifiedquire:
Yes let's do that, I'll have to check the travis builds manually to ensure nothing too bad happens

mzgol:
Done. I also reduced the duplication in the docs by referring to FAQ from installation notes.

mzgol:
https://travis-ci.org/karma-runner/karma/jobs/79486742 failed (Node 0.12), not sure why, I haven't changed anything related. A flake? Could you restart this one job?

mzgol:
Never mind, I updated the PR again. README previously refered to http://nodejs.org/download/ but this URL is hardly user-friendly. The main Node.js page is very clear how to download the proper Node for your platform so I linked to that instead.

Note that I changed the recommendation for OS X to use nvm as well. Not sure if that's easier for people than downloading the installer from the official site but the installed version doesn't update and Linux people can download from the site as well so I don't see why we should be inconsistent here - either point to the official site for all people or suggest nvm for both Linux & OS X.

Dignifiedquire:
Looks good to me, thanks a lot

Dignifiedquire:
One thing can you remove the `(*)` from the commit message, I think the parser will choke on that, just `chore: Tweak...` should be fine

GitCop:

 * Commit: ce1d26ffc549316b2333153a3e4bf8ac2dc712b4

---

mzgol:
> One thing can you remove the `(*)` from the commit message, I think the parser will choke on that, just `chore: Tweak...` should be fine

Done. I was previously following Angular conventions. :)

mzgol:
The previous run was green: https://travis-ci.org/karma-runner/karma/builds/79488599 so hopefully nothing unexpected happens this time. :)

mzgol:
It's green. :) Ready to land?

Dignifiedquire:
:octocat: is happy

KenanSulayman:
`Re iojs yes let's drop it, 4.0 has most of it anyway.` - nodejs 4 *is* iojs 4. In fact, node-legacy was moved into a *archive* repo (https://github.com/nodejs/node-v0.x-archive) and the iojs repo was renamed to node (https://github.com/nodejs/node). That is, I would support iojs 2.5 for a few months (at least we're using it in production until we hit 4.0 with the next major deployment); 3.0 is ABI identical to 4.0 with the difference that v8 4.5 introduced arrow functions.

mzgol:
@KenanSulayman io.js 2.x was already not tested so my PR didn't change much in this regard. :) It's also not supported by upstream anymore AFAIK so I'd get off it ASAP.

mzgol:
(anyway, as I said, I didn't remove io.js 2 here since it wasn't here in the first place so if you'd like it back, you should open a separate issue)

PR #1587, opened 2015-09-21
fix(reporter): preserve base/absolute word in error

chrisirhc:
Previously, any error log message containing the words `absolute` or `base` would have those words munged by the error formatter.

Contains broken tests, and fix.


Dignifiedquire:
Thanks :octocat:

PR #1592, opened 2015-09-24
feat(config): add restartOnFileChange option

drgould:
When autoWatch=true and restartOnFileChange=true, this will "complete"
the current run immediately then schedule a new one as before. This is
most useful in gulp/grunt/etc. tasks that auto-build/test/reload to reduce
the time-to-feedback when making many changes in quick succession.

### DIFF
--- a/client/karma.js
+++ b/client/karma.js
@@ -220,6 +220,9 @@ var Karma = function (socket, iframe, opener, navigator, location) {
       window.console.clear()
     }
   })
+  socket.on('stop', function () {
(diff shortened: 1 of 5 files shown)


drgould:
I signed it!

Dignifiedquire:
Thanks, will review as soon as I can

Dignifiedquire:
Thanks :octocat:

Dignifiedquire:
Published as `0.13.11`

PR #1618, opened 2015-10-12
feat(reporter): Replace memoizee library with a trivial solution.

DavidSouther:
The reporter functionality was pulling the entire memoizee library with all its dependencies for a simple memoization solution, and using almost none of the functionality. This seems like not the most critically performant codepath. This PR replaces that with a good-case solution that lowers the dependency footprint.

### DIFF
--- a/lib/reporter.js
+++ b/lib/reporter.js
@@ -3,7 +3,6 @@ var log = require('./logger').create('reporter')
 var MultiReporter = require('./reporters/multi')
 var baseReporterDecoratorFactory = require('./reporters/base').decoratorFactory
 var SourceMapConsumer = require('source-map').SourceMapConsumer
-var memoizeWeak = require('memoizee/weak')
 
 var createErrorFormatter = function (basePath, emitter, SourceMapConsumer) {
   var lastServedFiles = []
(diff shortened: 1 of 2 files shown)


GitCop:

 * Commit: 0151c8342e67c1ac4831c5a54efe06d3bac0691e

---

juliemr:
Looks good to me. @zzo for final review.

zzo:
please fix the (trivial) travis lint failure

DavidSouther:
> please fix the (trivial) travis lint failure

Done.

maksimr:
@DavidSouther I dont think it's correct fix. 
Property `sourceMap` may be object, see [test](https://github.com/karma-runner/karma/blob/master/test/unit/reporter.spec.js#L117). Current solution will return incorrect source map in this case.
Solution with `weakMap` allow cache object and clean cache when file with source map was removed.

Dignifiedquire:
Agreed, there is a reason a proper cache was used here instead of this simple solution. Also the weakMap solution should be much more memory efficient than this. Unless there is a good reason please revert this.

Dignifiedquire:
@DavidSouther I really don't mind a couple of files of code, but the question is the actual memory and performance footprint lowered by this? If you are concerned with file size you can implement a version using a weakMap instead, but please not using a simple object.

zzo:
I believe the concern here is the dependency tree is large for this one module - I'll revert until we can find a better solution

Dignifiedquire:
@zzo which I appreciate, but it shouldn't come at the cost of actual runtime behaviour and performance.

zzo:
agreed thanks for weighing in

DavidSouther:
@zzo @Dignifiedquire @maksimr Correct, I wanted this to remove the rather large memoizee dependency that's used in this one internal place.

I totally misread the tests, and thought only the string sourcemap contents were being used here. Because engines is still 0.10 || 0.12, WeakMap would be inappropriate. This seems like not the most performance critical code - it will be called once per error in testing output. What do you all think about using the string form of the sourcemap? `sourceMap.file || sourceMap.contents || (typeof sourceMap == 'string' ? sourceMap : JSON.Stringify(sourceMap)`.

Dignifiedquire:
We are already including a shim for `weakMap`, [core-js](https://github.com/zloirock/core-js) (https://github.com/karma-runner/karma/blob/master/package.json#L233) so I suggest using that.

DavidSouther:
Ah, perfect! Should I re-open this PR, or start a new PR?

Dignifiedquire:
Let's make it a new PR to have a clean slate (and I think Github gets confused otherwise)

PR #1619, opened 2015-10-12
feat(web-server): allow injection of custom middleware.

### LINKED ISSUE #1612: Ability to inject custom middleware
We're trying to switch from test'em to karma, and one big issue is the inability to inject custom middleware into the karma test server.

Here is where test'em allows middleware injection via a configuration option:
https://github.com/airportyh/testem/blob/e70cbd7f67444b1a9070102d58c6789d39946529/lib/server/index.js#L149

The karma webserver has a mechanism that's far more obtuse and restrictive (also, the comments make it seem like it's on the chopping block):
https://github.com/karma-runner/karma/blob/master/lib/web-server.js#L65

Is there any plans to allow something like this? If there is, I could whip something up.

shanear:
Took a crack at a simple way of adding custom middleware.

Closes #1612

### DIFF
--- a/docs/config/01-configuration-file.md
+++ b/docs/config/01-configuration-file.md
@@ -280,6 +280,35 @@ httpsServerOptions: {
 **Description:** A list of log appenders to be used. See the documentation for [log4js] for more information.
 
 
+## middleware
+**Type:** Array
+
+**Default:** `[]`
+
+**Description:** List of names of additional middleware you want the karma server to use. Middleware will be used in the order listed.
+
+You must have installed the middleware via a plugin/framework (either inline or via NPM). Additional information can be found in [plugins].
+
+The plugin must provide an express/connect middleware function (details about this can be found in [the Express docs](http://expressjs.com/guide/using-middleware.html). An example of custom inline middleware is shown below.
+
+**Example:**
+```javascript
+var CustomMiddlewareFactory = function (config) {
+  return function (request, response, /* next */) {
+    response.writeHead(200)
+    return response.end("content!")
+  }
+}
+```
+
+```javascript
+middleware: ['custom']
+plugins: [
+  {'middleware:custom': ['factory', CustomMiddlewareFactory]}
+  ...
+]
+```
+
 ## plugins
 **Type:** Array
 
--- a/lib/web-server.js
+++ b/lib/web-server.js
@@ -63,9 +63,16 @@ var createWebServer = function (injector, emitter, fileList) {
     // TODO(vojta): remove, this is only here because of karma-dart
     // we need a better way of custom handlers
     .use(injector.invoke(createCustomHandler))
-    .use(function (request, response) {
-      common.serve404(response, request.url)
+
+  if (config.middleware) {
+    config.middleware.forEach(function (middleware) {
+      handler.use(injector.get('middleware:' + middleware))
     })
+  }
+
+  handler.use(function (request, response) {
+    common.serve404(response, request.url)
+  })
 
   var serverClass = http
   var serverArguments = [handler]
--- a/test/unit/web-server.spec.js
+++ b/test/unit/web-server.spec.js
@@ -32,7 +32,7 @@ describe('web-server', () => {
   var m = mocks.loadFile(__dirname + '/../../lib/web-server.js', _mocks, _globals)
 
   var customFileHandlers = server = emitter = null
-
+  var middlewareActive = false
   var servedFiles = (files) => {
     emitter.emit('file_list_modified', {included: [], served: files})
   }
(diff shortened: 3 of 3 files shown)


maksimr:
@shanear Thanks! 

[Knowledge about `connect`](https://github.com/karma-runner/karma/pull/1619/files#diff-f3dbde1a03c9aaa5b7dc6a7457830109R49) confuse me.
What do you think about create `middleware` as `launcher` or `framework`?

```js
var DartMiddlewareFactory = function(config) {
return function(request, response, next) {
};
};

module.exports = {
  'middleware:Dart': ['type', DartMiddlewareFactory]
};
```

example karma.config.js

```json
...
middleware: ["Dart"]
...
```

P.S. I use name `middleware` as example.

shanear:
Interesting

It seems from the docs that launchers and frameworks are plugins, and all plugins are npm modules.
<http://karma-runner.github.io/0.13/dev/plugins.html>

Is that correct? It would be nice to be able to add project specific middleware without having to create a separate npm module for it.

Again this is helpful for working with less modern, enterprise-style api's.

maksimr:
>It would be nice to be able to add project specific middleware without having to create a separate npm module for it.

@shanear you can inline plugin:

```js
...
plugins: ['karma-*', {'middleware:InlineMiddleware': inlineMiddlewareFactory}]
middliware: ['inlineMiddleware']
...
```

shanear:
@maksimr great, i'll give that a shot.

shanear:
@maksimr how does this look?

maksimr:
@shanear Thanks! Looks good for me :+1: 

@zzo, @Dignifiedquire what do you think about this feature?

Dignifiedquire:
I think this is a nice addition, but please add some documentation with example and squash your commits into one.

shanear:
@Dignifiedquire cool! These docs look ok?

If so, i'll squash everything up.

docs/config/01-configuration-file.md
```diff
@@ -280,6 +280,34 @@ httpsServerOptions: {
...
 
 
+## middleware
+**Type:** Array
+
+**Default:** `[]`
+
+**Description:** List of names of additional middleware you want the karma server to use.
+
+You must have installed the middleware via a plugin/framework (either inline or via NPM). An example of custom inline middleware is shown below. Additional information can be found in [plugins].
```
Dignifiedquire: It might be good to mention that this is express/connect middleware, and have a link to their docs on how to write one.

Dignifiedquire: Also it would be good to note that the middleware is used in the order they are included, as the order can matter if I remember correctly.

shanear:
@Dignifiedquire i agree with your notes. Updated.

Let me know what you think.

zzo:
this looks very interesting - middleware authors need to ensure they call 'next' if they don't want to handle a request - def. an advanced feature.  What are all of the parameters the factory function gets - more than just 'config'?  The karma-dart factory function gets:

    function (emitter, /* config.files */ files, /* config.basePath */ basePath,v/* config.karmaDartImports */ customImports, /* config.hostname */ hostname, /* config.port */ port, logger, customFileHandlers, /* config.dart2js */ dart2js)

shanear:
@zzo Yeah, it's def advanced, but opens up a lot of flexibility in how Karma can be used (IE: we're using it to test our app's integration with a not-very-modern enterprise API). Plus many people working with node seem to be familiar with how connect/express work.

The parameters are provided from the DI system.

Dignifiedquire:
@shanear the ordering note is still missing, other than that I think it looks good.

@zzo yes if you provide a custom plugin you can specify the parameters you want using di, which should be explained in the plugins docs.

zzo:
got it thanks - yes I think this is a great feature and ideally the karma-dart (and maybe other plugins?) should be transitioned to to use this ideally :)

shanear:
@Dignifiedquire I added a sentence about the ordering.

"Middleware will be used in the order listed."

Dignifiedquire:
@zzo yes it might make a lot of sense to use this in karma-dart from what I remember.
@shanear sorry I can't read, go ahead and squash so we can ship this :)

Dignifiedquire:
Thanks a lot :octocat:

shanear:
awesome :metal:

thanks for all of y'alls help!

Dignifiedquire:
Published as `0.13.11`

zzo:
mary poppins?  I just looked at it the other day and it's up and running from what I could tell

Dignifiedquire:
@zzo no all good old fashioned hand work

Dignifiedquire:
@zzo but all automated, just need to run `grunt release` and I'm done :)

PR #1622, opened 2015-10-13
feat(reporter): Replace way-too-big memoizee with a trivial solution.

DavidSouther:
The reporter functionality was pulling the entire memoizee library with all its
dependencies for a simple memoization solution, and using almost none of the
functionality. This seems like not the most critically performant codepath. This
PR replaces that with a good-case solution that lowers the dependency footprint.

Version 2: WeakMap.


DavidSouther:
@zzo  @maksimr @Dignifiedquire Round 2, continuing from #1618

Dignifiedquire:
LGTM

maksimr:
LGTM

Dignifiedquire:
Thanks :octocat:

Dignifiedquire:
Published as `0.13.11`

PR #1623, opened 2015-10-13
docs(plugin): fix typo with inline plugin example.

shanear:
fix a small typo in the plugin docs.


Dignifiedquire:
Thanks :octocat:

PR #1624, opened 2015-10-13
fix(proxy): Replace CreatyProxyServer

DarthCharles:
It seems that some of the unit tests on proxy.spec.js are failing because they try to call this method on proxy.js on line 42:

```javascript
    var proxy = httpProxy.createProxyServer({
      target: {
        host: hostname,
        port: port,
        https: https
      },
      xfwd: true,
      secure: config.proxyValidateSSL
    })
```

it seems that the http-proxy.js file that karma is using doesn't include the createProxySever method but provides the createServer method. If I change it to httpProxy.createServer all tests seems to be working.

### DIFF
--- a/lib/middleware/proxy.js
+++ b/lib/middleware/proxy.js
@@ -39,7 +39,7 @@ var parseProxyConfig = function (proxies, config) {
       (proxyDetails.protocol === 'https:' ? '443' : '80')
     var https = proxyDetails.protocol === 'https:'
 
-    var proxy = httpProxy.createProxyServer({
+    var proxy = httpProxy.createServer({
       target: {
         host: hostname,
         port: port,


maksimr:
@DarthCharles It's strange because **http-proxy** should export [createProxyServer](https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy.js#L26-L28) method.

What problem you solve?

Thanks

DarthCharles:
Yeah @maksimr, but the tests seems to be failing anyway. Can you please run the tests with grunt test:unit to see what you get?

maksimr:
@DarthCharles in master all work fine

DarthCharles:
How weird, I re-npm install and now it seems to be working. Sorry for bothering :fearful: and thanks for you time !

maksimr:
@DarthCharles no problems

PR #1664, opened 2015-10-27
fix(karma): fix included:false pattern being ignored

### LINKED ISSUE #1530: Karma 0.13 ignores included:false pattern
Hi. In one of my projects Karma started to include files marked with "included:false" in the karma.conf.js after I upgraded to Karma 0.13.3. Switching back to 0.12.37 resolves the issue so either there was a intended behavioral change (either undocumented or breaking some wrong assumptions on my side) or a regression. I reproduced the issue on Windows 10 as well as Ubuntu 15.04.

The karma.conf.js looks something like this:
```js
  config.set({
    basePath: '../',
    ...
    files: [
      {pattern: 'src/scripts/backend/worker.js', included: false}, // Web-worker script with entry point
      'src/scripts/**/*.js',
      'test/unit/test_*.js'
    ],
    ...
  });
};
```
The worker.js script contains web-worker specific functionality and hence must not be executed directly.

Steps to reproduce:
* Download the reproducer from http://d0t.dbclan.de/dump/karma-included-false-ignored-in-0.13.zip and extract it somewhere
* Run ```npm install```
* Run ```npm test```
* Replace the package.json with package-working.json
* Re-run ```npm install```
* Re-run ```npm test```

**Expected:** As worker.js should not be included directly its contents should be ignored and the dummy test should pass in both runs of ```npm test```

**Actual:** For the first run with Karma 0.13.3 worker.js is included and Karma fails and complains about ```someWebworkerSpecificFunction``` in worker.js. The second run with Karma 0.12.37 works as expected.

As I said. I'm not sure if it's something I'm doing or not but the behavior definitely changed between those two versions.

raspo:
Closes #1530

### DIFF
--- a/lib/file-list.js
+++ b/lib/file-list.js
@@ -222,11 +222,20 @@ Object.defineProperty(List.prototype, 'files', {
       return from(self.buckets.get(p.pattern) || []).sort(byPath)
     })
 
-    var included = this._patterns.filter(function (pattern) {
-      return pattern.included
-    })
-    .map(function (p) {
-      return from(self.buckets.get(p.pattern) || []).sort(byPath)
+    // excluded is a map of excluded paths
+    var excluded = _.reduce(_.flatten(this._patterns.filter(function (pattern) {
+      return !pattern.included
+    }).map(function (p) {
+      return from(self.buckets.get(p.pattern) || [])
+    })), function (result, file) {
+      result[file.path] = true
+      return result
+    }, {})
+
+    var included = this._patterns.map(function (p) {
+      return from(self.buckets.get(p.pattern) || []).filter(function (pattern) {
+        return !excluded[pattern.path]
+      }).sort(byPath)
     })
 
     var uniqFlat = function (list) {
--- a/test/unit/file-list.spec.js
+++ b/test/unit/file-list.spec.js
@@ -21,6 +21,7 @@ function findFile (path, files) {
 
 var PATTERN_LIST = {
   '/some/*.js': ['/some/a.js', '/some/b.js'],
+  '/some/a.js': ['/some/a.js'],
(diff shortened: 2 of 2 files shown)


raspo:
I signed it!

Dignifiedquire:
Thanks a lot, could you change the commit message to something like this: `fix(file-list): ...`

Also it seems there is an issue with the tests now, please make sure they pass.

raspo:
@Dignifiedquire I updated the commit message. Can you help me figure out what is the problem with travis? I don't get any errors when I run tests locally with `grunt test`.

zzo:
oh heh heh  I was accidentally signed in as 'karmarunnerbot' when I made that previous comment - sorry it's me.

raspo:
When defining patterns all files are included by default and when you want to exclude some you have to specify `included: false`. It feels to me like the excluded patterns should take priority over the included ones. 
In either case, one of the two should have a higher priority and should probably be documented somewhere. Thoughts? @zzo @Dignifiedquire

zzo:
ya seems like if a file is specifically included (without using a regex) but it's being also excluded by an exclude regex the include should win in that case - at least that's how it must have been working before for that test to pass - or something like that :)

Dignifiedquire:
Sorry for not responding earlier. I agree withehat @zzo says, to keep the old behaviour includes should take precendence over excludes if the pattern is more specific. I'm happy to discuss chamges to that in the future but for now we need that behaviour nack so we don't break everyones code.

Dignifiedquire:
@raspo any chance you could look into finishing this?

Dignifiedquire:
@raspo ping

dignifiedquire:
@raspo ping?

dignifiedquire:
Solved in another PR

PR #1672, opened 2015-10-30
chore(travis): Test on Node 5, don't update npm in newer Nodes

mzgol:
Node 5 includes npm 3 by default so we don't have to update npm to make sure
it works on that version. On the other hand, most people don't upgrade npm
and just use the default included version, especially that it's very hard
to update npm on Windows. It's good to make sure Karma still works with npm 2,
especially that Node 4, current LTS will never upgrade to npm 3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,16 +4,28 @@ node_js:
   - "0.10"
   - "0.12"
   - "4"
+  - "5"
 
 env:
   global:
   - SAUCE_USERNAME: karmarunnerbot
   - secure: "bRVY+hYZwMf1SqVnMyZRJTLD0gN1hLx9/MwO8MM/qBiu3YNjXy49XElfMdzMKN6cZeKTmhcnjmZonbJuI1PQ2t+utGkyjnlVLJ/OlWptreKLzIlcbt4hrdPoTcjmUTwDWq9Ex9cVoYX8AzCasETttpczp3P+s3+vmOUj8z25JyU="
+  - CXX=g++-4.8
 
-# Make sure we have a new npm@3
+# Make sure we have a new npm on Node 0.10. npm >= 2 should work so by not updating them
+# here we have proper version coverage, especially that most people use the npm version
+# that's included in a Node version they use.
 before_install:
-  - npm install -g npm
+  - '[ "${TRAVIS_NODE_VERSION}" != "0.10" ] || npm install -g npm'
   - npm config set loglevel warn
+  - g++-4.8 --version
+
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
+      - g++-4.8
 
 before_script:
   - export DISPLAY=:99.0
--- a/lib/middleware/proxy.js
+++ b/lib/middleware/proxy.js
@@ -125,6 +125,6 @@ var createProxyHandler = function (proxies, urlRoot) {
   return middleware
 }
 
-exports.create = function (/* config */ config, /* config.proxies */ proxies) {
+exports.create = function (/* config */config, /* config.proxies */proxies) {
   return createProxyHandler(parseProxyConfig(proxies, config), config.urlRoot)
 }
--- a/package.json
+++ b/package.json
@@ -323,7 +323,7 @@
     "karma": "./bin/karma"
   },
   "engines": {
-    "node": "0.10 || 0.12 || 4"
+    "node": "0.10 || 0.12 || 4 || 5"
   },
   "version": "0.13.14",
   "license": "MIT"


mzgol:
Everything passed except Node 4. Not sure what's going on here...

Dignifiedquire:
Retriggered, lets see what happens

zzo:
I think what's going here is the build is failing because the log is too long - travis fails the build after 4mb of output - the node5 build has the same errors/warnings as the node4 build but does succeed.
I wonder if 'apt-get install build-essential' would fix this as it's complaining about a lack of c++ compiler.  or cranking down the npm loglevel to slient which wouldn't be very interesting if the build failed for some other reason :)

Dignifiedquire:
It might have something to do with `npm` being not updated anymore, as otherwise this is the same as what we are currently running.

zzo:
ok build-essential and g++ are already installed at their latest versions so that doesn't help.  trying loglevel silent.  the version of npm that comes with 4.2.1 (2.14.7) acts differently than npm with node5 (3.3.6) and latest npm (3.3.10)

mzgol:
I think they have to be in a new enough version if you want to compile something, the v8 used in Node>=4 can't be compiled via old gcc. Take a look at [these lines in .travis.yml of node-sass](https://github.com/sass/node-sass/blob/7680371a43dcf266ccaef3a542ebbadd33bef5a7/.travis.yml#L24-L30).

zzo:
welp got travis to pass with npm output set to silent: https://travis-ci.org/karma-runner/karma/builds/88365243
There's still a ton of output and it's still complaining about c++ (the problem is the gcc compiler is at version 199711L but [NAN](https://github.com/nodejs/nan/blob/master/nan.h) wants a c++ compiler at 201103L) however 199711L IS the latest version of gcc (at least on the travis VMs) so insanity ensues.
The deps it cannot install are optional but the amount of output is insane (as many others have complained about).  Even with 'silent' there's still a ton of output - in fact none of that error output is supressed at all as it's not coming from npm.
thoughts?

GitCop:

 * Commit: 94cb0e5974a028e242958957af539c796802b705
  * Your subject line is longer than 70 characters

---

zzo:
@mzgol unfortunately both 4.7 and 4.8 both report themselves as 199711

GitCop:

 * Commit: 754182ba1e1eb43fef6cc892dac6e414b2b3fbae
  * Your subject line is longer than 70 characters

---

zzo:
both g++ 4.7 and 4.8 report themselves as 199711 unfortunately unless my rusty C skills have completely rusted over...

mzgol:
node-sass compiles fine on Travis on Node 5: https://travis-ci.org/sass/node-sass/jobs/88302086. I'm checking what else do we need from their config to make it work.

mzgol:
I'm trying to install the proper gcc/g++ versions and export proper env variables to make the commands use it but `g++ --version` still prints 4.6.3... I see that [`node-sass` now declares its project language as `cpp` instead of `node_js`](https://github.com/sass/node-sass/commit/5ded6edbeb84ab41565794cb69f5dcab5672faac), I hope that's not necessary here. :O

zzo:
thanks there's also this: https://github.com/nodejs/nan/issues/506#issuecomment-152572679
so 4.8 should work but it needs that flag.  Not sure why that flag is not being supplied by gyp - i'll update that issue with more questions...

mzgol:
Hmm, I didn't add any flag and it compiled correctly this time: https://travis-ci.org/karma-runner/karma/jobs/88381644.

mzgol:
@zoo from the comment:
> You need to compile with -std=c++11 for __cplusplus >= 201103 but you get that for free when you target node.js >= 4.

So my PR in its current state should be fine. :)

mzgol:
I wonder if the gcc upgrade is needed, maybe g++ is enough?

zzo:
yes!  installing 4.8 and setting 'CXX=g++-4.8' is the magic - great!

mzgol:
OK, I've removed some excessive variables & packages, this should still work but let's see.

mzgol:
OK, I've tried with g++5 but the bundles are *huge* (https://travis-ci.org/karma-runner/karma/jobs/88392214), ~50 MB compared to ~8 MB for 4.8 so I reverted back to 4.8. The tests passed last time so it should be fine but we can wait to be 100% sure. :)

zzo:
agreed 5.0 unnecessary - 4.8 is fine thanks

Dignifiedquire:
Thanks guys for working through the compiler swamp :)

mzgol:
Haha, between my last 2 attempts a new ESLint version got released and the build now fails because of one excessive space. :dancers:

zzo:
classic - this was as good of a time of this to happen as any tho

mzgol:
The following code [in proxy.js](https://github.com/karma-runner/karma/blob/b1386194943547b4b58334e39d691d2f64009690/lib/middleware/proxy.js#L128-L130) is the problem:
```js
exports.create = function (/* config */ config, /* config.proxies */ proxies) {
  return createProxyHandler(parseProxyConfig(proxies, config), config.urlRoot)
}
```
Sounds like a regression in ESLint to me, I reported an issue: https://github.com/eslint/eslint/issues/4302.

We'll need to workaround it for now, though, I'll have to remove spaces between comments & variable names to make it pass.

mzgol:
Oops, I forgot to add `5` as an accepted Node version in engines. PR updated.

mzgol:
@zzo @Dignifiedquire All tests passed. This should be ready to go. :)

zzo:
spectacular lgtm

Dignifiedquire:
Thanks :octocat:

PR #1742, opened 2015-12-02
fix(test): closes #1741

### LINKED ISSUE #1741: Test of 'Expire' header fails in non-english locales
In  test UTF-8 used, while server returns localized date time with local timezone name in ASCII, so they not match.
Supertest, used in fixture is not very smart (for my test) - it not allow to check header with custom function (it start to check response body and so on) - or may be i'm missed.

Suggest - change string ZERO_DATE to RegExp that will get just mean part of datetime (before GMT)  because if it's correct - it's set well - we not require to test RFC encoder itself here - just that expire is set to really-past date.

Have pull request

comdiv:


Dignifiedquire:
Ref #1741

Dignifiedquire:
Thanks, can you add the `Closes #1741` to the end of the commit message please.

comdiv:
Rewrited

Dignifiedquire:
Thanks :octocat:

PR #1745, opened 2015-12-04
fix(middleware): Correct spelling of middleware logger name

danielcompton:
The `middleware:source-files` logger was spelled incorrectly.


Dignifiedquire:
Thanks :octocat:

PR #1762, opened 2015-12-21
style(static/context.html): update comment wording

jurko-gospodnetic:
Updated comment wording that confused me a bit while reading through code.


Dignifiedquire:
Thanks :octocat:

Issue #1067, opened 2014-05-18
--no-colors for run command


omouse:
I'm trying to do this:

    karma run /path/to/myconfig.js --no-colors --port 8080

And it keeps displaying ANSI colors. I'm trying to pipe the output from the `run` command into Emacs and other text editors and I really really don't want to have to add extra code that strips out the ANSI colors when the `run` command really should be accepting the `--no-colors` argument just like the `start` command does.

I've tried setting the option in my config with `color:false` but that apparently only applies to the first running instance of karma, not to any subsequent calls to the karma server.

budde377:
@Dignifiedquire I believe that we can close this issue :)

Dignifiedquire:
Yes :)

PR #1780, opened 2016-01-05
feat(logging): Add colors and log-level options to run-command

budde377:
Add colors and log-level arguments to run argument.
 Refactor log-setup functions for server and init.
 Correct bug in server where log-level was ignored before `parseConfig`

Closing #1067


budde377:
Nah, this can't be right.

PR #1791, opened 2016-01-07
Feature adding no colors to run command

budde377:
This will solve #1067 but also make some quite invasive changes. Including changing the constructor of the std. reporters. Since these should be generated with injection (according to [this todo](https://github.com/karma-runner/karma/blob/master/lib/launchers/base.js#L22)) this might not be such a big change after all. 

Furthermore this allows setting the `color` properties on an adapter for deciding if a given reporter should write to it.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -185,6 +185,9 @@ var describeRun = function () {
     .describe('fail-on-empty-test-suite', 'Fail on empty test suite.')
     .describe('no-fail-on-empty-test-suite', 'Do not fail on empty test suite.')
     .describe('help', 'Print usage.')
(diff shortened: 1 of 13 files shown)


Dignifiedquire:
Thanks, could please try either cherry-picking your commits into a new branch, or rebasing this one onto master as there seem to be some conflicts at the moment and it's hard to review like this.

budde377:
Here you go!

Dignifiedquire:
This will need some merging with master please

budde377:
Yes, will do this afternoon.

budde377:
Well, its afternoon somewhere. Here you go!

Dignifiedquire:
Thanks :octocat: you rock!

PR #1796, opened 2016-01-11
Feature do not fail when the test set is empty

budde377:
This feature allows the user to specify if running the runner or server (singlerun) should return non-zero exit code when executing an empty test-suite. The field `failOnEmptyTestSuite : boolean` is added to the configuration file with default value `true`.

**The problem:** If the option is set to `false` and an empty test-suite is executed then the program should exit with exit-code `0` and display a warning. 

**The solution:** Can be divided into two sub-problems:

  1. Adding the feature to the server for single-run executions
  2. Adding the feature to the runner.

The first case is relatively simple: Besides adding the above option to the configuration file, only an extra check needs to be added at `lib/server.js:254`. 

The second case however is a bit more tricky:

Sending the configuration from the runner to the server would allow the server to respond with the the appropriate exit-code. However, the runner would not know if the code was due to a successful execution or execution of an empty test-suite. Thus, the runner is not able to display a warning correctly. 

Alternatively one could reserve an exit-code to distinct between these cases, e.g. 

* `0` = success
* `1` = error
* `2` = empty test suite

and thus not having to transmit the configuration to the server, letting the logger decide on the appropriate exit-code and warning. 

Lastly, one could change the server response from: `...\x1FEXIT<exit-code>` to `...\x1FEXIT<non-empty-indicator><exit-code>`, where the `<non-empty-indicator>` would be `0` if the test-suite was empty and 1 otherwise, allowing the runner to act accordingly. This solution imposes no restrictions on the exit-code and does not require sending additional configuration from the runner to the server.

Here I went with the last option.

### DIFF
--- a/docs/config/01-configuration-file.md
+++ b/docs/config/01-configuration-file.md
@@ -224,6 +224,15 @@ customHeaders: [{
 
 **Description:** List of files/patterns to exclude from loaded files.
 
+## failOnEmptyTestSuite
+**Type:** Boolean
+
+**Default:** `true`
+
+**CLI:** `--fail-on-empty-test-suite`, `--no-fail-on-empty-test-suite`
+
+**Description:** Enable or disable failure on running empty test-suites. If disabled the program
+will return exit-code `0` and display a warning.
 
 ## files
 **Type:** Array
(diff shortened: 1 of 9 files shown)


Dignifiedquire:
Thanks a lot :) Will review tomorrow

Dignifiedquire:
Nice work :) looking all good to me. The only thing missing is some documentation in docs/config.

budde377:
Done :)

Dignifiedquire:
Thanks :octocat:

PR #1806, opened 2016-01-14
Feature allow custom browser names

budde377:
Custom browser-names can now be defined as described in the docs. Furthermore, as a side-effect, browser-names can also be defined when testing manually. This is done by adding the argument `displayName=<displayName>` to the query-string.


Dignifiedquire:
Thanks, you're on a roll :)

Dignifiedquire:
Could you add a e2e test for this please, so we are sure this works as intended?

budde377:
Yes! Will do this evening.

budde377:
Done

Dignifiedquire:
Thanks for all the great work :octocat:

PR #1807, opened 2016-01-14
fix: Switch all requires from fs to graceful-fs

### LINKED ISSUE #544: Make sure to not hit open file descriptor limit
The default limit on Moutain Lion is 256 and sometimes people get `EMFILE` error.

Let's make sure we are efficient.
Preprocessors should use in-memory cache.
Web server might do in-memory caching as well, considering number of captured browsers (eg. if there is only one browser, in-memory caching does not make any sense).

testower:
Took the liberty to provide a fix here. I'm still hitting EMFILE in certain situations. Could it be due to launchers? I can definitely run my tests with a much lower ulimit -n now though.

Partially fixes #544

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -1,6 +1,6 @@
 var path = require('path')
 var optimist = require('optimist')
-var fs = require('fs')
+var fs = require('graceful-fs')
 
 var Server = require('./server')
 var helper = require('./helper')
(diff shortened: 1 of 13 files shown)


testower:
I signed it!

Dignifiedquire:
Thanks a lot :octocat:

PR #1815, opened 2016-01-21
feat(launcher): Enable specification of retry-limit

### LINKED ISSUE #1126: Allow configuration for retryLimit
It would be really helpful to allow a config param in `karma.conf.js` for retryLimit for browser crashes.

It would avoid us to wrap the `grunt` or `karma` call in a command like:
```bash
$ for i in `seq 1 3`; do grunt karma:ci; test $? -ne 3 && break; done
```
command.

I don't see any available configuration in http://karma-runner.github.io/0.12/config/configuration-file.html, but I do see it in the code a possibility to pass it: https://github.com/karma-runner/karma/blob/9e0e3854853dc04dd273dd0c17878d76de24fb70/lib/launchers/retry.js#L3

Could we make this setting available in config file?
Thanks!

budde377:
Allow specification of retry limit in configuration file with the `retryLimit` property. 
I also reordered the entries in the documentation in alphabetical order.

Closes #1126

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -265,6 +265,7 @@ var Config = function () {
   this.browserNoActivityTimeout = 10000
   this.concurrency = Infinity
   this.failOnEmptyTestSuite = true
+  this.retryLimit = 2
 }
 
 var CONFIG_SYNTAX_HELP = '  module.exports = function(config) {\n' +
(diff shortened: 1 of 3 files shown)


lib/config.js
```diff
@@ -265,6 +265,7 @@ var Config = function () {
   this.browserNoActivityTimeout = 10000
   this.concurrency = Infinity
   this.failOnEmptyTestSuite = true
+  this.retryLimit = 1
```
Dignifiedquire: This should be `2` right?

budde377: Yes

Dignifiedquire:
This looks good to me, could you squash the last commit into the first please?

budde377:
Done

Dignifiedquire:
Thanks :octocat:

PR #1818, opened 2016-01-22
Try fixing IE

dignifiedquire:


test/client/stringify.spec.js
```diff
@@ -28,59 +27,62 @@ describe('stringify', function () {
...
     div.innerHTML = 'some <span>text</span>'
-    expect(stringify(div)).to.be.eql('<div>some <span>text</span></div>')
+    assert.deepEqual(stringify(div), '<div>some <span>text</span></div>')
   })
 
   it('should serialize DOMParser objects', function () {
-    var parser = new DOMParser()
-    var doc = parser.parseFromString('<test></test>', 'application/xml')
-    expect(stringify(doc)).to.be.eql('<test></test>')
+    if (DOMParser) {
```
mgol: This will throw if `DOMParser` is not defined so this check doesn't work. You need to check for `typeof` being `!== 'undefined'`.

test/client/stringify.spec.js
```diff
@@ -28,59 +27,62 @@ describe('stringify', function () {
...
+      var parser = new DOMParser()
+      var doc = parser.parseFromString('<test></test>', 'application/xml')
+      assert.deepEqual(stringify(doc), '<test></test>')
+    }
   })
 
   it('should serialize across iframes', function () {
     var div = document.createElement('div')
-    expect(__karma__.stringify(div)).to.be.eql('<div></div>')
+    assert.deepEqual(__karma__.stringify(div).trim(), '<div></div>')
```
mgol: Apply `toLowerCase()` to that; old IEs convert tag names to upper case, causing a failure here.

mgol: You need such a change in one of upper tests as well (the `should serialize html` one)

Dignifiedquire:
Thanks @mgol! any ideas on why IE 6 completely refuses to connect?

mgol:
It seems mocha is broken in IE<9? Could you try to update it? Also, it seems they [still support IE8](https://github.com/mochajs/mocha/issues/2078#issuecomment-175697735) but I'm not sure if they care about IE 6 & 7... What can we do if that's the case?

dignifiedquire:
@mgol hmm, I will update and we will have to see, I'm already including polyfills for older browsers so that might be enough to get things working with mocha. But I expect to see a big drop in library support for IE 6 - 9 now that Microsoft officially dropped them all from their support list.

mgol:
> But I expect to see a big drop in library support for IE 6 - 9 now that Microsoft officially dropped them all from their support list.

Probable but considering that Angular 1 & 2 & Ember still support IE 9, at least IE 9+ will be necessary to support by Karma for some time. Unfortunately, test runners need to be a little more conservative in dropping older browsers than libraries/frameworks that are using them. :)

dignifiedquire:
> Unfortunately, test runners need to be a little more conservative in dropping older browsers than libraries/frameworks that are using them. :)

Yeah I know, that's why I'm trying to fix this. Just saying that libraries like mocha might not bother to do so.

PixnBits:
FWIW Vista is IE9 and isn't EoL until 2017
Some embedded Windows still lives and are IE8 :'(
https://support.microsoft.com/en-us/lifecycle#gp/Microsoft-Internet-Explorer

mgol:
This is a little OT but I've made a table compiled from https://support.microsoft.com/en-us/lifecycle#gp/Microsoft-Internet-Explorer & https://support.microsoft.com/en-us/gp/lifeselectindex:

Desktop:
* IE 9: 2017-04-11 (Windows Vista)
* IE 11:
    * 2023-01-10 (Windows 8.1)
    * ????-??-?? (Windows 10)

Others:
* IE 7: 2016-04-12 (Windows Embedded for Point of Service (WEPOS))
* IE 8:
    * 2017-01-10* (Windows Thin PC (Embedded))
    * 2019-01-08 (Windows Embedded Standard 2009)
    * 2019-04-09 (Windows Embedded POSReady 2009)
* IE 9: 2020-01-14 (Windows Server 2008)
* IE 10:
    * 2023-01-10 (Windows Server 2012)
    * 2023-07-11 (Windows Embedded 8 Standard)
* IE 11:
    * 2020-01-14 (Windows Server 2008 R2)
    * 2020-10-13 (Windows Embedded Standard 7)
    * 2021-10-12 (Windows Embedded POSReady 7)
    * 2021-10-12* (Windows Thin PC (Embedded))
    * 2023-01-10 (Windows Server 2012 R2)
    * 2023-01-10 (Windows Embedded 8 + Windows 8.1 Industry Update)
    * ????-??-?? (Windows Server 2016)

dignifiedquire:
thanks @mgol so the only one that is fully unsupported right now is IE 6 as far as I understand. 

If I can get these tests to pass we should be back with support for 7 and above would that be enough for your use cases?

mgol:
@dignifiedquire [Sizzle](https://github.com/jquery/sizzle/) (jQuery selector engine) still supports IE 6 but we'll be dropping at least this version soon so if we made it work in IE 7+ we'd be happy. :)

dignifiedquire:
@mgol I have things passing in IE7 on my local VM saucelabs is still a bit whiny but we'll get there, mind taking this for a spin?

Issue #1343, opened 2015-03-10
Remove null characters from terminal output


mntnoe:
In `runner.js:parseExitCode()`, the exit code in the message is replaced with null characters (\0x00) as a side effect. The message is then sent to stdout. This results in squares (□□□□□□) being displayed in some terminals, like WebStorm's terminal. Is it possible to output only the message without the trailing null characters?

PR #1839, opened 2016-01-26
fix(runner): Remove null characters from terminal output

budde377:
This slices the buffer rather than modifying it. The buffer slice is returned, with the exit code, as an object:

```js
{
  exitCode: int,
  buffer: Buffer
}
```


dignifiedquire:
What is the exact difference for the user from doing this/what exactly is fixed here?

budde377:
Basically this fixes the issue: #1343, where a user reported that the `\x00` characters printed in the console would appear as symbols. This fix slices the tail off the buffer rather than replacing it with the `\x00` bytes.

dignifiedquire:
@budde377 thanks, just saw you mentioned it in the commit message -.-

Issue #136, opened 2012-10-21
Add possibility to stop a karma server


nidi3:
When you start testacular with singleRun=false from within a script (Eg. build script) there is no way to stop it again (apart from killing the process manually).

Someting like

`node testacular stop 9876`

would be nice.

budde377:
@dignifiedquire We can close this, right?

dignifiedquire:
Yes we can thanks to you :)

joeflack4:
Still staying open for me

```bash
(datalab) joe:datalab joeflack4$ ./node_modules/karma/bin/karma start --single-run
02 10 2018 12:40:51.691:WARN [watcher]: Pattern "/Users/joeflack4/projects/datalab/test/fixtures/**/*" does not match any file.
ℹ ｢wdm｣: Hash: 1af0291e2b7daf4c1038
Version: webpack 3.10.0
Time: 61ms
ℹ ｢wdm｣: Compiled successfully.
02 10 2018 12:40:51.712:WARN [watcher]: Pattern "/Users/joeflack4/projects/datalab/test/fixtures/**/*" does not match any file.
ℹ ｢wdm｣: Compiling...
ℹ ｢wdm｣: wait until bundle finished: noop
ℹ ｢wdm｣: wait until bundle finished: noop
ℹ ｢wdm｣: wait until bundle finished: noop
ℹ ｢wdm｣: wait until bundle finished: noop
ℹ ｢wdm｣: Hash: 80d465b0bcc75b9fc8d3
Version: webpack 3.10.0
Time: 1324ms
                               Asset     Size  Chunks                    Chunk Names
      test/javascripts/chart.test.js  1.01 MB       0  [emitted]  [big]  test/javascripts/chart.test
    test/javascripts/network.test.js   445 kB       1  [emitted]  [big]  test/javascripts/network.test
  test/javascripts/url-parse.test.js   380 kB       2  [emitted]  [big]  test/javascripts/url-parse.test
test/javascripts/definitions.test.js   407 kB       3  [emitted]  [big]  test/javascripts/definitions.test
    test/javascripts/utility.test.js   408 kB       4  [emitted]  [big]  test/javascripts/utility.test
   [7] ./node_modules/babel-runtime/helpers/classCallCheck.js 208 bytes {0} {1} {2} {3} {4} [built]
  [24] ./source/javascripts/utility.js 5.58 kB {0} {1} {3} {4} [built]
[...]

PR #1854, opened 2016-01-31
Add possibility to stop a karma server

budde377:
This is a fix for issue #136. This: 

* Add detached mode using the `karma start --detached` command.
* Add middleware for stopping a server (detached or not).
* Described the detached option.

Any pointers on testing this or describing the new command (`karma stop`)?

### DIFF
--- a/docs/dev/04-public-api.md
+++ b/docs/dev/04-public-api.md
@@ -118,6 +118,22 @@ runner.run({port: 9876}, function(exitCode) {
 })
 ```
 
+## karma.stopper
+
+### **stopper.stop(options, [callback=process.exit])**
+
+This function will signal a running server to stop.  Equivalent of `karma stop`.  
+
+```javascript
+var stopper = require('karma').stopper
+runner.stop({port: 9876}, function(exitCode) {
+  if (exitCode === 0) {
+    console.log('Server stop as initiated')
+  }
+  process.exit(exitCode)
+})
+```
+
 ## Callback function notes
 
 - If there is an error, the error code will be provided as the second parameter to the error callback.
--- a/test/e2e/stop.feature
+++ b/test/e2e/stop.feature
@@ -1,0 +1,44 @@
+Feature: Stop karma
+  In order to use Karma
+  As a person who wants to write great tests
+  I want to be able to stop Karma.
+
+  Scenario: A server can't be stopped if it isn't running
+    When I stop Karma
+    Then it fails with like:
+      """
+      ERROR \[stopper\]: There is no server listening on port [0-9]+
+      """
+
+  Scenario: A server can be stopped
+    Given a configuration with:
+      """
+      files = ['basic/plus.js', 'basic/test.js'];
+      browsers = ['PhantomJS'];
+      plugins = [
+        'karma-jasmine',
+        'karma-phantomjs-launcher'
+      ];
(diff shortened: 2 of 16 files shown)


dignifiedquire:
Thanks for this. I would really like to see an e2e test making sure this works as expected.

dignifiedquire:
> Any pointers on testing this or describing the new command (karma stop)?

Take a look at the support files for the e2e tests, they are relatively straightforward javascript functions https://github.com/karma-runner/karma/blob/master/test/e2e/steps/core_steps.js#L43 
We would need to add one for `stop` to test this. For docs on cucumberjs see here: https://github.com/cucumber/cucumber-js

budde377:
Cheers, I'll have a look at this ASAP.

budde377:
Now testing:
* Stopping a server that isn't running
* Stopping a server that is running
    - Succesfull termination
    - Informative error message

test/e2e/stop.feature
```diff
@@ -0,0 +1,42 @@
+Feature: Stop karma
+  TODO write this
```
dignifiedquire: missing todo ;)

dignifiedquire:
This is pretty sweet, one thing that would be nice is to have the `stopper` be also available on the public api (http://karma-runner.github.io/0.13/dev/public-api.html) and have some docs on how to use it. This could be used then in things like task runners to make it easier to spawn long running processes.

budde377:
Yeah, I can see that! I'll add it.

budde377:
@dignifiedquire Here you go.

docs/dev/04-public-api.md
```diff
@@ -111,6 +111,22 @@ runner.run({port: 9876}, function(exitCode) {
 })
 ```
 
+## karma.stopper
+
+### **stopper.stop(options, [callback=process.exit])**
+
+This function will signal a running server to stop.  Equivalent of `karma stop`.
```
dignifiedquire: Will this work if `detached=false`?

budde377: The `detached` setting is currently only working when running the server from CLI, since it's basically just calling itself in a new detached process. 

So yes and no: The stop command will stop the server regardless of the `detached` setting and the server will not run detached when starting it programically. If this is a priority I'll certainly look into it.

dignifiedquire: We are doing sth very similar here: https://github.com/karma-runner/grunt-karma/blob/master/tasks/grunt-karma.js#L114-L134 so it would be nice if we could replace that with this new option, rather than duplication that effort

budde377: Cool, I'll look into it during the weekend.

budde377:
@dignifiedquire I have added the detached feature. In order to do so, I had to write the configuration as a file, hence the `tmp` dependency.

dignifiedquire:
@budde377 I think the e2e tests are not passing anymore :(

budde377:
@dignifiedquire, This should do the trick.

budde377:
A bit too fast there, I'll take a look at the failing node 0.10.

budde377:
@dignifiedquire Done

dignifiedquire:
Thanks @budde377 merging this. Let's hope this time around things go smoother :)

PR #1874, opened 2016-02-05
feat(server): add 'listening' event with port number

karol-f:
Sometimes, when you want to use karma port programatically (e.g. to send manually browser to karma localhost:<port> address), you face problem with not knowing exact port number.

Port number might be different from one set in config as sometimes you run more than one instances of karma. Below image show such situation:
Karma port changed

This PR adds 'listening' event to karma. After this event you are sure that you have correct karma port number.

You may use it like this:
```
let server = new Server(karmaConfig, done);
server.once('listening', function (e, port) {
    console.log(port); //port number from event
    console.log(server.get('config').port); // also correct value
});
```

### DIFF
--- a/test/unit/server.spec.js
+++ b/test/unit/server.spec.js
@@ -149,6 +149,19 @@ describe('server', () => {
       expect(mockConfig.port).to.be.equal(9877)
     })
 
+    it('should emit a listening event once server begin accepting connections', () => {
+      server._start(mockConfig, mockLauncher, null, mockFileList, mockWebServer, browserCollection, mockSocketServer, mockExecutor, doneSpy)
+
+      var listening = sinon.spy()
+      server.on('listening', listening)
+
+      expect(listening).not.to.have.been.called
+
+      fileListOnResolve()
+
+      expect(listening).to.have.been.calledWith(9876)
+    })
+
     it('should emit a browsers_ready event once all the browsers are captured', () => {
       server._start(mockConfig, mockLauncher, null, mockFileList, mockWebServer, browserCollection, mockSocketServer, mockExecutor, doneSpy)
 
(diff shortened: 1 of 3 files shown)


karol-f:
@googlebot I signed it!

dignifiedquire:
Thanks, could you please

* add docs about this in https://github.com/karma-runner/karma/blob/master/docs/dev/04-public-api.md
* add a test to make sure this actually works as expected

karol-f:
@dignifiedquire done :)

test/unit/server.spec.js
```diff
@@ -149,6 +149,21 @@ describe('server', () => {
...
+      server._start(mockConfig, mockLauncher, null, mockFileList, mockWebServer, browserCollection, mockSocketServer, mockExecutor, doneSpy)
+
+      var listening = sinon.spy()
+      server.on('listening', listening)
+
+      expect(listening).not.to.have.been.called
+
+      fileListOnResolve()
+
+      server.emit('listening', 9876)
```
dignifiedquire: This is not really testing anything is it? You should probably emit the event on the `mockWebServer` instead

karol-f:
I've changed test to pass when `mockWebServer.listen` callback is executed - so when server starts listening.

dignifiedquire:
Thanks :) looks good now, if you could squash your commits into one then I can merge it.

karol-f:
Done :)

dignifiedquire:
Thanks :octocat: Just waiting for travis to get green now

karol-f:
@dignifiedquire It's green

dignifiedquire:
@karol-f thanks :)

segrey:
@karol-f thanks for the change

PR #1899, opened 2016-02-13
feat(config): Add `preprocessorOrder` option to config for when prepr…

sjelin:
…ocessors

must be run in a specific order.

Some karma users at Google discovered karma-coverage needs to be run before
karma-googmodule-preprocessor, hence we need this feature.

This required an update to `karma-browserify`, which can be found here: https://github.com/nikku/karma-browserify/pull/168.

Since this version is not yet released, this will break travis.

### DIFF
--- a/lib/preprocessor.js
+++ b/lib/preprocessor.js
@@ -2,6 +2,7 @@ var fs = require('graceful-fs')
 var crypto = require('crypto')
 var mm = require('minimatch')
 var isBinaryFile = require('isbinaryfile')
+var combineLists = require('combine-lists')
 
 var log = require('./logger').create('preprocess')
 
@@ -64,9 +65,11 @@ var createPreprocessor = function (config, basePath, injector) {
     return p
   }
 
+  var allPreprocessors = []
   patterns.forEach(function (pattern) {
-    config[pattern].forEach(instantiatePreprocessor)
+    allPreprocessors = combineLists(allPreprocessors, config[pattern])
   })
+  allPreprocessors.forEach(instantiatePreprocessor)
 
   return function preprocess (file, done) {
     patterns = Object.keys(config)
@@ -81,35 +84,37 @@ var createPreprocessor = function (config, basePath, injector) {
           throw err
         }
 
-        var preprocessors = []
-        var nextPreprocessor = createNextProcessor(preprocessors, file, done)
-
+        var preprocessorNames = []
         for (var i = 0; i < patterns.length; i++) {
           if (mm(file.originalPath, patterns[i], {dot: true})) {
             if (thisFileIsBinary) {
               log.warn('Ignoring preprocessing (%s) %s because it is a binary file.',
                 config[patterns[i]].join(', '), file.originalPath)
             } else {
-              config[patterns[i]].forEach(function (name) {
-                var p = instances[name]
-                if (p == null) {
-                  p = instantiatePreprocessor(name)
-                }
-
-                if (p == null) {
-                  if (!alreadyDisplayedWarnings[name]) {
-                    alreadyDisplayedWarnings[name] = true
-                    log.warn('Failed to instantiate preprocessor %s', name)
-                  }
-                  return
-                }
-
-                instances[name] = p
-                preprocessors.push(p)
-              })
+              preprocessorNames = combineLists(preprocessorNames, config[patterns[i]])
             }
           }
         }
+
+        var preprocessors = []
+        var nextPreprocessor = createNextProcessor(preprocessors, file, done)
+        preprocessorNames.forEach(function (name) {
+          var p = instances[name]
+          if (p == null) {
+            p = instantiatePreprocessor(name)
+          }
+
+          if (p == null) {
+            if (!alreadyDisplayedWarnings[name]) {
+              alreadyDisplayedWarnings[name] = true
+              log.warn('Failed to instantiate preprocessor %s', name)
+            }
+            return
+          }
+
+          instances[name] = p
+          preprocessors.push(p)
+        })
 
         nextPreprocessor(null, thisFileIsBinary ? buffer : buffer.toString())
       })
--- a/package.json
+++ b/package.json
@@ -266,6 +266,7 @@
     "body-parser": "^1.12.4",
     "chokidar": "^1.4.1",
     "colors": "^1.1.0",
+    "combine-lists": "^1.0.0",
     "connect": "^3.3.5",
     "core-js": "^2.1.0",
     "di": "^0.0.1",
--- a/docs/config/04-preprocessors.md
+++ b/docs/config/04-preprocessors.md
@@ -93,3 +93,50 @@ return `false` and the preprocessor would not be executed on the CoffeeScript fi
 [custom plugins]: ../dev/plugins.html
 [plugins]: plugins.html
 [issue788]: https://github.com/karma-runner/karma/issues/788
+
+## Order of execution
+
+If a file matches only one key in the preprocessors config object, then karma
+will execute the preprocessors over that file in the order they are listed in
+the corresponding array.  So for instance if the config object is:
+
+```js
+preprocessors: {
+  '*.js': ['a', 'b']
+}
+```
+
+Then karma will execute `'a'` before executing `'b'`.
+
+If a file matches multiple keys, karma will do its best to execute the
+preprocessors in a reasonable order.  So if you have:
+
+```js
+preprocessors: {
+  '*.js': ['a', 'b'],
+  'a.*': ['b', 'c']
+}
+```
+
+then for `a.js`, karma will run `'a'` then `'b'` then `'c'`.  If two lists
+contradict eachother, like:
+
+```js
+preprocessors: {
+  '*.js': ['a', 'b'],
+  'a.*': ['b', 'a']
+}
+```
+
+then karma will arbitrarily pick one list to prioritize over the other.  In a
+case like:
+
+```js
+preprocessors: {
+  '*.js': ['a', 'b', 'c'],
+  'a.*': ['c', 'b', 'd']
+}
+```
+
+Then `'a'` will definitely be run first, `'d'` will definitely be run last, but
+it's arbitrarily if karma will run `'b'` before `'c'` or vice versa.
--- a/test/unit/preprocessor.spec.js
+++ b/test/unit/preprocessor.spec.js
@@ -271,4 +271,54 @@ describe('preprocessor', () => {
       done()
     })
   })
+
+  it('should merge lists of preprocessors', (done) => {
+    var callOrder = []
+    var fakePreprocessorA = sinon.spy((content, file, done) => {
+      callOrder.push('a')
+      done(null, content)
+    })
+    var fakePreprocessorB = sinon.spy((content, file, done) => {
+      callOrder.push('b')
+      done(null, content)
+    })
+    var fakePreprocessorC = sinon.spy((content, file, done) => {
+      callOrder.push('c')
+      done(null, content)
+    })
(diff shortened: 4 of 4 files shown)


package.json
```diff
@@ -355,7 +355,7 @@
   "engines": {
     "node": "0.10 || 0.12 || 4 || 5"
   },
-  "version": "0.13.21",
+  "version": "0.13.22",
```
dignifiedquire: Please remove this, version bumps are generated automatically on release

package.json
```diff
@@ -313,7 +313,7 @@
     "grunt-npm": "0.0.2",
     "jasmine-core": "^2.3.4",
     "json3": "^3.3.2",
-    "karma-browserify": "^5.0.1",
+    "karma-browserify": "^5.0.2",
```
dignifiedquire: Please fix this, as the travis is failing because this version is not released yet

nikku:
I'd like to put a different solution for this on the table: Allow `config.preprocessors` to be an array instead of an object. This way the processing order for individual preprocessors is clear from the start and no additional configuration property is needed.

Example JSON:

```json
{
  "preprocessors": [
    [ ".js", "googmodule" ],
    [ ".js", "coverage" ]
  ]
}
```

@sjelin Could you provide us with an example project + karma configuration that fails with the current version of karma?

package.json
nikku: The change @ karma-browserify won't be a patch version, too because it will break with all karma versions < than the new one.

sjelin: Ok, but travis will be failing either way because it doesn't work with the current version of browserify

sjelin:
@nikku we were having problems with a config file like:
```js
"preprocessors":  {
  "**/*.js": [ "googmodule" ],
  "**/client/**/*.js": [ "coverage" ]
},
```

Which caused `googmodule` to run before `coverage`, which in caused problems in the coverage information (not sure exactly why).


The problem with the `preprocessors` array option is that you'd run into problems combining multiple config files if one assumed `preprocessors` was an array and the other assumed `preprocessors` was an object.

package.json
sjelin: Done.  The problem now is that I'll have nothing to reference in the browserify change

package.json
sjelin: Done.  Presumably the browserify tests will still break though.

dignifiedquire:
@nikku have you tried doing sth like this:

```js
"preprocessors":  {
  "!(client)/*.js": [ "googmodule" ],
  "client/**/*.js": ["googmodule", "coverage" ]
},
```

You would need to explicitly define the top level path, but you wouldn't need to add any code to karma to get this working.

sjelin:
Yes, that would probably work in this instance.  But users may have a more complicated group of files they need coverage data for.  For instance:

```js
"preprocessors":  {
  "a/**/*.js": ["googmodule", "coverage" ],
  "**/b/*.js": ["googmodule", "coverage" ],
  "c*d/*.js": ["googmodule", "coverage" ],
  "<else *.js>": [ "googmodule" ]
},
```

In this case I have no idea what pattern goes in the `<else *.js>` case.  I'm also not sure what would happen to a file like `a/b/x.js`, which ends up getting `["googmodule", "coverage", "googmodule", "coverage"]` run on it.

Plus, while it's good that it requires no karma code change, it feels like a hack.  I want `googmodule` to be run everywhere, and it should be simple to write that in the config.

lib/preprocessor.js
```diff
@@ -36,7 +36,7 @@ var createNextProcessor = function (preprocessors, file, done) {
   }
 }
 
-var createPreprocessor = function (config, basePath, injector) {
+var createPreprocessor = function (config, basePath, ppOrder, injector) {
```
dignifiedquire: Can we avoid this? I'd rather not risk breaking other preprocessors

sjelin: I guess so.  We could have a custom property of `config` like `!!preprocessorOrder` which would never be needed to match a real file.

nikku:
@sjelin Ok, how does the following (switched order of preprocessor definitions) behave then?

```
"preprocessors":  {
  "**/client/**/*.js": [ "coverage" ],
  "**/*.js": [ "googmodule" ]
},
```

sjelin:
I mean, that works (sometimes at least), but it's not *supposed* to work.  It relies on `Object.keys` to return the keys in the same order they're written down in, but the spec explicitly says `Object.keys` returns the keys in an arbitrary order.

dignifiedquire:
@sjelin what happens if you do:

"preprocessors":  {
  "**/*.js": [ "googmodule" ],
  "**/client/**/*.js": ["googmodule", "coverage" ]
},

sjelin:
I think you mean:
```js
"preprocessors":  {
  "**/*.js": [ "googmodule" ],
  "**/client/**/*.js": ["coverage", "googmodule"]
},
```

since `coverage` is supposed to run first.

Currently there is no de-duplication in [`preprocessor.js`](https://github.com/karma-runner/karma/blob/master/lib/preprocessor.js), so it would run either `["googmodule", "coverage", "googmodule"]` or `["coverage", "googmodule", "googmodule"]`, depending on the order of the `Object.keys` array.  I could write some kind of intelligent de-duplication/ordering into `preprocessor.js`.  The advantage is that it wouldn't change the function's signature or introduce new properties into the config file.  The disadvantage is that it would just be a secret version of `preprocessorOrder`

dignifiedquire:
Yes that's what I meant. 

I think it would be a very good solution to introduce deduping here and use that order as it would be much more explicit and would solve the more complex use case where you want different orders of the same preprocessors in different files.

As long as we properly document this it wouldn't be a hidden feature.

sjelin:
Ok, I'm good with that.

nikku:
+1 if you get that properly explained to people, too.

zzo:
thanks sammy fix travis :)

sjelin:
Ok, I wrote it the other way.  I know what you're going to say: "Why is it so complicated?"  It turns out merging arrays, trying to keep their orders in tact as much as possible, isn't easy.  I'm more than open to similar solutions.

lib/preprocessor.js
```diff
@@ -36,6 +37,94 @@ var createNextProcessor = function (preprocessors, file, done) {
...
+   *   if(list2[i] is not in list1)
+   *     insert list2[i] into merged_list in the earliest place that creates the
+   *     minimum number of inversions
+   * return merged_list
+   *
+   * It can be proven that this gives you the two properties mentioned in the
+   * header comment above
+   *
+   * For those who care, this is very slow, and taking
+   * O(list2.length*list2.length*(list1.length + list2.length)) time.
```
dignifiedquire: :laughing: who would have guessed we find some big O notation in here one day

lib/preprocessor.js
```diff
@@ -36,6 +37,94 @@ var createNextProcessor = function (preprocessors, file, done) {
...
+   */
+  var merged = list1.slice() // The merged list to return
```
dignifiedquire: Please put comments on their own line

lib/preprocessor.js
```diff
@@ -36,6 +37,94 @@ var createNextProcessor = function (preprocessors, file, done) {
...
+  var mergedIndexes = _.invert(merged)
+  for (var i = 0; i < list2.length; i++) {
+    var elem = list2[i]
+    if (mergedIndexes[elem] === undefined) {
+      // Count the inversions for every possible insertion position
+      var inversionCnts, j, k
+      if (typeof Int32Array !== 'undefined') {
```
dignifiedquire: What benefit does this distinction give us?

lib/preprocessor.js
```diff
@@ -36,6 +37,94 @@ var createNextProcessor = function (preprocessors, file, done) {
...
+        inversionCnts = new Int32Array(merged.length + 1)
+      } else {
+        inversionCnts = []
+        for (j = 0; j <= merged.length; j++) {
+          inversionCnts[j] = 0
+        }
```
dignifiedquire: Could be simpler: 

```js
inversionCnts = _.fill(Array(merged.length), 0)
```
```

File: lib/preprocessor.js [hunk 8]
```diff
@@ -36,6 +37,94 @@ var createNextProcessor = function (preprocessors, file, done) {
...
+   * For those who care, this is very slow, and taking
+   * O(list2.length*list2.length*(list1.length + list2.length)) time.
+   */
+  var merged = list1.slice() // The merged list to return
+  var mergedIndexes = _.invert(merged)
+  for (var i = 0; i < list2.length; i++) {
+    var elem = list2[i]
+    if (mergedIndexes[elem] === undefined) {
+      // Count the inversions for every possible insertion position
+      var inversionCnts, j, k
```
dignifiedquire: Please use a separate line for each variable declaration

lib/preprocessor.js
```diff
@@ -36,6 +37,94 @@ var createNextProcessor = function (preprocessors, file, done) {
   }
 }
 
+/* Merge two lists, removing duplicates, and doing everything possible to
```
dignifiedquire: I suggest we put this into its own file or actually pull it into it's own repo. This is quite complex but very useful and might help others as well when we publish it to npm. Also that makes testing it very easy :)

lib/preprocessor.js
```diff
@@ -36,6 +37,94 @@ var createNextProcessor = function (preprocessors, file, done) {
...
+ * maintain the order of the two lists.
+ *
+ * This function guarantees that the order of list1 is preserved (that is, if x
```
dignifiedquire: Style: Please use `//` for comments to be consistent with the rest of the code

lib/preprocessor.js
```diff
@@ -81,36 +172,39 @@ var createPreprocessor = function (config, basePath, injector) {
...
-                    log.warn('Failed to instantiate preprocessor %s', name)
-                  }
-                  return
-                }
-
-                instances[name] = p
-                preprocessors.push(p)
-              })
+              preprocessorNames = mergeLists(preprocessorNames,
+                  config[patterns[i]])
```
dignifiedquire: This can be one line

dignifiedquire:
@sjelin nice work, didn't expect this to get that complicated though :D 

I left some style nitpicks but other than that I'm pretty happy with this solution

dignifiedquire:
Also the commit message needs to mention that this is a breaking change, as it might change the order of preprocessors for some expecting the old behaviour.

lib/preprocessor.js
sjelin: It's a lot faster.  It forces all the following arithmetic to be integer arithmetic instead of floating point.    It's the inner loop of a potential bottleneck, so I thought it might be a good idea.  That said, unless someone had like 15 processors running on the same file, it probably wouldn't be noticeable.

lib/preprocessor.js
sjelin: That makes sense.  I'll do that

lib/preprocessor.js
dignifiedquire: That makes sense, I suggest adding a comment on which node versions this is actually not available so we can remove it if and when we drop support for those versions.

sjelin:
All comments addressed!  Though tbh once I split the list merging into its own repo (https://github.com/sjelin/combine-lists), most of the comments were no longer relevant.

dignifiedquire:
Thanks, nearly there, looks like linting is unhappy now. (`grunt lint` locally)

sjelin:
Fixed the lint problems.  Looks like some of the e2e tests are failing.  Are they flaky though?  It seems pretty random

dignifiedquire:
@sjelin restarted them, yes they are super flaky atm sometimes travis gets so slow they time out :(

dignifiedquire:
Thanks :octocat:

Issue #1894, opened 2016-02-12
Karma 0.13.20 does not detect files when using webpack


ianks:
After upgrading to 0.13.20, none of my tests ran. Sorry I cannot provide too much debugging information, but I don't know where to start.

Here is my karma.conf.js

```js
var webpackConfig = require('./../config/webpack.config')
var path = require('path')

module.exports = function (config) {
  config.set({
    basePath: path.join(__dirname, '..'),
    autoWatch: true,
    logLevel: config.LOG_INFO,
    browsers: [ 'PhantomJS' ],
    files: [
      'http://cdn.auth0.com/js/lock-8.2.1.min.js',
      'test/index.js'
    ],
    port: 8080,
    captureTimeout: 60000,
    frameworks: [ 'phantomjs-shim', 'mocha', 'chai', 'sinon-chai' ],
    client: {
      mocha: {}
    },
    reporters: [ 'mocha' ],
    preprocessors: {
      'test/index.js': [ 'webpack', 'sourcemap' ]
    },
    webpack: webpackConfig,
    webpackServer: {
      noInfo: true
    }
  })
}
```

With a simple index.js for webpack

```js
require('es6-promise').polyfill();
const testsContext = require.context('../.tmp/test', true, /Test$/);
testsContext.keys().forEach(testsContext);
```

filipearena:
+1

americoneto1:
+1

dignifiedquire:
This is due to https://github.com/karma-runner/karma/pull/1791 @budde377 could you check this out please if you get a moment?

dignifiedquire:
Released `0.13.21` that reverts the problematic commit.

budde377:
Will do!

PR #1902, opened 2016-02-13
Feature fix no colors to run command

budde377:
I have not identified and fixed the problem with the 3rd-party reporters not working. The problem was that the reporters didn't update the `USE_COLORS` property, from its default value: `false`. The feature relies highly on this property to be updated and when it wasn't, the reporter would be interpreted as a no-color reporter.

In-order not to break backwards compatibility, a new property was introduced: `EXCLUSIVELY_USE_COLORS` with `undefined` as it's default value. By using this property, the no-colours feature (in the run command) will work for default reporters, i.e. dots and process, and reporters choosing to update the property. 

When fixing this, it became obvious why 3rd-party reporters didn't wan't to update this property: They didn't want to implement two reporters, because this would require for their users to add two reporters to their configuration.

For instance; for the mocha reporter, used in #1894, to implement this feature they would first implement two reporters, say `reporter:mocha` and `reporter:mocha_color`, which would yield a configuration as: 

    reporters: ['mocha', 'mocha_color'] 

where omitting one would disable reporting w/w.o colours, depending on the omitted reporter.

I fixed this in 3294a15 by attempting (silently) to load a "coloured" version of a given reporter. Thus for the mocha reporter to support this feature, all they need to do is implement and export two reporters: `reporter:mocha` and `reporter:mocha_color`. The user-configurations would remain unchanged: I.e.

    reporters: ['mocha'] 

Given a reporter with name `foo` the coloured version must be named `foo_color`, i.e. with postfix `_color`.


dignifiedquire:
This looks good but just to be sure can you add a e2e test using the mocha reporter that checks for the correct output please?

budde377:
@dignifiedquire Yes, I'll add it this weekend.

dignifiedquire:
@budde377 thanks. Do you want to join the karma maintainer team? Main rules are, keep doing great work like you did, and don't push/merge to `master` on your own.

budde377:
@dignifiedquire Yes, I would very much like that :)

dignifiedquire:
@budde377 done, welcome to the team

budde377:
@dignifiedquire Here you go!

dignifiedquire:
@budde377 this needs a rebase after I merged your other PR

budde377:
@dignifiedquire, Here you go ;)

dignifiedquire:
Thanks. Here goes nothing

dignifiedquire:
Damn. There is another conflict sorry

budde377:
No prop. This should do the trick.

PR #1904, opened 2016-02-13
chore(dependencies): update bluebird v.3

### LINKED ISSUE #1859: Upgrade to Bluebird@3

BridgeAR:
Fixes #1859 

I can't see that the cancellation is needed in this case as the promise that is cancelled is a runaway promise anyway.

### DIFF
--- a/lib/file-list.js
+++ b/lib/file-list.js
@@ -150,7 +150,7 @@ List.prototype._refresh = function () {
   var self = this
   var buckets = this.buckets
 
-  return Promise.all(this._patterns.map(function (patternObject) {
+  var promise = Promise.map(this._patterns, function (patternObject) {
     var pattern = patternObject.pattern
 
     if (helper.isUrlAbsolute(pattern)) {
@@ -167,7 +167,7 @@ List.prototype._refresh = function () {
       return
     }
 
-    return Promise.all(files.map(function (path) {
+    return Promise.map(files, function (path) {
       if (self._isExcluded(path)) {
         log.debug('Excluded file "%s"', path)
         return Promise.resolve()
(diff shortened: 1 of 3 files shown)


BridgeAR:
Ping @dignifiedquire

dignifiedquire:
This is not enough, there is more cancellation happening in that file that will fail with the Bluebird@3. See for some details where I have started work on this. Also it is important that the cancellation continues to work on a refresh while doing a refresh for performance reasons

BridgeAR:
There is no other cancellation happening in the whole karma library. There might be peer dependencies that use cancellation but not in here anymore.

There is a CancellationError thrown but this is just the error itself that is thrown without any cancellation happening there. So this will work just as before.

I saw your work and my first attempt to solve this looked almost the same. But this is not the solution to this, as the way how it worked earlier is the same as how I implemented it right now. Earlier, the promise was cancelled and the other promise was returned. But that promise ended in nirvana, as it is not truly responsible for the refresh. That is handled in the `.refresh` instead of the `._refresh` function and that returns the promise that the refresh is listening to.

So if I'm not really mistaken, the cancellation never had any effect and there's already a test in place that does a refresh while a refresh is ongoing and this would not pass otherwise.

dignifiedquire:
I see, but if do not cancel, won't this line https://github.com/BridgeAR/karma/blob/master/lib/file-list.js#L201 still be called? resulting in multiple emits even though the earlier one should have been canceled and not be emitted?

BridgeAR:
True point. Fixed and improved the test to catch that

dignifiedquire:
Thanks will review closely later today

dignifiedquire:
I think this doing now what we want. Thanks a lot for being patient with me :)

dignifiedquire:
Just one thing can you change the commit messages please, the first to sth `feat(file-list): Upgrade bluebird to v3` and the second to sth like `test(unit): ..`

BridgeAR:
@dignifiedquire of course :) Fixed

dignifiedquire:
Thanks :octocat:

PR #1914, opened 2016-02-18
fix(test): Throw the expected exception

davidparsson:
Before, the exception was thrown in MockSourceMapConsumer's
constructor and not in its originalPositionFor() function,
which however was the expected thing to happen.

While this is practice the same behavior, it was misleading when reading the test.


dignifiedquire:
Thanks, can you please change the commit message to `test(unit): ...` as `fix` is reserved for fixes that touch running code.

davidparsson:
@dignifiedquire, done! I actually tried to find out which one of the two I should use, but did not succeed.

Regarding the author feedback, the change I've made should likely have been made in fe6ed7e59ae7c8151cb5f1da53aa5f343f73df0b when `.content` was added, but was likely ignored since the test did not start to fail without it.

dignifiedquire:
Thanks a lot! :octocat: is happy

PR #1915, opened 2016-02-18
fix(reporter): Disable source maps for URLs without line number

### LINKED ISSUE #1274: Breaks source map if error message contain url without line number and column
We have two cases when it possible

- In latest version of [karma-jasmine](https://github.com/karma-runner/karma-jasmine) when test contain variable which does not defined we have error message:

```txt
ReferenceError: plus is not defined in http://localhost:9876/base/test.js?5e4e9f0005ec2172d5edc1234305ab74e3267861 (line 6)
```

- Firefox [sometimes](https://github.com/karma-runner/karma/pull/1098) throw error with formatted stack trace. (NOTE: Maybe it's first case)

- Rare case. User can define test cases or test suites which contain url, for example:

```js
it('should go to the http://lolcalhost/base/rest', function(){...})
```
 
In those three cases [reporter.js](https://github.com/karma-runner/karma/blob/master/lib/reporter.js#L49) breaks with error:
```
ERROR [karma]: [TypeError: Line must be greater than or equal to 1, got 0]
TypeError: Line must be greater than or equal to 1, got 0
```

davidparsson:
There is no way to resolve source maps in this case anyway. Fixes #1274.

### DIFF
--- a/lib/reporter.js
+++ b/lib/reporter.js
@@ -49,7 +49,7 @@ var createErrorFormatter = function (basePath, emitter, SourceMapConsumer) {
 
       var file = findFile(path)
 
-      if (file && file.sourceMap) {
+      if (file && file.sourceMap && line) {
         line = parseInt(line || '0', 10)
 
         column = parseInt(column, 10)
--- a/test/unit/reporter.spec.js
+++ b/test/unit/reporter.spec.js
@@ -75,12 +75,15 @@ describe('reporter', () => {
     })
 
     describe('source maps', () => {
+      var originalPositionForCallCount = 0
+
       class MockSourceMapConsumer {
         constructor (sourceMap) {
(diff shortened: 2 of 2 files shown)


dignifiedquire:
Thanks :octocat: waiting for travis to do it's thing

davidparsson:
Is that failure really related to my changes? I find that hard to believe.

dignifiedquire:
> Is that Travis failure really related to my changes? I find that hard to believe.

No, just Travis being slow :(

IgorAufricht:
Hi, thanks for the fix. When can we expect this to be released? It has been merged for 3 months, but there hasn't been any release of karma, so the fix is not available to use. Thanks!

carlos-:
we are also expecting this changes, when will be this released?

dignifiedquire:
@carlos- this is available in 1.0

PR #1931, opened 2016-02-25
feat: Fail on load-errors.

### LINKED ISSUE #855: Throw an error on not found browsers

budde377:
Stop karma if a launcher, reporter, plugin, or preprocessor fails to
load, after attempting to load each of them, and reporting errors for
each load error. 

~~Also someone should prob. tell @nikku that the karma-browserify 
plugin is going to stop working due the changed signature of 
`createPreprocessor`.~~

Closes #855

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -10,8 +10,12 @@ var retryDecorator = require('./launchers/retry').decoratorFactory
 var processDecorator = require('./launchers/process').decoratorFactory
 
 // TODO(vojta): remove once nobody uses it
-var baseBrowserDecoratorFactory = function (baseLauncherDecorator, captureTimeoutLauncherDecorator,
-  retryLauncherDecorator, processLauncherDecorator) {
+var baseBrowserDecoratorFactory = function (
+  baseLauncherDecorator,
+  captureTimeoutLauncherDecorator,
+  retryLauncherDecorator,
+  processLauncherDecorator
+) {
   return function (launcher) {
     baseLauncherDecorator(launcher)
     captureTimeoutLauncherDecorator(launcher)
@@ -20,7 +24,7 @@ var baseBrowserDecoratorFactory = function (baseLauncherDecorator, captureTimeou
   }
 }
 
-var Launcher = function (emitter, injector) {
+var Launcher = function (server, emitter, injector) {
   var browsers = []
   var lastStartTime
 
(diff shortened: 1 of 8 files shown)


budde377:
@srawlins, can you sign the CLA?

nikku:
@nikku does not like that change for the reason you mentioned (`createPreprocessor` syntax changed). Isn't there another way to implement that in a backward compatible manner?

budde377:
@nikku Sorry for the indirection. 
I might have an idea.

dignifiedquire:
@budde377 is this ready for review?

budde377:
@dignifiedquire, yes. I don't know what to do about the missing CLA from @srawlins.

lib/launcher.js
```diff
@@ -11,7 +11,7 @@ var processDecorator = require('./launchers/process').decoratorFactory
 
 // TODO(vojta): remove once nobody uses it
 var baseBrowserDecoratorFactory = function (baseLauncherDecorator, captureTimeoutLauncherDecorator,
-  retryLauncherDecorator, processLauncherDecorator) {
+                                            retryLauncherDecorator, processLauncherDecorator) {
```
dignifiedquire: Let's not do this please, instead use this indentation:

```js
var name = function (
  arg1,
  arg2,
  arg3
) {
  // body
}
```

test/e2e/load.feature
```diff
@@ -0,0 +1,106 @@
...
+      singleRun = false
+      """
+    When I start Karma
+    Then it fails with like:
+      """
+      Cannot load browser "NonExistingBrowser": it is not registered! Perhaps you are missing some plugin\?
+      """
+    And it fails with like:
+      """
+Found 1 load error
```
dignifiedquire: Indentation?

lib/launcher.js
```diff
@@ -20,7 +20,7 @@ var baseBrowserDecoratorFactory = function (baseLauncherDecorator, captureTimeou
   }
 }
 
-var Launcher = function (emitter, injector) {
+var Launcher = function (server, emitter, injector) {
```
dignifiedquire: Are we sure this is not going to cause any issues in plugins?

budde377: If we don't want to change the signature, I can just pull out the server from the injector directly, as I did in `lib/preprocessor.js`.

dignifiedquire: Hmm not sure, what do you think?

dignifiedquire:
Closes #855

dignifiedquire:
Thanks a lot @budde377 and @srawlins for making this happen!

budde377:
@dignifiedquire I just want to notice, that I didn't do the whole extract server from injector thing, so we might see some launchers breaking if they rely on the signature of the launcer.

dignifiedquire:
@budde377 yeah I realised that, but I wanted to get it in, will fix reporters when they tell us, this is not really documented stable api anyway

PR #1934, opened 2016-02-26
fix(runner): Merge config.client.args with client.args provided by run

### LINKED ISSUE #1746: `karma run` overwrites client.args state
### Summary

`karma run` calls this line [lib/middleware/runner.js#L50](https://github.com/karma-runner/karma/blob/v0.13.15/lib/middleware/runner.js#L50) which overwrites the state of `config.client.args` with an empty array. This means that plugins that rely on `config.client.args` will run when calling `karma start`, but fail after calling `karma run` once (even when they are subsequently autorun from the `karma start` process).

### Details

I'm using [karma-cljs-test](https://github.com/honzabrecka/karma-cljs-test) to run my ClojureScript tests. It needs a main function to be provided in `config.client.args`. I was able to run my tests successfully through the auto runner in `karma start` but when I called `karma run`, my tests would fail to run as `config.client.args` had been overwritten with an empty array. Because `config` is a mutable object, once I had called `karma run` once, my tests would fail when being rerun by the `karma start` autorunner.

Commenting out 

```js
config.client.args = data.args
```

let my tests pass, but this is clearly not a viable solution. One option would be to only set `config.client.args` if `data.args` is not an empty array. 

This also raises the wider question of why a `karma run` is setting global mutable state. That behaviour was surprising to me, and I don't think it is intentional.

This behaviour seems to have been introduced in https://github.com/karma-runner/karma/commit/ca4c4d88b9a4a1992f7975aa32b37a008394847b, and relates to https://github.com/karma-runner/karma/commit/27b8d678cdc43f823d2531d25fc548d542dc9078, https://github.com/karma-runner/karma/issues/283, https://github.com/karma-runner/grunt-karma/issues/79, https://github.com/karma-runner/grunt-karma/issues/79, https://github.com/karma-runner/grunt-karma/commit/b658ec9701c7a35673a3011b703e3888364a3733 and https://github.com/karma-runner/grunt-karma/commit/168f7b7a6af02e78e44b87b27553cdb65785221c.

danielcompton:
Before this change, calling `karma run` would overwrite any value in config.client.args with the value provided in the `karma run` request, even if that value was an empty array. This commit does a _.merge to merge the two values together.

Fixes #1746

### DIFF
--- a/test/unit/middleware/runner.spec.js
+++ b/test/unit/middleware/runner.spec.js
@@ -5,6 +5,7 @@ import {Promise} from 'bluebird'
 import Browser from '../../../lib/browser'
 import BrowserCollection from '../../../lib/browser_collection'
 import MultReporter from '../../../lib/reporters/multi'
+var _ = require('lodash')
 var createRunnerMiddleware = require('../../../lib/middleware/runner').create
 var HttpResponseMock = mocks.http.ServerResponse
 var HttpRequestMock = mocks.http.ServerRequest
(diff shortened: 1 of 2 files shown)


test/unit/middleware/runner.spec.js
```diff
@@ -166,6 +166,28 @@ describe('middleware.runner', () => {
     request.emit('end')
   })
 
+  it('calling run with no client args should not overwrite existing client.args', (done) => {
```
dignifiedquire: Could you add a test for the actual merge with different values please?

danielcompton:
I've added tests for the following cases:

* existing config client.args, empty array run client.args
* existing config client.args, nothing passed for run client.args
* existing config client.args, value passed for run client.args
* no config client.args, value passed for run client.args
* empty map client.args, value passed for run client.args

The only wrinkle here, and I'm not sure what can be done about it, is if the user passes `[]` as client.args, and a map as run client.args. `_.merge` doesn't convert the `[]` into a map, and discards the run args.

budde377:
@danielcompton, How about checking the type of the client-args and take appropriate measures? 
E.g. using the helpers in https://github.com/karma-runner/karma/blob/master/lib/helper.js#L20 .

danielcompton:
What is this the expected behaviour for merging an array and a map, and for merging a map and an array?

budde377:
God question, maybe just let the run args override the existing?

dignifiedquire:
Yes I think in the case that the the types are not mergeable, i.e. array < object or object < array, we should override and print a warning.

danielcompton:
This is ready for review now.

dignifiedquire:
Thanks :octocat:

danielcompton:
Thank you!

Issue #1876, opened 2016-02-06
window.__karma__.files should escape file names


Joris-van-der-Wel:
One of the files being served in my project has a single quote in its name. Karma is unable to run any tests because of this.

This is the culprit: https://github.com/karma-runner/karma/blob/19dd8249c6ce692004f4d52f8c0b2bb863e1f3fb/lib/middleware/karma.js#L148

Why not use `JSON.stringify()`?

Moumi:
It seems that this issue can be solved for single quotes, but for double quotes... Well, you should never create double quotes in a filename. So, I worked on a fix and test for this and will provide a PR this evening for it.

Joris-van-der-Wel:
Would it not be easier to use: 
```javascript
return util.format("  %s: '%s'", JSON.stringify(filePath), file.sha)
```

This way you can not make any mistake in the escaping.

```javascript
console.log(JSON.stringify('foo\n"bar"')) // "foo\n\"bar\""
```

Moumi:
There are some cases that when a character is escaped the loading of the file would go wrong. The most safe way would be to find the cases that are 'common' and escape only those characters.

Joris-van-der-Wel:
How so?

All that we are doing is passing a JavaScript variable from one context (node) to another (the browser), `JSON.stringify` should take care of that. If something starts failing with that, you probably discovered a different issue.

alancutter:
I'm still hitting this problem in the latest release: https://github.com/karma-runner/karma/releases/tag/v0.13.22
The merged change doesn't seem to be included for some reason.

PR #1938, opened 2016-02-28
fix(karma): Fix issue regarding escape quotes

Moumi:
Change a line to escape single quote characters in filenames.

Closes *[#1876](https://github.com/karma-runner/karma/issues/1876)*


dignifiedquire:
Thanks, can you squash the two commits into one please?

Moumi:
@dignifiedquire I have squashed them together. Never done it before, so I hope it all went well. :wink:

dignifiedquire:
Thanks :octocat: looks good.

alancutter:
Also this should probably use JSON.stringify() instead of hand rolled replace() calls. Files with newlines in them will also cause problems.

Issue #1719, opened 2015-11-22
Allow specifying the javascript version in the debug html window


simonzack:
I would like to test javascript 1.8 as this is what I would like my target to run on (greasemonkey on firefox), but currently `SCRIPT_TYPE` is hard-coded

            return util.format(SCRIPT_TAG, SCRIPT_TYPE[fileExt] || 'text/javascript', filePath)

without `application/javascript;version=1.8` firefox runs ES5 only, so can there be some option to put this in the config?

Moumi:
@simonzack This is possible to do. However, in case the user uses multiple browsers, this might go wrong. If, in any case Chrome is used to run besides Firefox 4 (which supports 1.8 now), the tag will make the tests all fail. Even more, Karma will throw an **Error**. 

I was able to make the configuration for you and was wanting to add a PR for it, but I stumbled upon the error part of Karma when trying to add the tag when running Chrome. Maybe @dignifiedquire has an idea how to make this fail-safe?

dignifiedquire:
Not sure (I never had to use this in a real world example before)
But we just merged custom context and debug files in #1825  so maybe that can help you to specify exactly what you want.

Moumi:
I have seen that, but it is not exactly what I'm looking for. The case would be that Firefox 4 *supports* 1.8, but Chrome (even latest version) does *not*. This makes it that in the *karma.js*-file, when the Firefox 4 browser is run, it should add the tag of ";version=1.8", otherwise not. I'd think that there should be some check which browser is being run at that moment and perhaps the version of it. 

This might be a solution to it, but takes the effort to update this *whitelist* of browsers to be updated everytime a new update comes out...

Moumi:
@dignifiedquire: Do you know a possible way to capture the current browser being run? That would solve the issue mainly.

dignifiedquire:
@Moumi capture it where?

Moumi:
@dignifiedquire: At https://github.com/karma-runner/karma/blob/master/lib/middleware/karma.js just before the return statement:
*return util.format(SCRIPT_TAG, SCRIPT_TYPE[fileExt] || 'text/javascript', filePath)*

Moumi:
@dignifiedquire ping

dignifiedquire:
I don't think that will easily work as the file is simply served to every browser the same atm. You could check the express docs how to do browser detection in a single request

Moumi:
[...]

PR #1939, opened 2016-02-28
feat(config): Add configuration for javascript version. 

Moumi:
Add configuration for a javascript version tag. This tag is only applied to Firefox (now).

Closes *[#1719](https://github.com/karma-runner/karma/issues/1719)*


lib/middleware/karma.js
```diff
@@ -59,9 +60,21 @@ var getXUACompatibleUrl = function (url) {
   return value
 }
 
+function applyJavascriptVersionTag (req) {
```
dignifiedquire: Better name would be `isFirefox` or `hasJsVersionTags`

test/e2e/tag.feature
```diff
@@ -0,0 +1,40 @@
...
+      """
+    When I start Karma
+    Then it passes with:
+      """
+      ..
+      PhantomJS
+      """
+
+  Scenario: Execute a test in Firefox with version, with JavaScript tag
+    Given a configuration with:
```
dignifiedquire: This test will not actually check that the version tag is available, you should add another file that actually checks the html for the version tag

Moumi: How would I be able to do this. I am not familiar with e2e tests.

dignifiedquire: You can create a new file similar to what is in `basic/test.js`, e.g. `basic/version.js` and include it in the files here. That means that file will be run in the browser. Now in that file you can put some js in that reads the current html and checks for the version tag to be present, and fails if it's not present.

Moumi:
This PR is getting a bit messy. Closing it to create a new one.

Issue #1936, opened 2016-02-27
Karma reconnect screen display wrong time unit


goldensunliu:
### Expected behavior
should use Seconds instead of ms.
### Actual behavior
![image](https://cloud.githubusercontent.com/assets/1348993/13373866/f722dc86-dd40-11e5-98df-bddb527faa67.png)

### Steps to reproduce the behaviour

1. run karma
2. launch a browser and kill karma
3. look at the reconnect copy in English

Moumi:
I have seen this issue myself and have created a fix. Anytime this evening I will create a PR for it.

goldensunliu:
@Moumi thanks! Community power!

dignifiedquire:
Thanks for the report, @Moumi PR already got merged

PR #1940, opened 2016-02-28
fix(updater): Fix time unit on screen display from 'ms' to 'seconds'.

Moumi:
Change 'ms' to 'seconds' in **updater.js**.

Closes *[#1936](https://github.com/karma-runner/karma/issues/1936)*


dignifiedquire:
Thanks :octocat:

PR #1974, opened 2016-03-07
fix(files): Ignore included:false pattern

### LINKED ISSUE #1530: Karma 0.13 ignores included:false pattern

budde377:
Do not include a file, when `include=false`.
Give `include=false` precedence over `include=true`.

Closes #1530

### DIFF
--- a/test/e2e/files.feature
+++ b/test/e2e/files.feature
@@ -1,0 +1,153 @@
+Feature: Including files
+  In order to use Karma
+  As a person who wants to write great tests
+  I want to be able to include and exclude files
+
+  Scenario: Execute a test excluding a file
+    Given a configuration with:
+      """
+      files = [
+        {pattern: 'files/log_foo.js', included: false},
+        'files/*.js'
+       ];
+      browsers = ['PhantomJS'];
+      plugins = [
+        'karma-jasmine',
+        'karma-phantomjs-launcher'
+      ];
+      """
+    When I start Karma
+    Then it passes with:
+      """
+      .
+      PhantomJS
+      """
+
+  Scenario: Execute a test excluding an explicitly included file
+    Given a configuration with:
+      """
+      files = [
+        {pattern: 'files/log_foo.js', included: true},
+        {pattern: 'files/log_foo.js', included: false},
+        'files/*.js'
+       ];
+      browsers = ['PhantomJS'];
+      plugins = [
+        'karma-jasmine',
+        'karma-phantomjs-launcher'
+      ];
+      """
+    When I start Karma
+    Then it passes with:
+      """
+      .
+      PhantomJS
+      """
+
+  Scenario: Execute a test excluding an explicitly included file in another order
+    Given a configuration with:
+      """
+      files = [
+        'files/*.js',
+        {pattern: 'files/log_foo.js', included: true},
+        {pattern: 'files/log_foo.js', included: false}
+       ];
+      browsers = ['PhantomJS'];
+      plugins = [
+        'karma-jasmine',
+        'karma-phantomjs-launcher'
+      ];
+      """
+    When I start Karma
+    Then it passes with:
+      """
+      .
+      PhantomJS
+      """
+
+  Scenario: Execute a test excluding an file included with brackets patterns
+    Given a configuration with:
+      """
+      files = [
+        'files/test.js',
+        {pattern: 'files/log_foo.js', included: false},
+        {pattern: 'files/{log,bug}_foo.js', included: true}
+       ];
+      browsers = ['PhantomJS'];
+      plugins = [
+        'karma-jasmine',
+        'karma-phantomjs-launcher'
+      ];
+      """
+    When I start Karma
+    Then it passes with:
+      """
+      .
+      PhantomJS
+      """
+
+  Scenario: Execute a test excluding an file included with wildcard
+    Given a configuration with:
+      """
+      files = [
+        'files/test.js',
+        {pattern: 'files/+(log|bug)_foo.js', included: false},
+        {pattern: 'files/*.js', included: true}
+       ];
+      browsers = ['PhantomJS'];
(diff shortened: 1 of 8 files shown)


test/e2e/files.feature
```diff
@@ -0,0 +1,45 @@
...
+      PhantomJS
+      """
+
+  Scenario: Execute a test excluding an explicitly included file
+    Given a configuration with:
+      """
+      files = [
+        {pattern: 'files/log_foo.js', included: false},
+        {pattern: 'files/log_foo.js', included: true},
+        'files/*.js'
```
dignifiedquire: It would be good to have one test case where `files/*.js` is the first entry.

budde377:
@dignifiedquire There's an [integration test](https://github.com/karma-runner/integration-tests/blob/master/coverage-mocha-requirejs/karma.conf.js#L7) that fails, because it assumes some other semantic than changed precedence. Should the test be changed or should the semantics?

test/e2e/files.feature
budde377: :+1:

dignifiedquire:
@budde377 right, so we have to think about this carefully, because in the 0.12 and before the semantics where that the most specific pattern would match, rather than `false > true`. So ideally I thought we could revert to that, what do you think?

budde377:
@dignifiedquire. Lets do this the cool way! Inspired by the CSS specificity calculations, we should define a specificity-measure. I propose the following:

Minimatch supports:

 * Globstar
 * Wildcard
 * Extended glob matching
 * Brace expansion

Given a path, *p*, we can define a tuple *s(p) = (n<sub>g</sub>, n<sub>w</sub>, n<sub>e</sub>, n<sub>b</sub>)* where *n<sub>g</sub>*, *n<sub>w</sub>*, *n<sub>e</sub>*, and *n<sub>b</sub>* are the number of globstars, wildcards, extended glob match patterns, and brace expansion patterns, respectively. 

With any two paths, *p* and *q*, where *s(p)=(p<sub>1</sub>, p<sub>2</sub>, p<sub>3</sub>, p<sub>4</sub>)* and *s(q)=(q<sub>1</sub>, q<sub>2</sub>, q<sub>3</sub>, q<sub>4</sub>)* we can define a order of the corresponding tuples 

 *p<sub>1</sub> < p<sub>2</sub> ⇔ ∃ m > 0: ∀ i > m: p<sub>i</sub> = q<sub>i</sub> ∧ p<sub>m</sub> < q<sub>m</sub>* 

i.e. an lexicographical order.

Given two paths with different tuples, we can define the one with the smallest tuple (wrt. the order above) as the most specific. If the tuples are equal we can choose either one.

budde377:
Or we could just let the number of files generated from a given pattern decide how specific it is. The more files, the less specific.

dignifiedquire:
@budde377 :100: for the cool order you suggested above. We just need to make sure to document and test this well, so people understand what's happening :)

budde377:
@dignifiedquire Yes. I'll start on it this afternoon.

budde377:
After some failed attempts, I'm pretty sure that I have a solution. I'll have it implemented and tested primo next week.

dignifiedquire:
Thanks @budde377 for the great work, looking forward to see it :)

budde377:
@dignifiedquire I've now implemented an order. Please let me know if it needs more documentation etc.

dignifiedquire:
Nice work, will review more closely tomorrow

test/unit/helper.spec.js
```diff
@@ -245,4 +245,66 @@ describe('helper', () => {
...
+    it('should calculate right weight of pattern with glob star', () => {
+      helper.mmPatternWeight('**/a').should.be.deep.equal([1, 1, 0, 0, 0, 0])
+    })
+    it('should calculate right weight of pattern with misused glob star', () => {
+      helper.mmPatternWeight('***/a').should.be.deep.equal([1, 0, 3, 0, 0, 0])
+    })
+    it('should calculate right weight of pattern with more magic', () => {
+      helper.mmPatternWeight('{0..9}/?(a|b)c/[abc]/**/*.jsx?').should.be.deep.equal([10, 1, 1, 1, 1, 1])
+    })
+    it('should calculate right weight of pattern as worst glob set', () => {
```
dignifiedquire: > worst glob set

:+1: :laughing:

dignifiedquire:
Thanks @budde377 looking good, going to merge this and hope for the best :)

PR #1984, opened 2016-03-11
Transition from contextWindow to postMessage

twolfson:
I am currently attempting to get Karma running with Electron. There are a lot of benefits to this:

- Can hide Electron windows like PhantomJS
- Can load modules easily via `require`
- Environment is identical to production if we are building an Electron app (which I am)
- Built-in screenshot support for visual diffs

Unfortunately, the current Karma implementation is incompatible with Electron.

If we use an iframe, we lose the Node.js integration and must polyfill it via `window.parent`. This works great for top level modules but submodules have their global context executed in the `window.parent` which makes using `mocha` impossible =/

> For reference, here was our polyfill to date: https://github.com/twolfson/karma-electron/blob/3.0.1/lib/node-integration-iframe.mustache.js

Other alternatives from an iframe for Electron are `window.open`, `<webview>`, and launching multiple `BrowserWindow` instances. Unfortunately, these options are executed in separate processes so we cannot use `window.opener` as per usual.

- `window.open` defines a proxy where we can run `eval` and `postMessage`
    - Documentation: https://github.com/atom/electron/blob/v0.36.10/docs/api/window-open.md
- `<webview>` defines `executeJavaScript` and `send`
    - Both of these methods are similar to `window.open` except they can receive responses whereas neither `eval` nor `postMessage` can directly receive responses (although we can send a corresponding `postMessage` reply or build a bridge library to match requests/responses)
    - Documentation: https://github.com/atom/electron/blob/v0.36.10/docs/api/web-view-tag.md
- Launching new `BrowserWindow` via Electron seems too custom to Electron to be reusable elsewhere
    - I should note that we would have `sendSync` in the browser which allows for synchronous function calls and responses but this only works when sending from `renderer` to `main` processes

I decided to use `window.open` since it doesn't require any duplicate logic for other browsers and is easily polyfilled for older browsers to use `contextWindow`.

In this PR, I have a proof of concept working with the `karma-electron` test suite but the PR is far from done.

Matching test suite: https://github.com/twolfson/karma-electron/tree/ba7d54bf4d941200c9c15d94ebf0cbfd7546d5c7

What are your thoughts on this PR?


twolfson:
I signed it!

static/context.html
```diff
@@ -16,13 +16,96 @@
...
        an epic manner. -->
   <script type="text/javascript">
-    // sets window.__karma__ and overrides console and error handling
-    // Use window.opener if this was opened by someone else - in a new window
-    if (window.opener) {
-      window.opener.karma.setupContext(window);
-    } else {
-      window.parent.karma.setupContext(window);
-    }
+    (function () {
```
dignifiedquire: All this code should still be in a separate file rather than inlined here.

dignifiedquire:
Hey, 

first thanks for all this work! In general I like the idea of using `window.open`, but we need to be a 100% certain we can polyfill and still support all browsers back to IE7 as we do currently. I can't ship a version that drops that compatibility as too many people are relying on this.

static/context.html
twolfson: Yea, definitely. This was mostly to get the proof of concept up and out the door =)

twolfson:
The polyfill would be less of an actual polyfill and more of a fallback implementation for Electron:

```js
// Define a remote call method for Karma
var parentWindow = window.opener || window.parent;
var callParentKarma = function (method, args) {
  parentWindow.karma[method].apply(parentWindow.karma, args);
};

// If we don't have access to the window, then use `postMessage`
// DEV: In Electron, we don't have access to the parent window due to it being in a separate process
// DEV: We avoid using this in Internet Explorer as they only support strings
//   http://caniuse.com/#search=postmessage
if (!parentWindow.window) {
  callParentKarma = function (method, args) {
    parentWindow.postMessage({method: method, arguments: args}, window.location.origin);
  };
}
```

I think the next iteration of this PR won't be a fully polished product but another proof of concept demonstrating the concept works in other browsers

twolfson:
Alright, I was able to make a clean transition for everything (or at least I think it's clean). There are still some rough edges in this PR (e.g. lingering TODOs, using PhantomJS over Chrome/FF so I can dev faster) but the majority of it is here.

I know that a large PR can be frustrating as a reviewer. As a result, I had some interim branches for you to glance over:

https://github.com/twolfson/karma/compare/87c56e3...twolfson:dev/postmessage.work

Also, here was my initial game plan although I derailed from the path at some point:

https://gist.github.com/twolfson/f7d257966d7cdafdf30c

twolfson:
@dignifiedquire Would it be possible to get a review on this PR? I'm more than happy to break it down into smaller pieces if that helps

dignifiedquire:
@twolfson yes I will try to get to it tomorrow

.travis.yml
```diff
@@ -1,9 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-  - "0.10"
```
dignifiedquire: We need to pass our tests on all these platforms, please leave them in here

client/karma.js
```diff
@@ -175,21 +147,8 @@ var Karma = function (socket, iframe, opener, navigator, location) {
...
-    if (!hasError) {
-      this.start(this.config)
-    }
-
-    // remove reference to child iframe
-    this.start = UNIMPLEMENTED_START
-  }
-
+  // TODO: Does anyone use `this.store`???
+  //   https://gitter.im/karma-runner/karma?at=56e48e07618c335373eb497f
```
dignifiedquire: I'm not sure :/

static/debug.html
```diff
@@ -18,6 +18,8 @@
    into it. If it is before body, then it fails to find the body and crashes and burns in an epic
    manner. -->
   <script type="text/javascript">
+    // TODO: Determine if we need to update this or not...
```
dignifiedquire: Did you try debugging in electron?

test.sh
```diff
@@ -0,0 +1 @@
+npm run build && grunt test:client
```
dignifiedquire: Let's remove this from here please

test/client/karma.conf.js
```diff
@@ -68,7 +68,7 @@ if (process.env.TRAVIS) {
     browsers = Object.keys(launchers)
   }
 } else {
-  browsers.push('Chrome')
+  browsers.push('PhantomJS')
```
dignifiedquire: Let's change this so it always runs Chrome and PhantomJS

test/e2e/basic.feature
```diff
@@ -19,37 +19,3 @@ Feature: Basic Testrunner
       ..
       PhantomJS
       """
-  @not-jenkins
```
dignifiedquire: Why did you remove this?

dignifiedquire:
Alright this is starting to take nice form, I left some smaller comments inline. 
The main thing I'm pondering right now is if I want to get this in before or after shipping 1.0 (which I wanted to do last week already *cough*).
I'm going to test this on some of my projects tomorrow to see if there are any compat issues that I can see. 

Could you also do a rebase please of latest master so it merges cleanly?

twolfson:
Hooray, thanks for the review =) Most of the comments were on files/edits that were done for local development (i.e. make everything headless/simple to make iterations as fast as possible).

They were left in due to this PR still being up in the air. Now that it sounds like this will eventually be landed, then I agree that it's best to take them out and add finishing polish.

With respect to before/after 1.0, it's prob best to hold this off until after 1.0. I suggest shipping this feature as standalone as possible since it could break builds (unlikely but plausible) and we want to be easily revertable.

With respect to the requested changes in this PR, I will take care of them by the end of the weekend.

test/e2e/basic.feature
twolfson: It was causing windows to pop-up during local development. I will restore this.

test/client/karma.conf.js
twolfson: Yep, this change was for local development. I will restore it

static/debug.html
twolfson: Nope, but I will try and verify `debug.html` works as expected

test.sh
twolfson: Another artifact from local development. I will remove it

.travis.yml
twolfson: These were removed for local development. I will restore them

.travis.yml
```diff
@@ -37,3 +34,10 @@ before_script:
 
 script:
   - npm run travis
+
+notifications:
+  email:
+    recipients:
+      - todd@twolfson.com
+    on_success: change
+    on_failure: change
```
twolfson: Note to self: Remove these edits as well

client/karma.js
twolfson: I'm pretty sure it's no longer used. It was introduced in this commit which is 4 years old and apparently when the Jasmine adapter was still in this repo o_o

https://github.com/karma-runner/karma/commit/eedcbbd9cdc5d13b09bcef7dd12b1462a2a7a155

I can't find any references to `store` in `karma-jasmine` or any other spot checked repos so I think we are good to remove it:

https://github.com/karma-runner/karma-jasmine/search?utf8=%E2%9C%93&q=store

twolfson: To keep things simple, I think I will submit another PR to remove it exclusively

dignifiedquire: Sounds good :)

test/client/karma.conf.js
```diff
@@ -91,6 +91,11 @@ module.exports = function (config) {
       'test/client/*.js': ['browserify']
     },
 
+    // TODO: Remove useIframe override and make sure it's replaced with tests elsewhere
+    client: {
+      useIframe: false
+    },
```
twolfson: Note to self, remove this edit as well

test/client/karma.conf.js
```diff
@@ -147,6 +152,7 @@ module.exports = function (config) {
       'karma-mocha',
       'karma-chrome-launcher',
       'karma-firefox-launcher',
+      'karma-phantomjs-launcher',
```
twolfson: And remove this one...

twolfson:
Not sure if you caught this but what were your thoughts on the folder structure (i.e. relocating some `client/` files to a `common/` and creating a `context/` folder)?

I know that some developers like to keep the top level folder as clean as possible so an alternative might be a `browser/` folder which contains each of those folder or keep all the files in a `browser/` and rename `client/karma.js` to `browser/parent-karma.js`

twolfson:
I'm realizing that when we introduce `customContextFile` and `customDebugFile` that we will have a major breaking change when this PR's new context/debug file formats land. I'm going to break those off into another PR so we can sneak it in before 1.0.

twolfson:
:scope creep: =/

context/main.js
```diff
@@ -0,0 +1,24 @@
+// TODO: Be sure that we lint this...
```
twolfson: Another note to self: Be sure to take care of these TODOs...

twolfson:
The `customContextFile`/`customDebugFile` future proofing has been relocated to https://github.com/karma-runner/karma/pull/2019

twolfson:
I left this TODO in the tests but it prob belongs here. We are losing configs being automatically updating in `context.html` via `execute`. Previously, they were updated as we went since `client` would receive them from config and that object was shared to `context.html`.

Now, we are loading config via HTML and on load so it won't update seamlessly if someone has set `clearContext: false`. However, I'm not sure if that will be a problem or not =/

twolfson:
Alright, all of the comments have been taken care of and this PR is now on top of #2019 and #2020. Here's a screenshot of `debug.html` working in Electron:

selection_221

dignifiedquire:
Nice! I will be travelling the next week, so I might be slow to respond, but going to test this out and see how it works :)

twolfson:
This PR is now on top of the latest changes in #2019 and #2020

twolfson:
Bump on this PR and its related PRs =/

twolfson:
This PR should be good to review now as well :+1:

twolfson:
Bump on this issue again =/

twolfson:
Bump on this issue again =/

dignifiedquire:
@twolfson I finally got around to releasing 1.0. Sorry for the massive delay on this but I had a lot to do on my day job. Reviewing this now, in the meantime could you please rebase this onto the latest master?

twolfson:
I have rebased the commit to the latest master

dignifiedquire:
Thanks a lot @twolfson, going to merge this and put it into the new release

dignifiedquire:
@twolfson released as 1.1.0

twolfson:
woot, awesome. Thanks =D

PR #1989, opened 2016-03-12
Added commit message validation to Travis CI

twolfson:
When working on a repo, I like to commit frequently in a working branch. The `ghooks` were getting on my nerves so I removed them.

However, I have realized that I was able to open #1984 with a bad commit message format as a result. To remedy that, I'm adding the commit message validation to Travis CI.

In this PR:

- Added `validate-commit-msg` to `npm run travis` to enable non-local message validation

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,9 @@ env:
   - SAUCE_USERNAME: karmarunnerbot
   - secure: "bRVY+hYZwMf1SqVnMyZRJTLD0gN1hLx9/MwO8MM/qBiu3YNjXy49XElfMdzMKN6cZeKTmhcnjmZonbJuI1PQ2t+utGkyjnlVLJ/OlWptreKLzIlcbt4hrdPoTcjmUTwDWq9Ex9cVoYX8AzCasETttpczp3P+s3+vmOUj8z25JyU="
   - CXX=g++-4.8
+  include:
+  - node_js: "5"
+    env: VALIDATE_COMMIT_MSG=TRUE
 
 # Make sure we have a new npm on Node 0.10. npm >= 2 should work so by not updating them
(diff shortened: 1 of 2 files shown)


twolfson:
Ah, nvm. It looks like `validate-commit-msg` is only built to handle loading from files (e.g. `.git/COMMIT_MSG`). This won't work well with Travis CI without a long workaround =(

dignifiedquire:
Thanks, the idea is good though. Ideally we would have another check different from travis showing up similar to how we have the CLA check showing up which validates the commit message. But I haven't figured out how to do that yet.

twolfson:
Ah, that's easily done via Travis CI's build matrix:

https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix

I have a similar setup on some of my projects:

https://github.com/twolfson/gmsmith/blob/1.1.0/.travis.yml#L1-L19

https://travis-ci.org/twolfson/gmsmith/builds/114967673

twolfson:
I will take a similar attempt at a setup. Although, I think we will want a `bash` script to handle generating a file for the commit message (we could add a PR to support `-` for `stdin` on `verify-commit-msg` but this is simpler)

twolfson:
Alright, this PR is now good to go. I went with the `matrix` example as mentioned earlier. It's a bit unfortunate that we still go through installing Firefox for our one-off environment but such is life =/

Here's an expected failing build (this revision wasn't using `--no-merges` so it validated against a merge commit's message):

https://travis-ci.org/karma-runner/karma/builds/115744881

Here's a passing build:

https://travis-ci.org/karma-runner/karma/builds/115746435

dignifiedquire:
This is great thanks @twolfson, just going to let travis run its course before merging :)

dignifiedquire:
Thanks again for this!

PR #2001, opened 2016-03-18
fix(helper): Patched replaceWinPath from choking on `null` values

### LINKED ISSUE #2000: No `customContextFile` causes Windows to fail
I'm developing `karma-electron` with a fork off of `master` which uses #1984. I noticed that AppVeyor is failing while Travis CI is passing and it didn't do this on `0.13.22`.

**Failure on `master` fork:**

https://ci.appveyor.com/project/twolfson/karma-electron-launcher/build/job/a2glw31p686akewh

**Success on `0.13.22`:**

https://ci.appveyor.com/project/twolfson/karma-electron-launcher/build/80/job/ugvxnlxy518qy30n

After a bit of debugging, I realized this was due to us not doing a `null` check in `helper.normalizeWinPath`:

https://github.com/karma-runner/karma/blob/v0.13.22/lib/helper.js#L71-L75

https://github.com/karma-runner/karma/blob/b2cb31a204a295770f845384ea12fbbe0ad8e0d6/lib/config.js#L117-L118

### Expected behavior
- Ran `karma` and received following error:

```
C:\projects\karma-electron-launcher\node_modules\karma\lib\helper.js:72
  return exports.isDefined(path) ? path.replace(/\\/g, '/') : path
                                        ^
TypeError: Cannot call method 'replace' of null
    at Object.replaceWinPath [as normalizeWinPath] (C:\projects\karma-electron-launcher\node_modules\karma\lib\helper.js:72:41)
    at normalizeConfig (C:\projects\karma-electron-launcher\node_modules\karma\lib\config.js:119:37)
    at Object.parseConfig (C:\projects\karma-electron-launcher\node_modules\karma\lib\config.js:335:10)
    at new Server (C:\projects\karma-electron-launcher\node_modules\karma\lib\server.js:56:20)
    at Object.exports.run (C:\projects\karma-electron-launcher\node_modules\karma\lib\cli.js:262:7)
    at Object.<anonymous> (C:\projects\karma-electron-launcher\node_modules\karma\bin\karma:3:23)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:935:3
```

### Actual behavior
- Expected `karma` to pass without errors

### Enviroment Details
- Karma version (output of `karma --version`): Derivative off of `master` (post-`0.13.22`)

Here's what I'm doing but there's prob a better way to handle this (e.g. IE VMs):

1. Fork `karma-electron`
2. Enable AppVeyor for fork
3. Run a build and see it fail

### Recommendations
To prevent things like this landing on `master`, I suggest we set up CI with AppVeyor in addition to the current Travis CI. This can be simple like only running `grunt test` but it should catch issues like this one.

twolfson:
As reported in #2000, the current `master` is failing on Windows due to `customContextFile` being `null`. In this PR:

- Updated `replaceWinPath` to check for value being a string rather than not being undefined
- Fixes #2000

Here's AppVeyor failing on the current `master` and passing with our patched branch:

https://ci.appveyor.com/project/twolfson/karma-electron-launcher/build/94/job/dpsv7d5nv0sxcj6a

https://ci.appveyor.com/project/twolfson/karma-electron-launcher/build/95/job/vjt895wok3qwgvo3

### DIFF
--- a/lib/helper.js
+++ b/lib/helper.js
@@ -131,7 +131,7 @@ exports.formatTimeInterval = function (time) {
 }
 
 var replaceWinPath = function (path) {
-  return exports.isDefined(path) ? path.replace(/\\/g, '/') : path
+  return _.isString(path) ? path.replace(/\\/g, '/') : path
 }
 
 exports.normalizeWinPath = process.platform === 'win32' ? replaceWinPath : _.identity


dignifiedquire:
Thanks. To avoid these issues in the fututre do you think you vould help me creating an appveyorconfig for this repo?  It looks like you have some exp with that.

twolfson:
Sure thing, it's been a while since I actually wrote one. My `appveyor.yml` is mostly copy/paste from "that one time I set one up" but it's better than nothing.

A few nuances to be aware of:

- Builds run in series, not parallel (e.g. `node@0.10`, then `node@0.12`)
- Build notifications can be set on a per-user basis (i.e. via the site settings, not in `appveyor.yml`)
- Good tool for setting environment variables properly in Windows and Linux
    - https://www.npmjs.com/package/cross-env
- Node.js will autocorrect Linux file paths to their Windows counterparts in some scenarios (e.g. CLI, `fs`)
    - This is such a life saver <3
    - Reference code
        - https://github.com/nodejs/node/blob/v5.9.0/lib/fs.js#L230-L262
        - https://github.com/nodejs/node/blob/v5.9.0/lib/path.js#L685-L694
        - https://github.com/nodejs/node/blob/v5.9.0/lib/path.js#L146-L290
        - https://github.com/nodejs/node/blob/v5.9.0/lib/path.js#L11-L62

twolfson:
Good news, first run: Mocha tests are passing =) but getting an error with `grunt test:client`

Surprisingly AppVeyor provides Chrome, Firefox, and Internet Explorer:

https://www.appveyor.com/docs/installed-software#selenium

twolfson:
Going to be locally debugging on a VM from https://dev.windows.com/en-us/microsoft-edge/tools/vms/linux/ to move faster

twolfson:
Got local version more/less running. `grunt test:e2e` is stubborn so I'm prob going to skip it. Common errors:

- ` { [Error: Command failed: C:\Windows\system32\cmd.exe /s /c "E:\karma\bin\karma run E:\karma\tmp\sandbox\fb004ba8232c3cb7cbdc9927709cbf0b.karma.conf.js -- arg1 arg2"
'E:\karma\bin\karma' is not recognized as an internal or external command,
operable program or batch file.
]`
    - Usually caused by directly invoking `karma` via `exec`/`spawn` (e.g .`spawn('/path/to/karma', ['start'])`)
    - The problem is that Windows has no idea how to handle a Linux executable and would need a `.cmd` file
    - As a solution, we can call `node /path/to/karma start` which will handle loading/invocation
    - For reference, on Windows machines `npm` will create a `.bin/karma.cmd` which is what runs when someone types `karma`
- `Error: Step timed out after 5000 milliseconds`
    - On a VM, things run slower. It's a fact™
    - To resolve the error, we can bump the default timeout to 60 seconds (comically, 1 test is still failing so there might be a deeper problem there)
        - Inside `core_steps.js`, set `this.setDefaultTimeout(60 * 1000);`
        - https://github.com/cucumber/cucumber-js/tree/c884dda4305b4f82fef8df163ee2c9616f374710#timeouts

twolfson:
One more thing, we are generating a CLI string without escaping things like quotes or wrapping strings in quotes which can lead to problems with spaces (e.g. `/home/hello world/karma` would break an exec invocation since we would run `/home/hello`).

https://github.com/karma-runner/karma/blob/v0.13.22/test/e2e/steps/core_steps.js#L53

To remedy that, I typically use `buffered-spawn` which has the same `cmd`/`args` as `spawn` but has a callback of buffers like `exec`:

https://github.com/IndigoUnited/node-buffered-spawn

dignifiedquire:
> To resolve the error, we can bump the default timeout to 60 seconds (comically, 1 test is still failing so there might be a deeper problem there)
> Inside core_steps.js, set this.setDefaultTimeout(60 * 1000);

In theory that should already happen here on latest master: https://github.com/karma-runner/karma/blob/master/test/e2e/support/env.js

dignifiedquire:
> To remedy that, I typically use buffered-spawn which has the same cmd/args as spawn but has a callback of buffers like exec:
> 
> https://github.com/IndigoUnited/node-buffered-spawn

buffered-spawn looks really nice :+1:

twolfson:
I'm running against `0.13.22` to make sure AppVeyor passes when it should

twolfson:
For the past hour, I have been struggling to get the Karma as a dependency of itself working (i.e. `node_modules/karma -> ../`). I tried out

- Copying the directory via Windows CLI commands (FAILED)
    - Although, there was 1 time when it seemed to write the folder improperly and passed so I might exploit that...
    - https://ci.appveyor.com/project/twolfson/karma/build/10
- Using a symlink via the same mechanism as `npm` (FAILED)
    - https://ci.appveyor.com/project/twolfson/karma/build/11
    - https://github.com/npm/npm/blob/f2ae3baf9e4b48282bcfceb94ce861b9221450d0/lib/utils/link.js#L54
    - https://nodejs.org/api/fs.html#fs_fs_symlink_target_path_type_callback
    - https://nodejs.org/docs/latest-v0.10.x/api/fs.html#fs_fs_symlink_srcpath_dstpath_type_callback

twolfson:
Aaand removing the directory failed.

https://ci.appveyor.com/project/twolfson/karma/build/15

I'm getting tired and frustrated. Going to try the magic that worked in https://ci.appveyor.com/project/twolfson/karma/build/10 and if that works, then I will open a PR. Otherwise, I'm calling it a night

twolfson:
Alright, it passed. I'm confused but I blame being tired. Going to open a PR

https://ci.appveyor.com/project/twolfson/karma/build/16

dignifiedquire:
Thanks for all the things!

PR #2002, opened 2016-03-19
Added AppVeyor integration

twolfson:
In #2001, we discussed and decided to add an AppVeyor integration to catch Windows errors before they are landed. In this PR:

- Added AppVeyor integration
- Added AppVeyor badge to README

**Passing build:** https://ci.appveyor.com/project/twolfson/karma/build/16

For lots of helpful Windows pointers, see the thread in #2001.

**Notes:**

This build will initially fail due to #2001 (i.e. we don't properly handle no `customContextFile` on Windows).

PR #2004, opened 2016-03-19
Repaired AppVeyor for Node.js@0.12

twolfson:
As explained in #2002, we thought Node.js@0.12 wouldn't have issues since we had Node.js@5.0 running. Unfortunately, our `xcopy` methodology was space intensive and didn't work for `npm@2` (which is what Node.js@0.12 uses).

We got this repaired via an `npm` upgrade but found out that our previous attempt at symlinks was invalid. Now that we got it working, it's a much better approach than `xcopy`. In this PR:

- Moved from `xcopy` to symlinks for `node_modules/karma`
- Repairs failing AppVeyor builds on `master`

**Passing builds:**

https://ci.appveyor.com/project/twolfson/karma/build/31

### DIFF
--- a/package.json
+++ b/package.json
@@ -367,8 +367,10 @@
     "build": "grunt build",
     "test:appveyor": "grunt test-appveyor",
     "test:integration": "./scripts/integration-tests.sh",
+    "link": "node --eval \"path=require('path'); require('fs').symlinkSync(path.resolve(__dirname), path.resolve(__dirname, 'node_modules', 'karma'), 'junction')\"",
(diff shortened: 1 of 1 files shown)


twolfson:
We are leaving out Node.js@0.10 as we have some actual failures there =X

https://ci.appveyor.com/project/twolfson/karma/build/27/job/ptju2v60t4ikasg3

twolfson:
Hmm, I tried out Node.js@0.10 locally and was unable to reproduce. I think we should be fine since Node.js@0.10 is on the way out and we test it in Travis CI =/

dignifiedquire:
All green, that's how I like my tests!

PR #2006, opened 2016-03-21
Handling Symbol "stringification"

apires:
Currently Stringify fails sending a symbol to the console. This "fixes" symbol stringification. It's not a great way to fix it, but I don't believe the specs give us anything more than what I've been able to find?


apires:
Can I just forgo all rights to the code I've written in this patchset instead of signing this license? I'm not really comfortable sharing my address and phone number.

PR #2018, opened 2016-03-24
docs: Add '--' to 'no-single-run' documentation

stuartmemo:
Before this change, the documentation for singleRun stated that the CLI
commands were "--single-run, no-single-run". "no-single-run" requires
the double dash in order be used.  This commit adds the "--" prefix.


dignifiedquire:
Thanks :octocat:

PR #2044, opened 2016-04-06
chore(package): don't update the included npm in Node

mgol:
Node 0.10.44 updated npm to v2 (as a security update) so there's no need
to update it manually anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,7 @@ env:
   - node_js: "5"
     env: VALIDATE_COMMIT_MSG=TRUE
 
-# Make sure we have a new npm on Node 0.10. npm >= 2 should work so by not updating them
-# here we have proper version coverage, especially that most people use the npm version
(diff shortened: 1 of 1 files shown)


Florian-R:
@mgol This imply that test will be run with npm@2 instead of npm@3 (actual setup) or I miss something?

mgol:
@Florian-R That's true but it's already run with npm@2 on Node 0.12 & 4. It's usually best to test on the npm version that came with Node as most people don't update npm & on Windows it's really hard to do it: a regular `npm -g i npm` will install a newer npm but the included one will still be used anyway unless you dive deep into Windows PATH setup. We just had to update it for Node 0.10 as npm@1 was problematic.

Florian-R:
@mgol Thanks for the insight, was just curious about the rationale. :+1: for testing on common setup.

dignifiedquire:
Great, thanks @mgol :octocat:

PR #2057, opened 2016-04-17
Fixes code block rendering issue in contributing file.

alemangui:
Fixes issue in which a code block inside a list was mistakenly rendered inline with the language name visible.

It is currently rendered like this:
before

Now it's like this:
after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -310,18 +310,21 @@
 
 #### Breaking Changes
 
-* The public api interface has changed to a constructor form. To upgrade
-  change
-  ```javascript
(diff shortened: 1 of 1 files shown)


alemangui:
I signed it!

dignifiedquire:
Thanks, could you please update your commit message to follow our convention: http://karma-runner.github.io/0.13/dev/git-commit-msg.html

alemangui:
@dignifiedquire Is this better? I dont think the _scope_ is applicable for this change.

dignifiedquire:
Thanks :octocat: that's perfect

Issue #1514, opened 2015-07-23
Repeatedly disconnected from Chrome (mac)


wolfgangmeyers:
Chrome recently updated on my mac, and I'm now running Chrome 44.0.2403 (Mac OS X 10.10.3) and have Karma v0.13.3. Sometimes all of my tests will run, but they start breaking when the test runner disconnects from Chrome:
```
Chrome 44.0.2403 (Mac OS X 10.10.3): Executed 14 of 218 SUCCESS (0 secs / 0.248 secs)
23 07 2015 09:49:14.059:WARN [Chrome 44.0.2403 (Mac OS X 10.10.3)]: Disconnected (1 times
Chrome 44.0.2403 (Mac OS X 10.10.3): Executed 14 of 218 DISCONNECTED (10.082 secs / 0.248
Chrome 44.0.2403 (Mac OS X 10.10.3): Executed 15 of 218 DISCONNECTED (10.082 secs / 1.015
Chrome 44.0.2403 (Mac OS X 10.10.3): Executed 16 of 218 DISCONNECTED (10.082 secs / 1.026
Chrome 44.0.2403 (Mac OS X 10.10.3): Executed 17 of 218 DISCONNECTED (10.082 secs / 1.063
Chrome 44.0.2403 (Mac OS X 10.10.3): Executed 18 of 218 DISCONNECTED (10.082 secs / 1.082
```

Dignifiedquire:
Is Chrome in the background or foreground when that happens?

wolfgangmeyers:
In the background. This typically happens when I make changes to a code or test file in my IDE, causing the test run to fire.

wolfgangmeyers:
To elaborate, I notice that the tests always seem to go through fine when I launch Karma, but every time a file change happens it starts running the tests and then hangs for a bit, and complains about a browser disconnect.

I'm also using some other components with Karma like code coverage. Here is my config file:

```
// Karma configuration
// Generated on Thu May 14 2015 15:11:29 GMT-0700 (PDT)

module.exports = function(config) {
  config.set({

    // base path that will be used to resolve all patterns (eg. files, exclude)
    basePath: '../..',


    // frameworks to use
    // available frameworks: https://npmjs.org/browse/keyword/karma-adapter
    frameworks: ['jasmine'],


    // list of files / patterns to load in the browser
    files: [
      
      ...
    (all of my code files here...)
      ...

      'assets/js/tests/*.js',
      
      // templates
      'templates/*.html'
[...]

Issue #598, opened 2013-06-25
random DISCONNECTED messages from PhantomJS


ibawt:
Hi,
  Karma: 0.8.6
  PhantomJS: 1.9
  Node: v0.10.2, v0.10.7

  I'm seeing random disconnect messages in single run mode.  Is this a known problem?  I tried monkey patching the socket.io transports to only do xhr and jsonp but it still happens.  Any ideas?

mtscout6:
I've noticed that using pre-processors can cause this to increase as well. Since karma launches the browser then processes all the files. If your preprocessors take a long time to run then I'll bet you see a direct correlation.

nuweb:
I get the same error.
OSX - 10.9.5
Karma version: 0.12.31
phantomjs version 1.9.8
node --version 0.10.33

->grunt karma
Running "karma:unit" (karma) task
INFO [karma]: Karma v0.12.31 server started at http://localhost:9876/
INFO [launcher]: Starting browser PhantomJS
INFO [PhantomJS 1.9.8 (Mac OS X)]: Connected on socket vo0VCply3aIvM70g5H5C with id 93719535
PhantomJS 1.9.8 (Mac OS X): Executed 1 of 9 SUCCESS (0 secs / 0.001 secs)
PhantomJS 1.9.8 (Mac OS X): Executed 1 of 9 DISCONNECTED (12.386 secs / 0.001 secs)
Warning: Task "karma:unit" failed. Use --force to continue.

Aborted due to warnings.

bajtos:
Adding the following lines to my karma config solved the problem for me:

```diff
     // If browser does not capture in given timeout [ms], kill it
     captureTimeout: 60000,

+    // to avoid DISCONNECTED messages
+    browserDisconnectTimeout : 10000, // default 2000
+    browserDisconnectTolerance : 1, // default 0
+    browserNoActivityTimeout : 60000, //default 10000
```

Source: webpack/karma-webpack#8

sloria:
@bajtos' workaround worked for me.

I am only seeing this error when I run in --single-run mode.

veeracs:
@bajtos workaround did not work for me. I'm consistently getting DISCONNECTED error.

anijap:
@veeracs, I kept getting the same error on a build server with 3.5GB of RAM.
The PhantomJS startup time was consistently around >0.9sec && <=2sec over five builds.

I've increased the *browserNoActivityTimeout* to 3 seconds as a workaround.

[...]

PR #2127, opened 2016-05-10
Allow browserDisconnectTolerance to be set in non-singleRun mode.

dtinth:
We’re encountering a problem where sometimes Karma just disconnects and stops running tests until we restart Karma manually.

We found that some of our JavaScript code intermittently gets stuck when running tests — that Karma considers it a time out. One such example is an infinite loop in the code.

Looking at other issues (https://github.com/karma-runner/karma/issues/598, https://github.com/karma-runner/karma/issues/1514) suggests adding `browserDisconnectTolerance`. I’ve tried that and several other options to no avail.

From further source code inspection, in 19590e1f66fd6c3b0d3fc9e90000c705198e0e70 it turns out that __`browserDisconnectTolerance` is only active in `singleRun` mode.__ This pull request makes `browserDisconnectTolerance` work in non-singleRun mode.

## Test case

Here’s a test case to verify that my patch works:

### test.js

This test will get stuck in an infinite loop with 75% chance.

```js
describe('example', function () {
  it('ok1', function () { })
  it('ok2', function () { })
  it('bad1', function () {
    if (Math.random() > 0.5) for (;;) ;
  })
  it('bad2', function () {
    if (Math.random() > 0.5) for (;;) ;
  })
  it('ok3', function () { })
})
```

### karma.conf.js

Sets a ridiculously high amount of retries.

```js
module.exports = function (config) {
  config.set({
    basePath: '',
    frameworks: ['mocha'],
    files: [
      'test.js'
    ],
    reporters: [ 'progress' ],
    port: 9988,
    colors: true,
    logLevel: config.LOG_INFO,
    autoWatch: true,
    browsers: [ 'PhantomJS' ],
    browserDisconnectTolerance: 99999999999999999999999999999999999999999999999,
    retryLimit: 999999999999999999999999999999999999999999999999999999999999999,
    plugins: [
      'karma-mocha',
      'karma-phantomjs-launcher',
    ],
    singleRun: false,
  })
}
```

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -145,11 +145,6 @@ var normalizeConfig = function (config, configFilePath) {
     config.autoWatch = false
   }
 
-  if (!config.singleRun && config.browserDisconnectTolerance) {
-    log.debug('browserDisconnectTolerance set to 0, because of singleRun')
-    config.browserDisconnectTolerance = 0
-  }
-
   if (helper.isString(config.reporters)) {
     config.reporters = config.reporters.split(',')
   }
--- a/lib/server.js
+++ b/lib/server.js
@@ -292,26 +292,34 @@ Server.prototype._start = function (config, launcher, preprocess, fileList, webS
     }
   }
 
-  if (config.singleRun) {
-    self.on('browser_complete', function (completedBrowser) {
-      if (completedBrowser.lastResult.disconnected &&
-        completedBrowser.disconnectsCount <= config.browserDisconnectTolerance) {
-        self.log.info('Restarting %s (%d of %d attempts)', completedBrowser.name,
-          completedBrowser.disconnectsCount, config.browserDisconnectTolerance)
-        if (!launcher.restart(completedBrowser.id)) {
-          singleRunDoneBrowsers[completedBrowser.id] = true
-          emitRunCompleteIfAllBrowsersDone()
-        }
-      } else {
-        singleRunDoneBrowsers[completedBrowser.id] = true
+  self.on('browser_complete', function (completedBrowser) {
+    if (completedBrowser.lastResult.disconnected &&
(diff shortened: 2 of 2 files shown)


dtinth:
My employer has signed the CLA for me.

On a related note, new version of `eslint-config-standard` broke the builds, because this repo gets a __minor__ version bump when a new __major__ version of standard is released. I believe that this problem is out of the scope of this pull request.

dignifiedquire:
Thanks, could you please update your commit message to follow our convention: http://karma-runner.github.io/0.13/dev/git-commit-msg.html

dtinth:
Thanks for your reply. Will do it tomorrow.

dtinth:
I’ll put a draft commit message here in case you would like to comment on it before I do the actual amendment tomorrow:

```
fix(web-server): Restart disconnected browser in non-singleRun mode.

Allow browserDisconnectTolerance config option to be set in non-singleRun mode.

In some cases, the browser would get stuck in an infinite loop (e.g. because of
a faulty code/test). This blocks browser’s event loop, preventing it from
reporting back to Karma. Karma then considers the browser ‘DISCONNECTED’.

Prior to this commit, the `browserDisconnectTolerance` option will only apply
in singleRun mode. In above scenario, the end-user must restart Karma to
continue running tests inside Karma-managed browsers. This commit fixes this
problem by always honoring the aforementioned option.
```

I marked this as a “fix” because the [configuration docs](http://karma-runner.github.io/0.13/config/configuration-file.html) did not mention that this option will not apply in non-singleRun mode. Please let me know if it needs changing.

dignifiedquire:
Thanks, commit message looks good to me. Can you please also rebase onto latest master as that should make CI happy again

dtinth:
Thanks for the feedback. Rebased. 😉

dignifiedquire:
Thanks :octocat:

PR #2149, opened 2016-05-25
feat(reporter): improve source map handling and reporting.

mprobst:
- Avoid prepending basepath to every file path.
  The original path is usually relative to the project, and much more meaningful
  than basepath + path, which is usually a long, absolute path.
- Format as [mapped location] <- [original location]
  The mapped location is the more important part for users, it's the source
  file they are working on.
- Resolve source map file paths relative to the original file path.
  It's common for tools that generate a source map to not have a complete
  understanding of what paths they should generate, in which case they just use
  the local path. An example tool is TypeScript.
  Resolving the path against the original file path makes these paths
  unambiguous in the project.

### DIFF
--- a/lib/reporter.js
+++ b/lib/reporter.js
@@ -1,4 +1,5 @@
 var util = require('util')
+var resolve = require('url').resolve
 var log = require('./logger').create('reporter')
 var MultiReporter = require('./reporters/multi')
 var baseReporterDecoratorFactory = require('./reporters/base').decoratorFactory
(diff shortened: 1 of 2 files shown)


mprobst:
@zzo PTAL

maksimr:
@mprobst thanks for PR

Please look at failed tests - https://ci.appveyor.com/project/dignifiedquire/karma/build/296/job/0ukhlfgwcm548tjg

mprobst:
@maksimr @dignifiedquire Thanks for looking into this! I fixed the test failure (grumble grumble windows grumble).

zzo:
thanks martin looks good

dignifiedquire:
Thanks :octocat: sourcemaps can always use more love

PR #2154, opened 2016-05-28
chore: resetup wallabyjs for babel

maksimr:


lib/middleware/karma.js
```diff
@@ -232,5 +232,14 @@ var createKarmaMiddleware = function (
...
 
+createKarmaMiddleware.$inject = [
+  'filesPromise',
+  'serveStaticFile',
+  'serveFile',
+  'injector',
+  'config.basePath',
+  'config.urlRoot'
+]
+
```
dignifiedquire: Why do we need this?

dignifiedquire:
needs rebase onto `master`

lib/middleware/karma.js
maksimr: @dignifiedquire wallaby can't parse special comment like   /* config.basePath */ in arguments of the function

dignifiedquire: makes sense lets drop the comments then, as with this we double define things

maksimr:
@dignifiedquire done

PR #2155, opened 2016-05-28
fix(client): does not throws an error for non DOM object that has `ta…

### LINKED ISSUE #2139: console.log throws an error for non DOM object that has 'tagName' property
Karma overwrites native console.log and tries to stringify every attribute it gets passed into it so it can create an output. In the process if it encounters a DOM node or a DOM Element it attempts to serialize it https://github.com/karma-runner/karma/blob/master/client/stringify.js#L41 

The problem is that detection of whether something is a DOM Node is very optimistic. I'm faced with a problem when trying to console.log a Backbone.View ( or any view that inherits form a Backbone.View ) because every Backbone.View has 'tagName' attribute, and this causes the following error in Karma because it expects a DOM node, but gets a Backbone.View in stead.

`PhantomJS 2.1.1 (Linux 0.0.0) ERROR
  TypeError: undefined is not a function (evaluating 'target.dispatchEvent(e)')
  at http://localhost:9877/karma.js:1363`

Possible solution is to change the way of checking if something is a DOM Node / Element http://stackoverflow.com/a/4754156/3193466

### Expected behavior

` var V = new Backbone.View(); console.log( V )`
`// Will output the View object in the console`

### Actual behavior

` var V = new Backbone.View(); console.log( V )`
`PhantomJS 2.1.1 (Linux 0.0.0) ERROR
  TypeError: undefined is not a function (evaluating 'target.dispatchEvent(e)')
  at http://localhost:9877/karma.js:1363``
### Enviroment Details

- Karma version (output of `karma --version`): Karma version: 0.13.22
- Relevant part of your `karma.config.js` file: Nothing specific

1. Instantiate a Backbone.View ( Or use any object that has 'tagName' attribute' )
2. console.log the instance

maksimr:
…gName` property

fixes #2139

### DIFF
--- a/client/stringify.js
+++ b/client/stringify.js
@@ -1,5 +1,8 @@
 var serialize = require('dom-serialize')
 var instanceOf = require('./util').instanceOf
+var isNode = function (obj) {
+  return (obj.tagName || obj.nodeName) && obj.nodeType
+}
 
 var stringify = function stringify (obj, depth) {
   if (depth === 0) {
@@ -38,7 +41,7 @@ var stringify = function stringify (obj, depth) {
         return '<!--' + obj.nodeValue + '-->'
       } else if (obj.outerHTML) {
         return obj.outerHTML
-      } else if (obj.tagName || obj.nodeName) {
+      } else if (isNode(obj)) {
         return serialize(obj)
       } else {
         var constructor = 'Object'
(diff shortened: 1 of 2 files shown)


dignifiedquire:
thanks. could you add a test for this please?

maksimr:
@dignifiedquire sure

maksimr:
@dignifiedquire done

dignifiedquire:
Thanks :octocat:

PR #2183, opened 2016-06-23
Allow frameworks to inject middleware

### LINKED ISSUE #1110: Allow proxy config changes by plugins
I have a plugin that would like to specify some proxies. Adding them to the configuration doesn't work because the proxies are bound early and my plugin injected to late. Call graph:

[middleware/proxy.js:66](https://github.com/karma-runner/karma/blob/3873c531/lib/middleware/proxy.js#L66)
[middleware/proxy.js:126](https://github.com/karma-runner/karma/blob/3873c531/lib/middleware/proxy.js#L126)
[web-server.js:54](https://github.com/karma-runner/karma/blob/3873c531/lib/web-server.js#L54)
[web-server.js:46](https://github.com/karma-runner/karma/blob/3873c531/lib/web-server.js#L46)
[web-server.js:29](https://github.com/karma-runner/karma/blob/3873c531/lib/web-server.js#L29)
[web-server.js:74](https://github.com/karma-runner/karma/blob/3873c531/lib/web-server.js#L74)
[server.js:288](https://github.com/karma-runner/karma/blob/3873c531/lib/server.js#L288)

The problem is that the injector runs my plugin after the [`webServer: ['factory', ws.create],`](https://github.com/karma-runner/karma/blob/3873c531/lib/server.js#L288) line which where the proxies are bound.

There is a [`// TODO(vojta): extract the proxy into a plugin`](https://github.com/karma-runner/karma/blob/3873c531/lib/web-server.js#L53), so this is obviously something that is to be done, can I help out now?

All links are to commit 3873c531, so that they don't rot overtime.

jamestalmage:
The `Server.prototype._start` function was injected with `webServer` and `socketServer`. This prevents frameworks from being able to manipulate `config.middleware` before those servers are created.

The solution is simply to defer injection until after all frameworks have loaded, and then fetch those values with `injector.get`.

There were no e2e tests involving middleware. I have added two, including one that verifies the new capabilities.

The second commit adds a test verifying that `config.proxies` can be configured by frameworks now as well (Fixes #1110)

### DIFF
--- a/test/e2e/pass-opts.feature
+++ b/test/e2e/pass-opts.feature
@@ -19,5 +19,4 @@ Feature: Passing Options
     Then it passes with no debug:
       """
       .
-      PhantomJS
       """
--- a/lib/server.js
+++ b/lib/server.js
@@ -129,8 +129,8 @@ Server.prototype.refreshFiles = function () {
 // Private Methods
 // ---------------
 
-Server.prototype._start = function (config, launcher, preprocess, fileList, webServer,
-                                    capturedBrowsers, socketServer, executor, done) {
+Server.prototype._start = function (config, launcher, preprocess, fileList,
+                                    capturedBrowsers, executor, done) {
   var self = this
   if (config.detached) {
     this._detach(config, done)
@@ -142,6 +142,9 @@ Server.prototype._start = function (config, launcher, preprocess, fileList, webS
   config.frameworks.forEach(function (framework) {
     self._injector.get('framework:' + framework)
   })
+
+  var webServer = self._injector.get('webServer')
+  var socketServer = self._injector.get('socketServer')
 
   // A map of launched browsers.
   var singleRunDoneBrowsers = Object.create(null)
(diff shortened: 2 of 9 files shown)


test/e2e/pass-opts.feature
```diff
@@ -19,5 +19,4 @@ Feature: Passing Options
     Then it passes with no debug:
       """
       .
-      PhantomJS
```
jamestalmage: It is not immediately clear to me why my changes cause the `PhantomJS` line not to be printed for this one particular test.

jamestalmage:
Hmm.

It seems AppVeyor fails during `npm install`. Perhaps related to me adding `express` as a dev-dependency?

dignifiedquire:
Thanks, this looks really good :) Could you squash the latest commit into the one before, then I can merge it

jamestalmage:
Done.

dignifiedquire:
Will merge when ci is green

jamestalmage:
Woot! Green!

dignifiedquire:
Thanks :octocat:

jamestalmage:
I see this just barely missed the `1.1.0` release. Will you be releasing again soon?

dignifiedquire:
Yes, I hope to get an answer on #2186 soon and then do a new release.

PR #2202, opened 2016-06-26
terminated statements in JS template for karma config

Mobiletainment:


Mobiletainment:
I signed it!

dignifiedquire:
Thanks but we don't use semicolons in karma that's why there are not in there.

PR #2212, opened 2016-06-27
feat: upstreamProxy config option to deal with proxies

pghalliday:
I have a use case where I can only load content via a web proxy that I don't control. This web proxy prepends a directory to the front of URLs which breaks the internal urls inserted by the Karma middleware.

This pull request adds an `upstreamProxy` option that allows the additional proxy path to be configured and used in URLs. To complete the use cases and implement e2e tests I also added `hostname`, `port` and `protocol` fields to use in launchers to point them at the proxy.

This is my first contribution here so please be gentle ;)

### DIFF
--- a/docs/config/01-configuration-file.md
+++ b/docs/config/01-configuration-file.md
@@ -647,6 +647,44 @@ is handed off to [socket.io](http://socket.io/) (which manages the communication
 between browsers and the testing server).
 
 
+## upstreamProxy
+**Type:** Object
+
+**Default:** `undefined`
+
+**Description:** For use when the Karma server needs to be run behind a proxy that changes the base url, etc
+
+If set then the following fields will be defined and can be overriden:
+
+### path
+**Type:** String
+
+**Default:** `'/'`
+
+**Description:** Will be prepended to the base url when launching browsers and prepended to internal urls as loaded by the browsers
+
+### port
+**Type:** Number
+
+**Default:** `9875`
+
+**Description:** Will be used as the port when launching browsers
+
+### hostname
+**Type:** String
+
+**Default:** `'localhost'`
+
+**Description:** Will be used as the hostname when launching browsers
+
+### protocol
+**Type:** String
+
+**Default:** `'http:'`
+
+**Description:** Will be used as the protocol when launching browsers
+
+
 ## urlRoot
 **Type:** String
 
--- a/lib/config.js
+++ b/lib/config.js
@@ -72,22 +72,36 @@ var createPatternObject = function (pattern) {
   return new Pattern(null, false, false, false, false)
 }
 
+var normalizeUrl = function (url) {
+  if (url.charAt(0) !== '/') {
+    url = '/' + url
+  }
+
+  if (url.charAt(url.length - 1) !== '/') {
+    url = url + '/'
+  }
+
+  return url
+}
+
 var normalizeUrlRoot = function (urlRoot) {
-  var normalizedUrlRoot = urlRoot
-
-  if (normalizedUrlRoot.charAt(0) !== '/') {
-    normalizedUrlRoot = '/' + normalizedUrlRoot
-  }
-
-  if (normalizedUrlRoot.charAt(normalizedUrlRoot.length - 1) !== '/') {
-    normalizedUrlRoot = normalizedUrlRoot + '/'
-  }
+  var normalizedUrlRoot = normalizeUrl(urlRoot)
 
   if (normalizedUrlRoot !== urlRoot) {
     log.warn('urlRoot normalized to "%s"', normalizedUrlRoot)
   }
 
   return normalizedUrlRoot
+}
+
+var normalizeProxyPath = function (proxyPath) {
+  var normalizedProxyPath = normalizeUrl(proxyPath)
+
+  if (normalizedProxyPath !== proxyPath) {
+    log.warn('proxyPath normalized to "%s"', normalizedProxyPath)
+  }
+
+  return normalizedProxyPath
 }
 
 var normalizeConfig = function (config, configFilePath) {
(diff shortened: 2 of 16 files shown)


docs/config/01-configuration-file.md
```diff
@@ -617,6 +617,44 @@ is handed off to [socket.io](http://socket.io/) (which manages the communication
 between browsers and the testing server).
 
 
+## upstreamProxy
+**Type:** Object
+
+**Default:** `undefined`
```
dignifiedquire: Isn't the default he object containing the default values of the options listed below?

lib/config.js
```diff
@@ -127,6 +141,22 @@ var normalizeConfig = function (config, configFilePath) {
   // normalize urlRoot
   config.urlRoot = normalizeUrlRoot(config.urlRoot)
 
+  // normalize and default upstream proxy settings if given
+  if (config.upstreamProxy) {
+    var proxy = config.upstreamProxy
+    proxy.path = proxy.path ? normalizeProxyPath(proxy.path) : '/'
+    proxy.hostname = proxy.hostname || 'localhost'
```
dignifiedquire: this should be a check for `_.isUndefined(proxy.hostname)` as this precludes someone from setting the `hostname` to `''`

lib/config.js
```diff
@@ -127,6 +141,22 @@ var normalizeConfig = function (config, configFilePath) {
...
+    proxy.port = proxy.port || 9875
```
dignifiedquire: this should be a check for `_.isUndefined(proxy.port)` as this precludes someone from setting the `port` to `0`

lib/config.js
```diff
@@ -127,6 +141,22 @@ var normalizeConfig = function (config, configFilePath) {
   // normalize urlRoot
   config.urlRoot = normalizeUrlRoot(config.urlRoot)
 
+  // normalize and default upstream proxy settings if given
+  if (config.upstreamProxy) {
+    var proxy = config.upstreamProxy
+    proxy.path = proxy.path ? normalizeProxyPath(proxy.path) : '/'
```
dignifiedquire: this should be a check for `_.isUndefined(proxy.path)` as this precludes someone from setting the `path` to `''`

lib/launcher.js
```diff
@@ -39,10 +39,13 @@ var Launcher = function (server, emitter, injector) {
...
   }
 
-  this.launchSingle = function (protocol, hostname, port, urlRoot) {
+  this.launchSingle = function (protocol, hostname, port, urlRoot, upstreamProxy) {
     var self = this
     return function (name) {
       var url = protocol + '//' + hostname + ':' + port + urlRoot
+      if (upstreamProxy) {
+        url = upstreamProxy.protocol + '//' + upstreamProxy.hostname + ':' + upstreamProxy.port + upstreamProxy.path + urlRoot.substr(1)
+      }
```
dignifiedquire: Maybe combine this to sth like this:

```js
if (upstreamProxy) {
  protocol = upstreamProxy.protocol
  hostname = upstreamProxy.hostname
  port = upstreamPoxy.Port
  urlRoot = upstreamProxy.path + urlRoot
}

var url = protocol + '//' + hostname + ':' + port + urlRoot
```

test/e2e/steps/core_steps.js
```diff
@@ -71,71 +72,102 @@ module.exports = function coreSteps () {
...
-          }, done)
-        }
-        var runOut = command === 'runOut'
-        if (command === 'run' || command === 'runOut') {
-          _this.child = spawn('' + runtimePath, ['start', '--log-level', 'warn', configFile])
-          var done = function () {
-            cleansingNeeded = true
-            _this.child && _this.child.kill()
-            callback()
+  this.When(/^I (run|runOut|start|init|stop) Karma( with log-level ([a-z]+))?( behind a proxy on port ([0-9]*) that prepends '([^']*)' to the base path)?$/, function (command, withLogLevel, level, behindProxy, proxyPort, proxyPath, callback) {
```
dignifiedquire: This file is getting really large, maybe put this step into its own file?

dignifiedquire:
@pghalliday thanks, this looking really good :) some small nitpicks but other than that great work !

lib/config.js
pghalliday: ok I'll update this

lib/config.js
pghalliday: I'll update this too

lib/config.js
pghalliday: and again :)

lib/launcher.js
pghalliday: agreed, i'll update this

test/e2e/steps/core_steps.js
pghalliday: I was worried about that too, but I wasn't quite sure about everything going on here and didn't want to duplicate it. Not sure the best way to break it up - i'll give it some thought

pghalliday:
Note that I just raised another PR to solve the same problem but differently (#2213). I'm not sure which would be preferred. Personally I'm a little conflicted. The other PR that converts everything to relative URLs seems cleaner but the approach here seems more explicit

docs/config/01-configuration-file.md
pghalliday: No the default is undefined as the launcher URL is not affected when it is undefined

test/e2e/steps/core_steps.js
pghalliday: Ah I misread - I was worrying about the size of the step you're worrying about the size of the file. That's easier to deal with :)

pghalliday:
I didn't split up the core_steps.js - my initial attempts seemed to open something of a rabbit hole. However I did refactor the proxy stuff after figuring out the World object and After hooks

dignifiedquire:
Sorry for the delay, could you rebase this onto the latest master please?

pghalliday:
no problem, rebased

pghalliday:
Gah, that travis failure on Node 5 doesn't happen locally :( Also I don't see how it's related to this change

dignifiedquire:
I restarted it, lets see if it gets happier

pghalliday:
thx, that fixed it :)

dignifiedquire:
Thank you :octocat:

pghalliday:
Cool, thanks :)

PR #2213, opened 2016-06-28
feat: use relative URLs internally

pghalliday:
This is an alternative approach to that taken in #2212. It solves the same problem, however it may be more flexible. It may have other issues particularly with regard to URLs for directories that don't have a trailing slash, I don't think any such URLs are generated internally though


pghalliday:
Hmm, looking into the CI failures

pghalliday:
Seems to break on integration tests, specifically with requirejs. Maybe better to go with the more explicit fix in #2212 at least that change does not change anything if the option is not used. This relative url approach does change the behaviour subtly in all cases

pghalliday:
Not sure why the AppVeyor build fails, seems to bomb out after npm install and before doing anything. Doesn't happen on the other branch though :s

client/main.js
```diff
@@ -6,14 +6,30 @@ var Karma = require('./karma')
...
 var util = require('../common/util')
 
-var KARMA_URL_ROOT = require('./constants').KARMA_URL_ROOT
+var KARMA_SOCKET_RELATIVE_URL = require('./constants').KARMA_SOCKET_RELATIVE_URL
+
+// Construct the absolute path to the socket as socket.io
+// doesn't like relative paths.
+// Also we constructed the socket.io path
+// so we know it will only contain .. sequences followed by a path
+var baseParts = location.pathname.split('/')
```
dignifiedquire: This will be problematic as older browser don't have the `location` object

dignifiedquire:
I agree with you lets go with #2212 that seems to be the safer approach and makes sure that things don't suddenly change for users.

pghalliday:
Cool, closing this one :)

PR #2219, opened 2016-06-29
fix: Add crossorigin attribute to script HTML tags

pardoman:
**The Problem**
When an error occurs in a script that is not served by the
default server spawned by karma, the stack trace gets lost and
only the message `"Script Error"` is captured and reported back.
This makes it hard to pin-point where the error originated from.

**The Solution**
Add `crossorigin` attribute to generated script tags.

Solution is inspired by this article:
https://blog.getsentry.com/2016/05/17/what-is-script-error.html

### DIFF
--- a/lib/middleware/karma.js
+++ b/lib/middleware/karma.js
@@ -27,7 +27,7 @@ var urlparse = function (urlStr) {
 var common = require('./common')
 
 var VERSION = require('../constants').VERSION
-var SCRIPT_TAG = '<script type="%s" src="%s"></script>'
+var SCRIPT_TAG = '<script type="%s" src="%s" crossorigin="anonymous"></script>'
 var LINK_TAG_CSS = '<link type="text/css" href="%s" rel="stylesheet">'
 var LINK_TAG_HTML = '<link href="%s" rel="import">'
 var SCRIPT_TYPE = {
--- a/test/unit/middleware/karma.spec.js
+++ b/test/unit/middleware/karma.spec.js
@@ -197,7 +197,7 @@ describe('middleware.karma', () => {
 
     response.once('end', () => {
       expect(nextSpy).not.to.have.been.called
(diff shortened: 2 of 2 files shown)


pardoman:
To help better explain what is being fixed here, see the following images.

Here's the issue I'm experiencing due to running tests from another server (which given my setup, I require this to be the case). Notice the "Script Error" output circled in red.
image

With the code in this PR, I was able to get the full stack trace of the error:
image

pardoman:
I signed it!

dignifiedquire:
Thanks for this. Can you please change the commit message to follow our [convention](http://karma-runner.github.io/1.0/dev/git-commit-msg.html)

pardoman:
Sure thing, let me try that.

pardoman:
@dignifiedquire updated

pardoman:
Hmm, I see errors in tests. I'll take a look

pardoman:
I've updated the unit tests

I still haven't figured out how to update the e2e tests for `cucumberjs:ci` which are still failing (locally)

pardoman:
Looks like all checks passed.

dignifiedquire:
Thanks, just one last thing can you squash the two commits into one?

pardoman:
Done, let's wait for the tests to run again. I've also rebased my branch.

dignifiedquire:
Thanks :octocat:

pardoman:
My pleasure 👍

Krinkle:
In retrospect, this "bug fix" released in a patch version (v1.1.1) is in fact an entirely new restriction that is imho considered a breaking change. Anyone using Karma with one of the test files served from another server needs to add `crossOriginAttribute: false` to the configuration. (And that option doesn't exist until v1.4.0).

This happens to coincide with regression fix https://github.com/karma-runner/karma/pull/2220 for https://github.com/litixsoft/karma-mocha-reporter/issues/67.

Upgrading to v1.1.1 fixed that but then broke everything else. The only way to get it fixed was to update directly to v1.4.0.

This could've been avoided if the option was disabled by default and kept opt-in until a later semver-minor or semver-major release.

Alternatively, it would've been avoided if it was supported to have Karma serve one of `conf.files` from a path in `conf.proxies`. That way it would've come from the same origin by default.

PR #2220, opened 2016-06-30
fix(reporter): inject correct config option

4kochi:
* `'config.browserLogOptions'` -> `'config.browserConsoleLogOptions'`


4kochi:
I signed it!

dignifiedquire:
Thanks so much for this

Issue #1727, opened 2015-11-26
Allow for karma.conf.ts (TypeScript) when using cli


tkrotoff:
When running `node_modules/.bin/karma start karma.conf.ts` (written in TypeScript) I get:
```
ERROR [config]: Invalid config file!
  SyntaxError: Unexpected reserved word
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:413:25)
    at Object.Module._extensions..js (module.js:452:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.parseConfig (node_modules/karma/lib/config.js:276:22)
    at new Server (node_modules/karma/lib/server.js:56:20)
    at Object.exports.run (node_modules/karma/lib/cli.js:231:7)
    at Object.<anonymous> (node_modules/karma/bin/karma:3:23)
    at Module._compile (module.js:434:26)
    at Object.Module._extensions..js (module.js:452:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:475:10)
```

Also it does not seem to automatically detect `karma.conf.ts` when running `node_modules/.bin/karma start`:
```
[karma]: No captured browser, open http://localhost:9876/
INFO [karma]: Karma v0.13.15 server started at http://localhost:9876/
INFO [Chrome]: Connected on socket 9S-HxtcaP6-xFCRfAAAA with id manual-8425
Chrome ERROR
  You need to include some adapter that implements __karma__.start method!
```

On the other hand when running the task using gulp and the same `karma.conf.ts`:
```TypeScript
import * as karma from 'karma';

function runKarma(done): void {
  const config = {
    configFile: __dirname + '/karma.conf.ts'
  };

  const server = new karma.Server(config, done);
  server.start();
}

gulp.task('run', done => runKarma(done));
```

Everything works as expected so I suspect code specific to the cli to not support TypeScript.

Dignifiedquire:
I'm afraid Karma does not support TypeScript at all in configuration files. Not sure why it works when you use it through gulp (maybe some part of your build process handles it somehow, or more likely it is just processed by node with support for ES2015 instead of TypeScript, as your config files looks like it does not use anything that is actually TypeScript specific only ES2015 modules)

tkrotoff:
@Dignifiedquire

> it is just processed by node

[...]

PR #2224, opened 2016-07-01
feat(config): add support for TypeScript

wesleycho:
- Add TypeScript support for `karma.conf.ts`

This should address #1727.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -94,10 +94,14 @@ var processArgs = function (argv, options, fs, path) {
       configFile = './karma.conf.js'
     } else if (fs.existsSync('./karma.conf.coffee')) {
       configFile = './karma.conf.coffee'
+    } else if (fs.existsSync('./karma.conf.ts')) {
+      configFile = './karma.conf.ts'
     } else if (fs.existsSync('./.config/karma.conf.js')) {
       configFile = './.config/karma.conf.js'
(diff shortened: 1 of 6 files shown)


lib/cli.js
```diff
@@ -94,10 +94,14 @@ var processArgs = function (argv, options, fs, path) {
       configFile = './karma.conf.js'
     } else if (fs.existsSync('./karma.conf.coffee')) {
       configFile = './karma.conf.coffee'
+    } else if (fs.existsSync('./karma.conf.ts')) {
+      configFile = './karma.conf.ts'
     } else if (fs.existsSync('./.config/karma.conf.js')) {
       configFile = './.config/karma.conf.js'
     } else if (fs.existsSync('./.config/karma.conf.coffee')) {
       configFile = './.config/karma.conf.coffee'
+    } else if (fs.existsSync('./config/karma.conf.ts')) {
```
maksimr: Here should be `.config`

dignifiedquire:
Would be good to have some test for this.

wesleycho:
Any tip as to where to look for putting a test? I didn't see anything for coffeescript or livescript, and the way they are set up, it may not be so simple to set up since support is checked globally.

wesleycho:
Found it, added a test. I also added a config template and code to the formatters to mirror the coffeescript and livescript code, although it seems they are not used.

dignifiedquire:
Thanks :octocat:

Issue #2140, opened 2016-05-17
Add support for range requests to the server


zhuker:
### Expected behavior

Files served by karma http server should support Range: bytes=42-43

### Actual behavior

Entire file is output

### Enviroment Details
Karma version: 0.13.22

- Karma version (output of `karma --version`):
- Relevant part of your `karma.config.js` file

### Steps to reproduce the behaviour

1. Write in a test
```javascript
var xhr = new XMLHttpRequest();
xhr.open("GET", "/base/big_binary_file.dat", true);
xhr.setRequestHeader("Range", "bytes=42-43");
xhr.responseType =  "arraybuffer";
xhr.onload = () => {
expect(xhr.response.byteLength).toBe(2) 
}
```
2. Run test
3. Test fails

dignifiedquire:
Thanks for the report, it's an interesting feature so if someone wants to implement it I'd be happy to accept a PR for it.

PR #2226, opened 2016-07-01
feat(web-server): Allow Range headers in web server.

TheModMaker:
Adds the ability to use Range headers in the web server (#2140).

### DIFF
--- a/docs/config/02-files.md
+++ b/docs/config/02-files.md
@@ -126,7 +126,15 @@ proxies: {
 },
 ```
 
+## Webserver features
+
+* [Range requests][].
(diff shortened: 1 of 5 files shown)


dignifiedquire:
Thanks, could you also add a note to the documentation so users know this exists and works please?

TheModMaker:
I couldn't find an obvious place to add to the documentation, so I added a section in the Files file.

Also, re-reading the spec, my handling of multiple ranges was incorrect.  So I removed support for multiple ranges since it would be more complicated to add it.

dignifiedquire:
Thanks :octocat:

PR #2222, opened 2016-06-30
fix(plugin): can not load plugins

zincli:
can not load plugins while karma installed in sub dir in node_modules

Close [#2221](https://github.com/karma-runner/karma/issues/2221)

PR #2229, opened 2016-07-02
fix(plugin): fix plugin loading bug #2222

zincli:
Closes #2222


zincli:
wrong commit info ...

PR #2252, opened 2016-07-13
fix(context): Updated postMessage listener to stop validating non-Karma messages

### LINKED ISSUE #2239: Received `postMessage` for "undefined" but the method doesn't exist
### Expected behavior

Projects that use postMessage should not break karma

### Actual behavior

This change makes karma incompatible with projects that use postMessage https://github.com/karma-runner/karma/commit/81b1f04c70eb14ca675f60b3f374bb50b96d80fe#diff-e935b6c040afc12b7a014e625f4f0cedR40

### Enviroment Details

https://bitbucket.org/atlassian/simple-xdm/addon/pipelines/home#!/results/%7Bcc56070d-5ec1-43f2-ab8e-07cf4e6309a5%7D

- Karma version (output of `karma --version`): 1.1.1
- Relevant part of your `karma.config.js` file N/A

1. Write test that uses postMessage
2. Run test with karma
3. Karma receives postMessage it doesn't expect and balks

Consider changing log to warn rather than error or namespacing your postMessages

twolfson:
As reported in #2239, our current `postMessage` hook is looking at all post messages, not only the ones for Karma. In this PR:

- Moved to namespaced `postMessage` variable checks
- Updated `postMessage` logic to only assert if a truthy method was given
- Fixes #2239 

/cc @dignifiedquire

### DIFF
--- a/client/karma.js
+++ b/client/karma.js
@@ -35,12 +35,14 @@ var Karma = function (socket, iframe, opener, navigator, location) {
       }
 
       // Take action based on the message type
-      var method = evt.data.method
-      if (!self[method]) {
-        self.error('Received `postMessage` for "' + method + '" but the method doesn\'t exist')
-        return
+      var method = evt.data.__karmaMethod
+      if (method) {
+        if (!self[method]) {
+          self.error('Received `postMessage` for "' + method + '" but the method doesn\'t exist')
+          return
+        }
+        self[method].apply(self, evt.data.__karmaArguments)
       }
-      self[method].apply(self, evt.data.arguments)
     }, false)
   }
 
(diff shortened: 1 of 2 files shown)


dignifiedquire:
Thanks looks good to me. Any ideas yet on how to add a test for `postMessage`?

twolfson:
It should be something like adding a `postMessage` call in one of our test suites and verifying nothing is thrown but as mentioned in #2239, I ran out of time to explore that =/

AnilRedshift:
Is there an ETA for this fix - it is causing our builds on Safari to fail ~50% of the time. The fix proposed seems reasonable to me.

al-the-x:
@AnilRedshift I can confirm (along with others from #2239) that rolling back to 1.0.0 fixes the issue for immediate purposes.

dignifiedquire:
release will happen next week

dignifiedquire:
Thanks :octocat:

Issue #2194, opened 2016-06-26
karma v1.0.0: You need to include some adapter that implements __karma__.start method!


SerkanSipahi:
After upgrading our project from karma v0.13.22 to v.1.0.0 we get following error: 
`You need to include some adapter that implements __karma__.start method!` 

Here is a the repo: https://github.com/SerkanSipahi/app-decorators

##### Reproducing error:
```
git clone https://github.com/SerkanSipahi/app-decorators.git
cd app-decorators
// change in package.json karma to 1.0.0
make install
make compile
make test
```

Anyone have an idea What causes this error ?

dignifiedquire:
Thanks for the report, I think the problem is somewhere with karma-jspm. When I comment it out the error goes away. Might be some internals that changed that karma-jspm relied on.

SerkanSipahi:
@dignifiedquire thank you the quickl response. I will relate this issue to karma-jspm.

hpurmann:
Unfortunately, it also breaks `karma-sinon-chai` https://github.com/kmees/karma-sinon-chai/issues/38

RIAstar:
The problem indeed seems to be with karma-jspm. I tracked it down to some duplicate file references in the globs on the `files` array.
See https://github.com/Workiva/karma-jspm/issues/162#issuecomment-229071459 for a workaround.

That said, the result of this duplicate file reference is that no files are loaded at all, including `adapter.js`, hence the message that the adapter has no `start` function, since there is no adapter to begin with.
So perhaps this could be a bug on the Karma side as well?
In any case, the fact that the file loading fails silently is not very nice.

jgerigmeyer:
Not surprisingly, it breaks `karma-chai-sinon` (https://github.com/tubalmartin/karma-chai-sinon) as well.

RIAstar:
@jgerigmeyer : that's odd. I use `karma-chai-sinon` as well and don't have any issues with that (only with `karma-jspm`).

Versions I use: 
- "karma-chai-sinon": "^0.1.5"
- "karma-mocha": "^1.1.1"

SerkanSipahi:
@jgerigmeyer do you use jspm in your project? if yes, can you share your setup?

jgerigmeyer:
I am not using jspm. With this setup:

```json
    "karma": "1.1.0",
    "karma-chai-sinon": "0.1.5"
```

I get the following:

```bash
[...]

PR #2255, opened 2016-07-13
fix(file-list): Ensure duplicate files are handled properly

m-a-r-c-e-l-i-n-o:
Previously, if file patterns amounted to duplicate files, the pattern resolution function would fail and Karma would continue on with an empty files list, namely "included". This led to confusing, generic adapter errors. This commit fixes that by replacing the offending code with a (likely) equivalent. 
To replicate issue, include patterns that amount to duplicate files.
To see the error, wrap the following statement `if (other && other.compare(p) < 0) return` (contained inside `file-list.js`) in a `try {} catch () {}` block. In essence, `other.compare` does not exist.

Might be related to issue #2194

### DIFF
--- a/lib/file-list.js
+++ b/lib/file-list.js
@@ -233,7 +233,7 @@ Object.defineProperty(List.prototype, 'files', {
       var bucket = expandPattern(p)
       bucket.forEach(function (file) {
         var other = lookup[file.path]
-        if (other && other.compare(p) < 0) return
+        if (other && byPath(other, p) < 0) return
         lookup[file.path] = p
         if (p.included) {
           included[file.path] = file


m-a-r-c-e-l-i-n-o:
@dignifiedquire Sorry to pull you into this, I saw that you were involved in what could be a related [issue](https://github.com/karma-runner/karma/issues/2194). Would you happen to know what the intention of [`if (other && other.compare(p) < 0) return`](https://github.com/karma-runner/karma/blob/master/lib/file-list.js#L236) was suppose to be? As mentioned above, `other.compare(p) ` does not appear to exist in the [`file`](https://github.com/karma-runner/karma/blob/master/lib/file.js) object. I replaced it with `byPath(other, p)`, but I'm not entirely sure that it's the way to go, especially since a few related e2e tests aren't passing.

dignifiedquire:
Thank you for figuring this out, the commit introducing the commit was https://github.com/karma-runner/karma/commit/db42a7fb1d37fa1196759b4f6b12e39f612fae5c and the `.compare` comes from here: https://github.com/karma-runner/karma/blob/master/lib/config.js#L42 

Probably the issue is that `other` is not an instance of `Pattern` and that is the issue that needs fixing.

m-a-r-c-e-l-i-n-o:
@dignifiedquire Thank you for the additional info, does normalizing the patterns before they are included in the lookup sound like a decent strategy to you? Something along the lines of:
```
if (p.constructor.name !== 'Pattern') {
   p = new Pattern(p.pattern, p.served, p.included, p.watched, p.nocache)
}
lookup[file.path] = p
```

dignifiedquire:
I think we should make sure that [here](https://github.com/m-a-r-c-e-l-i-n-o/karma/blob/5c682a373c202508be9c0787cae1b85725efb899/lib/file-list.js#L53) and [here](https://github.com/m-a-r-c-e-l-i-n-o/karma/blob/5c682a373c202508be9c0787cae1b85725efb899/lib/file-list.js#L271) only pattern instances are passed in, so that we are sure that `this._pattern` is always and array of `Pattern`

m-a-r-c-e-l-i-n-o:
@dignifiedquire I actually considered that, too. But, I don't fully understand the data flow in the system and thought that it would be risky to try to address the roots. However, if you're suggesting it, it makes me feel a bit more confident about that route — it seems to be more ideal. I thought about doing something along the lines of:
```
this._patterns = patterns.map(function (p) {
  if (p.constructor.name === 'Pattern') {
    return p
  }
  return new Pattern(p.pattern, p.served, p.included, p.watched, p.nocache)
})
```
Is that what you had in mind?

dignifiedquire:
Sorry @m-a-r-c-e-l-i-n-o for not getting back on this earlier, I was away on a work trip and didn't have much time to look into this. It's important to get fixed properly so I will see that I can get you good answer by the end of today.

dignifiedquire:
After reading your comment, yes I think this is the right way to go. Have you tested if the issues in #2194 are going away after a fix like this?

m-a-r-c-e-l-i-n-o:
@dignifiedquire No worries. I just tested it and found two odd behaviors.

1. The [patterns object](https://github.com/m-a-r-c-e-l-i-n-o/karma/blob/5c682a373c202508be9c0787cae1b85725efb899/lib/file-list.js#L53) will throw a "no adapter" error during unit tests if the object is replaced, so `map` is not usable here. Not a big deal, but would be nice to know why that is. You can try doing `this._patterns = patterns.concat([])` to experience what I am talking about.

2. [This](https://github.com/m-a-r-c-e-l-i-n-o/karma/blob/5c682a373c202508be9c0787cae1b85725efb899/lib/file-list.js#L53) and [this](https://github.com/m-a-r-c-e-l-i-n-o/karma/blob/5c682a373c202508be9c0787cae1b85725efb899/lib/file-list.js#L271) do not appear to be roots of [this](https://github.com/m-a-r-c-e-l-i-n-o/karma/blob/5c682a373c202508be9c0787cae1b85725efb899/lib/file-list.js#L232). In my setup, the patterns object [here](https://github.com/m-a-r-c-e-l-i-n-o/karma/blob/5c682a373c202508be9c0787cae1b85725efb899/lib/file-list.js#L51) is empty and [here](https://github.com/m-a-r-c-e-l-i-n-o/karma/blob/5c682a373c202508be9c0787cae1b85725efb899/lib/file-list.js#L232) its full, so the error is still present since the patterns haven't been normalized. I'm assuming it's being populated by the plugins?

We need to understand this data flow a little better. Any thoughts?

dignifiedquire:
Thanks a lot, turns out I have to fix it in that nasty location as the adapters might be editing things directly. See my PR for a fix, which I will merge and release as soon as CI is happ.

PR #2267, opened 2016-07-20
Call .resume() on process streams so they don't get blocked

voltrevo:
If you manage to make chrome emit errors in the console, like we do with webrtc, the node stream buffer for stderr fills up and locks up the browser. This just calls `.resume()` to prevent that.

Here's a tarball of a simplified project that repro's the issue:
https://drive.google.com/file/d/0By71SGpctHHxVzl1bkx6bGg3b2c/view?usp=sharing

Simply run `karma start --browsers ChromeFakeDevices` and chrome should lock up when it's done around 200 of the 1000 identical tests.

### DIFF
--- a/lib/launchers/process.js
+++ b/lib/launchers/process.js
@@ -135,7 +135,17 @@ var ProcessLauncher = function (spawn, tempDir, timer) {
 
 ProcessLauncher.decoratorFactory = function (timer) {
   return function (launcher) {
-    ProcessLauncher.call(launcher, require('child_process').spawn, require('../temp_dir'), timer)
+    var spawn = require('child_process').spawn
+
+    var spawnWithoutOutput = function () {
+      var proc = spawn.apply(null, arguments)
+      proc.stdout.resume()
+      proc.stderr.resume()
+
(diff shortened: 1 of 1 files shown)


voltrevo:
I signed it!

package.json
```diff
@@ -284,7 +284,8 @@
     "Jon Bretman <jon.bretman@gmail.com>",
     "Jonathan ES Lin <ernsheong@gmail.com>",
     "Julian Connor <julian.connor@venmo.com>",
-    "Jurko Gospodnetić <jurko.gospodnetic@pke.hr>"
+    "Jurko Gospodnetić <jurko.gospodnetic@pke.hr>",
+    "Andrew Morris <voltrevo@gmail.com>"
```
dignifiedquire: Can you remove this please, these get autogenerated on release.

voltrevo: @dignifiedquire fixed

voltrevo:
@dignifiedquire fixed

dignifiedquire:
Thanks :octocat:

Issue #2237, opened 2016-07-07
How to debug the "You need to include some adapter that implements __karma__.start method!" error?


C3-TKO:
### Expected behavior
I expected that karma is still working after update from 0.13.9 to 1.1.0. If karma is not working properly anymore i would like to have some more information how to fix this problem myself

### Actual behavior
07 07 2016 08:36:13.016:INFO [karma]: Karma v1.1.0 server started at http://localhost:8080/
07 07 2016 08:36:13.019:INFO [launcher]: Launching browser PhantomJS with unlimited concurrency
07 07 2016 08:36:13.087:INFO [launcher]: Starting browser PhantomJS
07 07 2016 08:36:14.570:INFO [PhantomJS 2.1.1 (Mac OS X 0.0.0)]: Connected on socket /#kR-iedk86tRVJxHqAAAA with id 89582790
PhantomJS 2.1.1 (Mac OS X 0.0.0) ERROR
  You need to include some adapter that implements __karma__.start method!

### Enviroment Details
https://github.com/C3-TKO/junkan/blob/update-karma-to-1.0.0/karma.conf.js
https://github.com/C3-TKO/junkan/blob/update-karma-to-1.0.0/package.json

### Steps to reproduce the behaviour

1. Checkout branch https://github.com/C3-TKO/junkan/tree/update-karma-to-1.0.0
2. npm install
3. npm run test

I would be glad if you could tell me hoy debug this issue myself. I already checked other issues with the same error and most of the times this is an error regarding a dependency but i failed at recognizing how to debug this myself and how to find out where the adapter fails. It should be the one of karma-mocha.

dignifiedquire:
This is related due to some plugins not being updated and ready for karma 1.0. See #2194 for details and workarounds.

PR #2277, opened 2016-07-25
fix(file-list): ensure patterns are comparable

### LINKED ISSUE #2194: karma v1.0.0: You need to include some adapter that implements __karma__.start method!

dignifiedquire:
Thanks to @m-a-r-c-e-l-i-n-o for pointing me in the right direction
Fixes #2194

This supersedes #2255 

Also fixes 
- https://github.com/karma-runner/karma-jasmine/issues/127
- https://github.com/karma-runner/karma-mocha/issues/102
- https://github.com/karma-runner/karma/issues/2237

### DIFF
--- a/lib/file-list.js
+++ b/lib/file-list.js
@@ -20,6 +20,7 @@ var File = require('./file')
 var Url = require('./url')
 var helper = require('./helper')
 var log = require('./logger').create('watcher')
+var createPatternObject = require('./config').createPatternObject
 
 // Constants
 // ---------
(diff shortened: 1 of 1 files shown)


lib/file-list.js
```diff
@@ -230,6 +231,13 @@ Object.defineProperty(List.prototype, 'files', {
     var lookup = {}
     var included = {}
     this._patterns.forEach(function (p) {
+      // This needs to be here sadly, as plugins are modifiying
+      // the _patterns directly resulting in elements not being
```
tamird: @dignifiedquire for my own edification, can you give an example of a plugin doing this? Specifically, is [karm-jspm](https://github.com/Workiva/karma-jspm) doing this?

dignifiedquire:
yes this is one place https://github.com/Workiva/karma-jspm/blob/master/src/init.js#L106

tamird:
Ah, understood. LGTM.

PR #2288, opened 2016-07-29
docs: include jasmine-core in installation instructions

camargo:
Update to include jasmine-core in install doc, which is a peer dependency of karma-jasmine


dignifiedquire:
Thanks :octocat:

Issue #1751, opened 2015-12-07
Missing error handler on `socket`.


binarykitchen:
I am seeing this in the current test set up and would like to catch this kind of error from the karma server instance. Is that possible?

csmalin:
I'm only seeing this problem when using karma-jenkins-reporter or karma-junit-reporter, are you using any of these plugins?

binarykitchen:
no

priley86:
+1 seeing this with karma-junit-reporter in my tests with Node 4.23 LTS...any ideas?

```
    "karma": "~0.13.15",
    "karma-chrome-launcher": "~0.2.0",
    "karma-coverage": "~0.4.2",
    "karma-firefox-launcher": "~0.1.6",
    "karma-jasmine": "~0.3.6",
    "karma-junit-reporter": "^0.2.2",
    "karma-ng-html2js-preprocessor": "^0.1.2",
    "karma-phantomjs-launcher": "~0.2.0",
    "socket.io": "^1.3.7",
```
PhantomJS 1.9.8 (Mac OS X 0.0.0): Executed 48 of 277 (26 FAILED) ERROR (0.732 secs / 0.769 secs)
Missing error handler on `socket`.
TypeError: Cannot read property '68890579' of null
    at specSuccess.specSkipped.specFailure (/Users/patrick/Projects/community/node_modules/karma-junit-reporter/index.js:89:22)
    at onSpecComplete (/Users/patrick/Projects/community/node_modules/karma/lib/reporters/base.js:78:12)
    at null.<anonymous> (/Users/patrick/Projects/community/node_modules/karma/lib/events.js:13:22)
    at emitTwo (events.js:92:20)
    at emit (events.js:172:7)
    at onResult (/Users/patrick/Projects/community/node_modules/karma/lib/browser.js:216:13)
    at Socket.<anonymous> (/Users/patrick/Projects/community/node_modules/karma/lib/events.js:13:22)
    at emitOne (events.js:82:20)
    at Socket.emit (events.js:169:7)
    at Socket.onevent (/Users/patrick/Projects/community/node_modules/socket.io/lib/socket.js:330:8)
    at Socket.onpacket (/Users/patrick/Projects/community/node_modules/socket.io/lib/socket.js:290:12)
    at Client.ondecoded (/Users/patrick/Projects/community/node_modules/socket.io/lib/client.js:193:14)
[...]

Issue #61, opened 2012-09-28
Ability to override config properties via command line arguments


guylabs:
Hi,

we have a CI environment where we run multiple projects and each one has e webapp running for the tests. Now we have some kind of logic which searches a free port to start the server and tell the application on which port the server is running. Now i just found out that I could override the port setting in the config by starting testacular with this command line argument:

--port=1234

But I can't create objects in the command line to use for example for the proxies property. Is there a possibility to add them via command line arguments or system properties ?

Thanks and regards,

Guy

PR #2297, opened 2016-08-03
fix(web-server): Allow karma to run in project which path contains HTML URL encoded characters

Dam-ien:
Karma fails on Jenkins (standard "Multibranch pipeline" project) when it checks out branches containing '/' as it replaces '/' by '%2F' in the folder name.

Alternatively, if the project parent folder contains %2F in its name, karma will not execute properly.

Fixes errors seen on #1751, #61.


dignifiedquire:
Thank you :)

PR #2316, opened 2016-08-16
fix(launcher): Extra concurrent browser on launcher restart

### LINKED ISSUE #2280: The concurrency limit is passed when a timeout occurs and the browser restarts
### Expected behavior
With a concurrency limit specified the concurrently run browsers should be limited to that value.

### Actual behavior
When a browser passes one of the timeouts specified in the configuration below, that browser restarts. The next browser is also allowed to start, setting the amount of concurrent browsers to 2. Even though the concurrency limit is set to 1.

### Environment Details

Karma 1.1.1

``` JavaScript
config.set({
  ...
  concurrency: 1,
  captureTimeout: 60000,
  browserDisconnectTimeout: 60000,
  browserDisconnectTolerance: 3,
  browserNoActivityTimeout: 60000
});
```

1. Use the above configuration
2. Let a browser timeout
3. Check the amount of active browsers

SteveVanOpstal:
Fixes #2280

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -100,7 +100,7 @@ var Launcher = function (server, emitter, injector) {
         browser.on('done', function () {
           // We are not done if there was an error as first the retry takes
           // place which we catch with `browser_process_failure` if it fails
-          if (browser.error) return
+          if (browser.error || browser.state === browser.STATE_RESTARTING) return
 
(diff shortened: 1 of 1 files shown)


dignifiedquire:
Thanks :octocat:

Issue #2249, opened 2016-07-12
External scripts without the necessary `Access-Control-Allow-Origin` header are rejected in karma v1.1.1


dotch:
### Expected behavior

External scripts can be loaded regardless of the `Access-Control-Allow-Origin` headers.
### Actual behavior

External scripts without the necessary `Access-Control-Allow-Origin` header are rejected with:

```
Script from origin '...' has been blocked from loading by 
Cross-Origin Resource Sharing policy: 
No 'Access-Control-Allow-Origin' header is present on the 
requested resource. Origin '...' is therefore not allowed access.
```

This happens with Karma v1.1.1 when external scripts are served without the necessary `Access-Control-Allow-Origin` header. 
In Karma v1.1.0 the scripts were loaded. This change in behavior is due to https://github.com/karma-runner/karma/commit/5690ffe82bb7ff195576f6f43dd8c281e51aa833

Sometimes users might not have control over the external scripts they need to include in their test run. An option to turn off the new default behavior might be sensible.
### Enviroment Details
- Karma version (output of `karma --version`): v1.1.1
- A modern browser that respects `crossorigin="anonymous"`

frnk94:
Since: https://github.com/karma-runner/karma/pull/2219

maksimr:
Already fixed in PR #2318

PR #2318, opened 2016-08-16
fix(runner): Fix loading of external scripts without cors headers

dotch:
As described in #2249, external scripts without the necessary `Access-Control-Allow-Origin` header are rejected with:

```
Script from origin '...' has been blocked from loading by 
Cross-Origin Resource Sharing policy: 
No 'Access-Control-Allow-Origin' header is present on the 
requested resource. Origin '...' is therefore not allowed access.
```

This PR reverts the change that introduced this behavior. 

If we want to keep the benefits that were introduced with #2219 we could create a configuration option for this. 
In that case we'd have to decide what we want as a default behavior.


dignifiedquire:
Let's keep the current behaviour as the default one, and allow to disable it via configuration.

dotch:
I added a `crossOriginAttribute` configuration option. (open to better naming)

dignifiedquire:
Thanks :octocat: I'm afraid I don't have any better name ideas either.

Issue #1788, opened 2016-01-06
Karma 0.13.19 taking long time to complete when run via gulp


mikelax:
I am opening a new ticket as this seems to be fallout from #1782 and the socket.io upgrade to 1.4.0.

I've just noticed a side effect from upgrading to 0.13.19; gulp karma test tasks take much longer to exit after the test run.

```javascript
gulp.task('test:ng', function(done) {
  new Server({
    configFile: __dirname + '/karma.conf.js',
    singleRun: true
  }, done).start();
});
```
Previously the above task used to complete immediately after tests have completed. on 0.13.9, there is a pause of ~30 seconds after the karma runner completes the tests and the task exits.

```javascript
gulp.task('test:ng', function(done) {
  new Server({
    configFile: __dirname + '/karma.conf.js',
    singleRun: true
  }, function() {
    console.log('done');
    done();
  }).start();
});
```
The above outputs done to the console but the gulp task still doesn't complete until after ~30 seconds.

cc @nazar

dsine-de:
I'm currently implementing karma with gulp for the first time and can confirm this.
I was thinking karma wasn't exiting at all, but after reading this issue I waited ~30 secs and it exits correctly (but it takes very long).

lukaselmer:
+1

It's actually version 0.13.19 which causes the problem (the hotfix for socket.io). See https://github.com/karma-runner/karma/commit/3ab78d6

lukaselmer:
Until this issue is resolved, here's a simple workaround: Patch your package.json:

```json
"devDependencies": {
    ...
    "karma": "=0.13.18",
    "socket.io": "1.3.7",
   ...
  },
```

Dignifiedquire:
This might also be caused by an internal change in socket.io@1.4.0

redboul:
Hi,
I'm encountering the same thing on a project with gulp but using karma CLI ends the same way.
Thx @lukaselmer for the workaround

austenLacy:
ditto here

hamishtaplin:
Also having this problem.

aragonezr:
Same problem here. [Mac (node 5.4.1, npm 3.3.12, gulp 3.9.0, karma 0.13.19)]
The patch mentioned by @lukaselmer works for me.

mikezhuyuan:
Same issue. It works after changed to sockect.io 1.3.7 as mentioned by @lukaselmer. Please fix.

BeneStem:
[...]

PR #2395, opened 2016-10-07
#1788 fix karma exiting process only after a long delay, when using i…

johannesjo:
…t via gulp. This should fix the issue, though I'm not 100% sure there are any drawbacks.


johannesjo:
I signed it!

johannesjo:
It seems like mike PR only kills the process as a whole, which is no improvement. I'm closing this.

PR #2404, opened 2016-10-13
Fix plugin loading bug #2221

### LINKED ISSUE #2221: could not load plugins while karma not installed in node_modules/karma
### Expected behavior

loading plugins even if karma installed in sub directory in node_modules
### Actual behavior

can not load plugins
### Enviroment Details
- Karma version 1.1.0 and 0.0.13 (output of `karma --version`):

karma installed by a npm installing tool, located in `node_modules/.npminstall/karma`, the same as all other karma plugins.

zincli:
sorry for PR again, I made a mistake deleted the previous branch. This fixes #2221 .

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -30,7 +30,7 @@ exports.resolve = function (plugins, emitter) {
         requirePlugin(plugin)
         return
       }
-      var pluginDirectory = path.normalize(path.join(__dirname, '/../..'))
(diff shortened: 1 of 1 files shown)


dignifiedquire:
Thank you. I agree there is a bug to be fixed here, but I am not sure this will work, given that `process.cwd()` could be another place, i.e. when being run with a global installation of `karma-cli`. Have you tested that use case?

PR #2445, opened 2016-11-10
feat(web-server): allow overriding of default http module

### LINKED ISSUE #2424: Can't use http2 with Karma
### Expected behaviour

Uses http 2
### Actual behaviour

Uses http 1.1

Is there any way to use http2 in with Karma server?

I'm running tests from gulp using:

```
  new karma.Server({ ... });
```

and my site is heavily utilizing http2's concurrent connections (no bundling). It's working great in production and we're living the dream, but the tests are running horribly slow because the karma server doesn't have http2. Is there anyway to either turn on http2... or provide node's http2 server or possibly even just tell karma to use another http server that does support http2?

itslenny:
allow setting an external module for use by KarmaServer

Closes #2424

### DIFF
--- a/docs/config/01-configuration-file.md
+++ b/docs/config/01-configuration-file.md
@@ -537,6 +537,17 @@ Determines the use of the Node `http` or `https` class.
 Note: Using `'https:'` requires you to specify `httpsServerOptions`.
 
 
+## httpModule
+**Type:** String
+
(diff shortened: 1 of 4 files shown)


dignifiedquire:
Cool, that was easier than I expected. Thank you :octocat:

PR #2471, opened 2016-11-28
feat(index): Expose config on the index file

nicojs:
I would like to be able to only use the `config` part of karma. Using `require('karma/lib/config')` feels like i'm trashing a private party, i would rather be invited  :cake:.

### DIFF
--- a/docs/dev/04-public-api.md
+++ b/docs/dev/04-public-api.md
@@ -134,6 +134,20 @@ stopper.stop({port: 9876}, function(exitCode) {
 })
 ```
 
+## karma.config.parseConfig([configFilePath], [cliOptions])
+
+This function will load given config file and return a filled config object. 
+This can be useful if you want to integrate karma into another tool and want to load
+the karma config while honoring the karma defaults. For example, the [stryker-karma-runner](https://github.com/stryker-mutator/stryker-karma-runner)
+uses this to load your karma configuration and use that as stryker configuration.
+
+```javascript
+const cfg = require('karma').config;
+const path = require('path');
+// Read karma.conf.js, but override port with 1337
+const karmaConfig = cfg.parseConfig(path.resolve('./karma.conf.js'), { port: 1337 } ); 
+```
+
 ## Callback function notes
 
 - If there is an error, the error code will be provided as the second parameter to the error callback.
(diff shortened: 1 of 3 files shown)


dignifiedquire:
This exposes the whole config module, rather than the actual config being used/loaded. I'm curious, what do you want to do with it?

test/unit/index.spec.js
```diff
@@ -0,0 +1,9 @@
+var cfg = require('../../lib/config')
+
+describe('index', () => {
+  var index = require('../../lib/index')
+
+  it('should expose the `config` object', () => {
+    expect(index.config).toBe(cfg)
```
dignifiedquire: Needs to be`.to.be.eql(` as we use chai not jasmine :) (that's why the tests are failing)

nicojs:
> This exposes the whole config module, rather than the actual config being used/loaded. I'm curious, what do you want to do with it?

I want to use it in [stryker-karma-runner](https://github.com/stryker-mutator/stryker-karma-runner) to configure stryker with karma config.

We want to support a config setting `karmaConfigFile: 'karma.conf.js'` in the `stryker.conf.js`. After loading the `karma.conf.js` file using `require('karma').config.parseConfig('./karma.conf.js')` we would be able to set stryker's `files` property for example.

dignifiedquire:
Interesting. Happy to merge this, but could you add some docs that this is exposed now? Also I think it would better to only expose the parseConfig` as this reduces the api surface we have to gurantee stability on.

test/unit/index.spec.js
nicojs: Thanks! :cold_sweat:

I couldn't run the tests locally because of out-dated dependencies and had now way to update them in the train. I will update this PR

docs/dev/04-public-api.md
```diff
@@ -134,6 +134,20 @@ stopper.stop({port: 9876}, function(exitCode) {
 })
 ```
 
+## karma.config.parseConfig([configFilePath], [cliOptions])
+
+This function will load given config file and returns a filled config object. 
+This can be useful if you want to integrate karma into another tool and want to load
+the karma config while honoring the karma defaults. For example, the [stryker-karma-runner](https://github.com/stryker-mutator/stryker-karma-runner)
```
nicojs: I've added some surreptitious advertising, i hope that's ok :grin:

nicojs:
Thank you!

PR #2478, opened 2016-12-02
fix(server): No error is returned when using a browser that is not in…

### LINKED ISSUE #2403: No error is returned when using a browser that is not installed.
### Expected behaviour

Given that the target browser is `PhantomJS`
and the plugin package for `PhantomJS` is not installed.
When the following code runs:

``` js
  new karma.Server({
    configFile: __dirname + '/karma.conf.js',
    singleRun: true
  }, function(status) {
    console.log(status);
  }).start();
```

Then the console should print an error and an error status != 0 like this:

``` console
12 10 2016 17:25:43.747:ERROR [launcher]: Cannot load browser "PhantomJS": it is not registered! Perhaps you are missing some plugin?
12 10 2016 17:25:43.748:ERROR [karma]: Found 1 load error
{something that is not 0}
```
### Actual behaviour

The console shows a status 0:

``` console
12 10 2016 17:25:43.747:ERROR [launcher]: Cannot load browser "PhantomJS": it is not registered! Perhaps you are missing some plugin?
12 10 2016 17:25:43.748:ERROR [karma]: Found 1 load error
0
```
### Environment Details

``` js
{
    browsers: ['PhantomJS'],
}
```

See above.
### Additional notes

This breaks the recommended usage shown in [gulp-karma](https://github.com/karma-runner/gulp-karma) as the job will exit with status 0 and pass CI checks although no tests actually ran!

maksimr:
…stalled

close #2403

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -184,6 +184,7 @@ Server.prototype._start = function (config, launcher, preprocess, fileList,
       var noLoadErrors = self.loadErrors.length
       if (noLoadErrors > 0) {
         self.log.error('Found %d load error%s', noLoadErrors, noLoadErrors === 1 ? '' : 's')
+        process.exitCode = 1
         process.kill(process.pid, 'SIGINT')
       }
     })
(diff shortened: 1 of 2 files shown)


dignifiedquire:
Thanks :octocat:

Issue #2477, opened 2016-12-01
BindAddress option


grifball:
I'm sorry if this option already exists and I just didn't notice it.
I'm using karma on a private project on a server with port 9876 open.
Rather than going through and closing that port in the firewall, it's easier just to bind to 127.0.0.1 instead of 0.0.0.0.
So I added support into karma for that, it's a simple change, I included a diff of the files.
I'm using karma version 1.3.0
[diff new old.txt](https://github.com/karma-runner/karma/files/623373/diff.new.old.txt)

Let me know what you guys think of having this option.

maksimr:
@grifball Karma don't have this option now
We had a [pull request(PR)](https://github.com/karma-runner/karma/pull/1214) about it with long discussion.
It will be nice if you look at this PR and reanimate it

Thanks!

relates https://github.com/karma-runner/karma/issues/1593

grifball:
I looked at #1311 and #1214 and they both seem fine to solve this issue, so I added them in and fixed the unit tests so that the dummy web server accepts the correct arguments. I'll make a PR.

PR #2479, opened 2016-12-03
Adding listen address option so that IPv6 and loopback interfaces can be used.

grifball:
reference: #2477

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -104,7 +104,7 @@ var normalizeProxyPath = function (proxyPath) {
   return normalizedProxyPath
 }
 
-var normalizeConfig = function (config, configFilePath, configModule, cliOptions) {
+var normalizeConfig = function (config, configFilePath) {
   var basePathResolve = function (relativePath) {
     if (helper.isUrlAbsolute(relativePath)) {
       return relativePath
@@ -197,21 +197,6 @@ var normalizeConfig = function (config, configFilePath, configModule, cliOptions
 
   if (config.formatError && !helper.isFunction(config.formatError)) {
     throw new TypeError('Invalid configuration: formatError option must be a function.')
-  }
-
-  // if the user changed listenAddress, but didn't set a hostname, warn them
-  var testConfig = new Config()
-  testConfig.hostname = null
-  testConfig.listenAddress = null
-  if (configModule != null) {
-    configModule(testConfig)
-  }
-  if (cliOptions != null) {
-    testConfig.set(cliOptions)
-  }
-  if (testConfig.hostname == null && testConfig.listenAddress != null) {
-    log.warn('ListenAddress is set but hostname isn\'t. If your browsers fail ' +
-      'to connect, consider setting the hostname option.')
   }
 
   var defaultClient = config.defaultClient || {}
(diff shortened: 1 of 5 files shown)


grifball:
I signed the cla. Not exactly sure how this works. Is that all I have to do?

maksimr:
LGTM

maksimr:
@grifball could you squash commits in to one

Thanks

grifball:
I started writing something that automatically tries to set hostname when listenAddress is set but hostname isn't, but it seems out of scope. Figuring out which IP to use for hostname based on the value of listenAddress is really difficult, and can even be impossible in more complex interface setups. Maybe this isn't what you had in mind when you said "influence?"
I think that a better solution would be to give a warning if listenAddress is set and hostname isn't.

grifball:
Let me know what you think of this solution. I'm not sure how to change the label on this PR.
I'll squash the commits once the code looks good.

lib/config.js
```diff
@@ -199,6 +199,21 @@ var normalizeConfig = function (config, configFilePath) {
...
+  testConfig.hostname = null
+  testConfig.listenAddress = null
+  if (configModule != null) {
+    configModule(testConfig)
+  }
+  if (cliOptions != null) {
+    testConfig.set(cliOptions)
+  }
+  if (testConfig.hostname == null && testConfig.listenAddress != null) {
+    log.warn('ListenAddress is set but hostname isn\'t. If your browsers fail ' +
```
dignifiedquire: It would be nice to print the current values of both the variables here.

lib/config.js
```diff
@@ -199,6 +199,21 @@ var normalizeConfig = function (config, configFilePath) {
     throw new TypeError('Invalid configuration: formatError option must be a function.')
   }
 
+  // if the user changed listenAddress, but didn't set a hostname, warn them
+  var testConfig = new Config()
+  testConfig.hostname = null
+  testConfig.listenAddress = null
+  if (configModule != null) {
+    configModule(testConfig)
```
dignifiedquire: Let's not do this, as this means we are evaluating the code from the user twice. Best case this slows things down, worst case it has unexpected side effects. You can add the check probably around the original invocition of the `configModule`

dignifiedquire:
Thank you, yes I think it's fine to just warn the user about this. Left some comments about the implementation.

grifball:
Alright, I implemented those changes. Let me know if anything else needs changing.

wesleycho:
Do we want weak equality (`==` & `!=`) checks here? Seems like a better way would be to check specifically if it is a non-trivial string.

grifball:
This should be a safe enough check. Unless they're setting these values to null and expecting the server to work correctly, which explicit string checks wouldn't solve anyway.
If they're setting these values to non-string types, I assume there's other checks in the config file for this, but it's out of scope in this PR anyway. The added code won't change how we handle config type checking.

grifball:
I added a simple test to make sure that listen address was called with the configured listenAddress. That should cover all the functionality for this PR.

dignifiedquire:
Thank you :octocat:

Issue #2361, opened 2016-09-16
Feature Request: Export all the constants


marthinus-engelbrecht:
### Expected behaviour

All the constants would be exported. 
### Actual behaviour

Only the VERSION constant is exported in the [index.html](https://github.com/karma-runner/karma/blob/master/lib/index.js#L22)
### Environment Details
- Karma version 1.3.0
### Reasoning

We use gulp and we want to specify the log level via the Server constructor but it would be more reliable to use the constants defined inside karma than magic string like 'DEBUG'

marthinus-engelbrecht:
This should be a trivial change. I'm more than willing to do a pull request

dignifiedquire:
Sure :) PR welcome.

PR #2480, opened 2016-12-04
feat(constants): export constants

wesleycho:
- Export constants for public usage

This should implement #2361.

### DIFF
--- a/docs/dev/04-public-api.md
+++ b/docs/dev/04-public-api.md
@@ -148,6 +148,56 @@ const path = require('path');
 const karmaConfig = cfg.parseConfig(path.resolve('./karma.conf.js'), { port: 1337 } ); 
 ```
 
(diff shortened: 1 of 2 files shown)


maksimr:
LGTM

dignifiedquire:
Thank you, can you please add documentation for this?

wesleycho:
Where would the best place to add the documentation on the exported constants be? I could add information about some of it to the configuration file docs, but it seems out of place for some of the values such as `DEFAULT_PORT` and `DEFAULT_HOSTNAME`.

dignifiedquire:
https://github.com/karma-runner/karma/blob/master/docs/dev/04-public-api.md < public api is the place to put it I would say

wesleycho:
Added docs there for all of the constants

PR #2484, opened 2016-12-06
fix(middleware): update Buffer usage

### LINKED ISSUE #2429: Deprecation warnining in Node 7
### Expected behavior
Things to work! (i.e. no deprecation warnings crashing Karma)

### Actual behavior

[This deprecation warning:](https://ci.appveyor.com/project/isiahmeadows/thallium/build/93/job/9trd5oiry82wed4u#L615)

```
node.exe : (node:2636) DeprecationWarning: Using Buffer without `new` will soon stop working. Use `new Buffer()`, or preferably `Buffer.from()`, `Buffer.allocUnsafe()` or `Buffer.alloc()` instead.
```

### Environment Details

- Relevant part of your `karma.config.js` file: N/A (I think - [here's my config](https://github.com/isiahmeadows/thallium/blob/master/karma.conf.js))
- OS: Windows only

### Steps to reproduce the behavior

1. Run Karma in Node 7
2. Watch the output

-----

This came up in CI, and has been failing my Windows builds (oddly, it's not affecting my Linux builds). Downgrading Node did fix the problem, though.

wesleycho:
- Update Buffer usage to non-deprecated methods of creating a Buffer

Note that this does not fully fix #2429, as there is a plugin that needs to be looked into. However, it should be noted that these methods were only added in Node 5.10.0, which means that this is a breaking change in order to satisfy Node 7. This should probably be bundled with a major version bump when this is chosen to be merged.

### DIFF
--- a/lib/middleware/common.js
+++ b/lib/middleware/common.js
@@ -5,6 +5,7 @@
 var mime = require('mime')
 var _ = require('lodash')
 var parseRange = require('range-parser')
+var Buffer = require('safe-buffer').Buffer
 
 var log = require('../logger').create('web-server')
 
@@ -40,10 +41,10 @@ var createServeFile = function (fs, directory, config) {
         return 200
       } else if (range === -1) {
         // unsatisfiable range
-        responseData = new Buffer(0)
+        responseData = Buffer.alloc(0)
         return 416
       } else if (range.type === 'bytes') {
-        responseData = new Buffer(responseData)
+        responseData = Buffer.from(responseData)
         if (range.length === 1) {
           var start = range[0].start
           var end = range[0].end
--- a/package.json
+++ b/package.json
@@ -320,6 +320,7 @@
     "qjobs": "^1.1.4",
     "range-parser": "^1.2.0",
     "rimraf": "^2.3.3",
+    "safe-buffer": "^5.0.1",
     "socket.io": "1.7.1",
     "source-map": "^0.5.3",
     "tmp": "0.0.28",


maksimr:
@wesleycho It does not work in [**node@0.12**](https://nodejs.org/docs/latest-v0.12.x/api/buffer.html) - https://ci.appveyor.com/project/dignifiedquire/karma/build/939/job/irjlq04x92xmt295.

We should create fallback for node@0.12 and think about should we deprecate node@0.12  or not

Thanks

cc @dignifiedquire

dignifiedquire:
As @maksimr said this won't work in Node 0.12 and below. Given the LTS schedule I already have a plan of moving karma to require Node >= 4, but even then we can not upgrade to use `Buffer.from` as it is only available in Node >= 5. But we we could use [safe-buffer](https://www.npmjs.com/package/safe-buffer) to polyfill this feature today :)

wesleycho:
Updated to use `safe-buffer`

maksimr:
@wesleycho could you rebase on master

wesleycho:
Done

dignifiedquire:
Thank you :octocat:

isiahmeadows:
Actually, my issue is ATM invalid as far as I can tell. My current failures are from highly uninformative PhantomJS stack traces (one line isn't enough context), but here's not the correct place to complain about that.

dignifiedquire:
Thanks @isiahmeadows for the update, but updating the APIs we use was still a good idea 😄

Issue #2190, opened 2016-06-24
Karma auto install socket.io 1.4.8 it makes ie9 can not run karma test.


chenbojian:
### Expected behavior

can run karma test in IE9
### Actual behavior

can not run karma test in IE9. only if I lock socket.io version to 1.4.6, it can work.
### Enviroment Details
- Karma version (output of `karma --version`):
  0.13.19
- Relevant part of your `karma.config.js` file
### Steps to reproduce the behaviour

1.
2.
3.

shirish87:
Reason for this is the use of `Uint8Array` in the [1.4.8 release](https://github.com/socketio/socket.io-client/blob/2bed4f433dcc09487c9be15ff013d767570aa8d9/socket.io.js#L1227) of `socket.io-client`, which isn't natively supported in IE9.

dignifiedquire:
Yes #2188 shows that these failures happen. Looks like we need to add a shim for Uint8Array now

dignifiedquire:
Ref https://github.com/socketio/engine.io/issues/407

vedharish:
Using [this](https://github.com/inexorabletash/polyfill/blob/9fdf3e202a02aa27318ea1895e4b0516f9434e83/typedarray.js) as a polyfill for socket-io 1.4.8 for IE9 seems to work

PR #2486, opened 2016-12-06
chore: upgrade socket.io

wesleycho:
- Upgrade socket.io to v1.7.1, which contains fix for the Uint8Array regression in IE9

This should adequately address #2190.


dignifiedquire:
Thanks :octocat:

Issue #694, opened 2013-08-11
Propose: capture window.confirm


DemonShi:
As karma already captures window.alert, I think it would be useful to capture window.confirm too.

PR #2492, opened 2016-12-08
feat(client): capture confirm & prompt

wesleycho:
- Capture `window.confirm` & `window.prompt` usages in client

I think this addresses #694, although it is a little vague as to how it is desired to be done

### DIFF
--- a/context/karma.js
+++ b/context/karma.js
@@ -80,8 +80,21 @@ var ContextKarma = function (callParentKarmaMethod) {
       self.log('dump', arguments)
     }
 
+    var _confirm = contextWindow.confirm
+    var _prompt = contextWindow.prompt
+
     contextWindow.alert = function (msg) {
       self.log('alert', [msg])
+    }
+
+    contextWindow.confirm = function (msg) {
+      self.log('confirm', [msg])
+      _confirm(msg)
+    }
+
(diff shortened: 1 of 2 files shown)


context/karma.js
```diff
@@ -84,6 +84,14 @@ var ContextKarma = function (callParentKarmaMethod) {
       self.log('alert', [msg])
     }
 
+    contextWindow.confirm = function (msg) {
+      self.log('confirm', [msg])
+    }
+
+    contextWindow.prompt = function (msg, defaultVal) {
+      self.log('prompt', [msg, defaultVal])
+    }
```
dignifiedquire: I'm not sure, but I feel like we should call the original as well, because if a test relies on it that would fail now if I understand it correctly

wesleycho:
Updated - should we consider changing the alert behavior to match these new ones at some point too for parity? I think the normal alert will return undefined, so the current implementation does match the return, but doesn't match the browser freezing of JS block execution.

dignifiedquire:
Probably, but I'm not sure given that this behaviour was around for so long already, and nobody ever brought it up.

dignifiedquire:
Looks like CI/tests are not happy

dignifiedquire:
Thank you :octocat:

Issue #2228, opened 2016-07-01
browserConsoleLogOptions does not filter console by level


jambonrose:
### Expected behavior

I expect that if `browserConsoleLogOptions.level === 'error'`, then any instance of `console.log`, `console.debug` or `console.warn` will not appear in my terminal.
### Actual behavior

Instances of `console.debug` are still shown in my console, despite the following in my Karma configuration:

``` console
    browserConsoleLogOptions: {
      level: 'error',
      format: '%b %T: %m',
      terminal: true
    },
```

Specifically, `console.debug('hello there!');` results in `DEBUG: 'hello there!'` being printed when I run tests.

Note that switching `browserConsoleLogOptions.terminal` to `false` correctly silences all `console` output.
### Enviroment Details

Karma 1.1.0 with Webpack 2 through Babel 6 on macOS 10.11.5 with npm 3.8.8 and node v5.11.0. Please see [package.json](https://github.com/jambonsw/webpack2-karma-logging-error-mve/blob/master/package.json) for full requirements.
### Steps to reproduce the behaviour
1. Clone the example repo found at https://github.com/jambonsw/webpack2-karma-logging-error-mve
2. Install the packages with `npm install`
3. Run tests with `npm test`

vigreco:
It seems that `browserConsoleLogOptions.level` is only taken into account for log files. There's no filter for reporters.
I'd like to help out with this one.

Also, I noticed that the current log levels are: `['log', 'error', 'warn', 'info', 'debug']`
Since **log** has an higher level than **error**, console.log messages are always printed also if `browserConsoleLogOptions.level` is set to `'error'`.
@dignifiedquire is this a wanted behaviour?

romulocintra:
@dignifiedquire  if you describe the correct behaviour i can take over this  bugfix and propose a pr with possible solution

dignifiedquire:
@vigreco yes we should keep this. The reasoning being that you usually use `console.log` for debugging, while still wanting to keep only errors otherwise.

blackswanny:
can you merge it?

IgorAufricht:
This change made the LOG level messages the lowest priority, which I think should not be the case based on the comment by @dignifiedquire above:
> This should be changed to what was described in the original report. Moving log down between warn and info.

See #2602.

PR #2493, opened 2016-12-08
fix(browser): filter browser logging by level

wesleycho:
- Do not log lower levels if `browserConsoleLogOptions.level` is set

This should address #2228.

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -14,6 +14,14 @@ exports.LOG_ERROR = 'ERROR'
 exports.LOG_WARN = 'WARN'
 exports.LOG_INFO = 'INFO'
 exports.LOG_DEBUG = 'DEBUG'
+exports.LOG_PRIORITIES = [
+  exports.LOG_DISABLE,
+  exports.LOG_ERROR,
+  exports.LOG_WARN,
+  exports.LOG_INFO,
+  exports.LOG_DEBUG,
+  exports.LOG_DISABLE
+]
 
 // Default patterns for the pattern layout.
 exports.COLOR_PATTERN = '%[%d{DATE}:%p [%c]: %]%m'
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -1,5 +1,6 @@
 var util = require('util')
 
+var constants = require('../constants')
 var helper = require('../helper')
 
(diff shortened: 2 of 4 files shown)


lib/reporters/base.js
```diff
@@ -65,14 +66,19 @@ var BaseReporter = function (formatError, reportSlow, useColors, browserConsoleL
 
   this.onBrowserLog = function (browser, log, type) {
     if (!browserConsoleLogOptions || !browserConsoleLogOptions.terminal) return
+    type = type.toUpperCase()
+    if (browserConsoleLogOptions.level) {
+      var logPriority = constants.LOG_PRIORITIES.indexOf(browserConsoleLogOptions.level.toUpperCase())
+      if (constants.LOG_PRIORITIES[type] > logPriority) return
```
dignifiedquire: Isn't `constants.LOG_PRIORITIES[type]` a string, rather than a number?

wesleycho: Whoops...that was meant to be indexOf

lib/constants.js
```diff
@@ -14,6 +14,14 @@ exports.LOG_ERROR = 'ERROR'
 exports.LOG_WARN = 'WARN'
 exports.LOG_INFO = 'INFO'
 exports.LOG_DEBUG = 'DEBUG'
+exports.LOG_PRIORITIES = [
```
blackswanny: I think you missed to add the 'LOG' level itself to constants.js, 
exports.LOG_LOG = 'LOG'

exports.LOG_PRIORITIES = [
      exports.LOG_DISABLE,
      exports.LOG_ERROR,
      exports.LOG_WARN,
      exports.LOG_INFO,
      exports.LOG_DEBUG,
      exports.LOG_DISABLE,
      exports.LOG_LOG
]

wesleycho: There is no `LOG_LOG` in this file.

blackswanny: right, but we need to add it.
LOG_LOG maps to 'console.log' function as well as LOG_DEBUG maps to 'console.debug'
Try your pull request with console.log and browserConsoleLogOptions.level = 'debug' and you will see that it doesn't work, cause 
constants.LOG_PRIORITIES.indexOf(type)
will have -1 (not found)

dignifiedquire:
Thank you :octocat:

PR #2506, opened 2016-12-14
Fix Log Levels in Docs

darcyadams:
The `log4js-node` [docs](https://github.com/nomiddlename/log4js-node/wiki/Category-levels) list different log level constants. When switching over to these, `logLevel` configuration option starts working.

### DIFF
--- a/docs/config/01-configuration-file.md
+++ b/docs/config/01-configuration-file.md
@@ -409,11 +409,11 @@ httpsServerOptions: {
 
 **Possible values:**
 
-  * `config.LOG_DISABLE`
-  * `config.LOG_ERROR`
-  * `config.LOG_WARN`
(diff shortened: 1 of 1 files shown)


darcyadams:
I signed it!

dignifiedquire:
I am not sure how this is supposed to work, the options you mention above are not defined anywhere in the config file: https://github.com/karma-runner/karma/blob/master/lib/config.js

darcyadams:
if you look at `logger.js` it appears you are passing the log level directly into `log4js`...
https://github.com/karma-runner/karma/blob/master/lib/logger.js#L40
..and the log level constants in their docs do not match yours. In my testing using their log level constants work, while your do nothing and leave the log level at it's default.

PR #2533, opened 2017-01-12
fix(runner): Make process kill timeout configurable

### LINKED ISSUE #2447: <any_browser_here> process gets killed forcibly after 2000 ms
### Expected behaviour
The behaviour as to when to kill the browser should be configurable. There is a hardcoded value in karma/lib/launchers/process.js which is `var killTimeout = 2000`

### Actual behaviour
Since the value is not configurable the browser gets killed after 2s. If there's something running within the browser (e.g. when using karma-coverage and istanbul works on a large number of files), it gets killed.

### Environment Details

- Relevant part of your `karma.config.js` file: not available

1. Configure karma-coverage to run on a big code base
2. Run karma
3. Observe how the browser gets killed after 2s before it has a chance to finish processing the coverage data

vivganes:
Add a new user configurable property processKillTimeout to set timeout for any spawned process
Closes #2447

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -199,6 +199,10 @@ var normalizeConfig = function (config, configFilePath) {
     throw new TypeError('Invalid configuration: formatError option must be a function.')
   }
 
+  if (config.processKillTimeout && !helper.isNumber(config.processKillTimeout)) {
+    throw new TypeError('Invalid configuration: processKillTimeout option must be a number.')
+  }
+
   var defaultClient = config.defaultClient || {}
   Object.keys(defaultClient).forEach(function (key) {
     var option = config.client[key]
@@ -321,6 +325,7 @@ var Config = function () {
   this.browserDisconnectTimeout = 2000
   this.browserDisconnectTolerance = 0
   this.browserNoActivityTimeout = 10000
+  this.processKillTimeout = 2000
   this.concurrency = Infinity
   this.failOnEmptyTestSuite = true
   this.retryLimit = 2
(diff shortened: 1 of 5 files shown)


wesleycho:
Is this something that can be tested? I'm not seeing any pre-existing tests around whether it kills it after 2s.

vivganes:
@wesleycho  I was thinking that the existing tests in process.spec.js tested that. I got some tests failing because the value 2000 was not given in my earlier commit.  

I had to add another commit to fix the test cases.

I did not know how to run test cases locally earlier. Now, I know how to run them.  So, will make sure to run them before pushing my commit.

wesleycho:
Ah, I was looking at the wrong area.

This generally looks good to me though approach-wise.

vivganes:
@dignifiedquire Thanks for the suggestion. Just added tests. Can you please check if they look good?

dignifiedquire:
Thank you :octocat:

dignifiedquire:
@vivganes I just realised there was no docs update, could you PR that please, so that the documentation shows this new option?

vivganes:
Sure. I will do that today.

On 18-Jan-2017 10:33 PM, "Friedel Ziegelmayer" <notifications@github.com>
wrote:

@vivganes <https://github.com/vivganes> I just realised there was no docs
update, could you PR that please, so that the documentation shows this new
option?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/karma-runner/karma/pull/2533#issuecomment-273535371>,
or mute the thread

PR #2536, opened 2017-01-15
feat(config): allow config to be a default export

mattlewis92:
This allows folks using typescript karma config files to do
```
// karma.conf.ts
export default function(config) {
  config.set({
    basePath: './'
  })
}
```


dignifiedquire:
Thank you :octocat:

PR #2537, opened 2017-01-15
chore: switch from ghooks to husky

mattlewis92:
ghooks is now deprecated in favour of husky. Also requires your local git checkout to be nuked and started from scratch to uninstall the old ghooks


dignifiedquire:
Thank you :octo

PR #2541, opened 2017-01-19
docs(runner): Document processKillTimeout property

### LINKED ISSUE #2447: <any_browser_here> process gets killed forcibly after 2000 ms

vivganes:
Documentation for processKillTimeoutProperty
Closes #2447

### DIFF
--- a/docs/config/01-configuration-file.md
+++ b/docs/config/01-configuration-file.md
@@ -200,6 +200,7 @@ The `captureTimeout` value represents the maximum boot-up time allowed for a bro
 If any browser does not get captured within the timeout, Karma will kill it and try to launch
 it again and, after three attempts to capture it, Karma will give up.
 
+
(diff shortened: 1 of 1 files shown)


dignifiedquire:
Thank you :octocat:

PR #2542, opened 2017-01-20
feat(config): a 3rd option to run the tests by dynamically loading test scripts without using iframe or new window

chan1cyrus2:
Implement a new option for user to run their test. The new option doesn't use iframe or a new window to run the tests. It combines both client.html and context.html files and then load the scripts dynamically when execute is received. The call flow remains the same as the other two options. This solution is used on our project that we need to support a lightweight browser which does not support both iframe and opening a new window. Since the scripts are loaded dynamically, the loading time is longer than the other two options.


chan1cyrus2:
I signed it!

zzo:
@dignifiedquire what do you think?

console.log
```diff
@@ -1,5 +1,5 @@
-PhantomJS 2.1.1 (Mac OS X 0.0.0)
-PhantomJS 2.1.1 (Mac OS X 0.0.0)
-PhantomJS 2.1.1 (Mac OS X 0.0.0)
-PhantomJS 2.1.1 (Mac OS X 0.0.0)
-PhantomJS 2.1.1 (Mac OS X 0.0.0)
+PhantomJS 2.1.1 (Linux 0.0.0)
+PhantomJS 2.1.1 (Linux 0.0.0)
+PhantomJS 2.1.1 (Linux 0.0.0)
+PhantomJS 2.1.1 (Linux 0.0.0)
+PhantomJS 2.1.1 (Linux 0.0.0)
```
dignifiedquire: this seems like an unrelated change

dignifiedquire: Not even sure why this file exists in the first place tbh

static/client_with_context.html.html
```diff
@@ -0,0 +1,125 @@
+
+<!DOCTYPE html>
```
dignifiedquire: `.html.html`?

chan1cyrus2: my bad

chan1cyrus2:
Thanks!

I have added a e2e test case to cover the scenario (first time writing test case on cucumberjs, let me know if I do sth wrong) and updated the documentation on configuration.

I am not sure the part on overriding client_with_context.html in the same way the other templates can be overridden, can you elaborate more?

chan1cyrus2:
I have added a e2e test case to cover the scenario (first time writing test case on cucumberjs, let me know if I do sth wrong) and updated the documentation on configuration.

I am not sure the part on overriding client_with_context.html in the same way the other templates can be overridden, can you elaborate more?

dignifiedquire:
> I am not sure the part on overriding client_with_context.html in the same way the other templates can be overridden, can you elaborate more?

there is an config option `customContextFile`, which is used here: https://github.com/karma-runner/karma/blob/master/lib/middleware/karma.js#L146 to allow the user to customise the html of the actual file being loaded in the browser. It would be good to have this option for the new template as well.

chan1cyrus2:
Thanks! I have added the option. Please review.

Issue #2556, opened 2017-01-30
karma 1.4.0 broke IE8 support


gkatsev:
### Expected behaviour
Running karma in IE8 is possible.

### Actual behaviour
IE8 does not work anymore.

### Environment Details

- Karma version (output of `karma --version`):
- Relevant part of your `karma.config.js` file

Karma version: 1.4.1
My karma config is available [here](https://github.com/videojs/video.js/blob/master/test/karma.conf.js)

### Steps to reproduce the behaviour

1. Start karma with something like
```sh
karma start test/karma.conf.js --no-single-run
```
2. navigate to `0.0.0.0:9876` once it loads
3. notice that there are errors in the console.
![errors in ie8](https://cldup.com/DscjamRE5N.png)

I noticed this first when greenkeeper updated the dependency to karma 1.4.0. After several travis build retries, it still wouldn't connect to IE8 (see [build](https://travis-ci.org/videojs/video.js/builds/196417985#L1193-L1199)).
It looks like it's related to the socket.io update and probably broke inadvertently.
Dropping back down to version 1.3.0 of karma, everything works great still.

Potential solutions, they're not mutually exclusive:
* add the `es3` preset to make sure that words like `default` are quoted properly.
* investigate why the socket.io update causes an error
* drop support for IE8 in a major update

I'd be happy to try and help making the first two options happen. In the meantime, I'm staying with the 1.3.0 version of karma.

gkatsev:
Here's greenkeeper's issue against the videojs project which alerted me to this: https://github.com/videojs/video.js/issues/3946

dignifiedquire:
Thank you for the report. If you could help us figure out what exactly the issues are, that would be great. As far as I understood socket.io broke IE8 but fixed it later at which point I upgraded the dependency. But it sounds like sth ist still not right. 
Agreed on using ES3 preset, want to do a PR for that?

gkatsev:
I'll make a PR for the es3 preset and I'll investigate the socket.io thing. Thanks for the quick response!

gkatsev:
The es3 preset PR: https://github.com/karma-runner/karma/pull/2557

gkatsev:
So, the socket.io issue is that uglify 2.7 disabled support for IE8 in a minor version update. Since karma  uses the minified file it broke. Switching to the unminified file works great. I'll open an issue/PR against socket.io-client

gkatsev:
[...]

PR #2557, opened 2017-01-30
chore(package): add es3 babel preset

gkatsev:
Adding the es3 babel preset ensures the highest compatibility with
browsers like IE8.
In the babelrc file, the preset comes before the es2015 preset because
the presets are executed in reverse order. So, the es2015 preset will
run before the es3 preset, which is what we want to have happen.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,3 @@
 {
-  "presets": ["es2015"]
+  "presets": ["es3", "es2015"]
 }
--- a/package.json
+++ b/package.json
@@ -342,6 +342,7 @@
     "LiveScript": "^1.3.0",
     "babel": "^6.5.2",
     "babel-preset-es2015": "^6.9.0",
+    "babel-preset-es3": "^1.0.1",
     "babel-register": "^6.9.0",
     "chai": "^3.3.0",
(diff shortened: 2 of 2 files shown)


gkatsev:
I signed it!

dignifiedquire:
I just realized that the current setup only runs the client files through browserify, but without using babel. So this is actually not going to have any effect other than changing the output when we transpile our tests. Would you mind adding babelify to https://github.com/karma-runner/karma/blob/master/gruntfile.js#L13 that should help :)

gkatsev:
Yup, I'll do that.

gkatsev:
So, investigating #2556, and now I'm not sure that this is strictly necessary right now. But I can still complete this PR, if that's wanted.

dignifiedquire:
Thanks @gkatsev lets see if we can get socket.io fixed and if that's enough then I'm happy to leave this alone.

gkatsev:
Ok, I'll close this PR then. Worst case scenario, karma can switch to loading the unminified file.

gkatsev:
I'll re-open this PR if it ends up being necessary.

PR #2561, opened 2017-02-08
docs(contributing): add sample commit message

danielcompton:
Add an example commit message so contributors can see a concrete example of what is expected.


dignifiedquire:
Thank you :octocat:

PR #2562, opened 2017-02-08
feat(watcher): Debounce autoWatchBatchDelay

### LINKED ISSUE #2331: Add autoWatchDelay option
## Problem

When incrementally recompiling ClojureScript, many files are changed as the compilation process runs. Currently, it is difficult to use Karma to autorun tests after detecting a ClojureScript recompile, because as soon as the first file is changed, Karma wants to rerun the tests.
## Current behaviour

There is a config option `autoWatchBatchDelay` available which at first seems like it would be ideal for this use case. It starts a timer after it detects the _first_ file system change, and waits n milliseconds before running tests. It is possible to use this option, but it is suboptimal. 

The timer starts from the first change detected, so when configuring this parameter, I need to specify what the maximum time my ClojureScript compile could take, so it doesn't start rerunning while recompiling. I can shoot for a low number that my app will recompile in (say 1 second), but then will hit problems if an incremental recompile takes longer (which happens from time to time when editing certain files).
## Proposed behaviour

I propose that there is a new option `autoWatchDelay` (or perhaps changing the behaviour of the current option) to start the delay timer from the _last_ filesystem change detected, waiting n more milliseconds before starting the tests. In practice this might look like resetting the timer, each time a new file change was detected.

IMHO this behaviour is probably closer to what most people are wanting when they use `autoWatchBatchDelay`.

danielcompton:
I'm opening this up for feedback early, to make sure I'm on the right track here. My intention is to change the behaviour of `autoWatchBatchDelay` to debounce file change events, and not run the tests until at least `autoWatchBatchDelay` ms have elapsed since the last file change.

Closes #2331

### DIFF
--- a/lib/file-list.js
+++ b/lib/file-list.js
@@ -49,13 +49,13 @@ function byPath (a, b) {
 // emitter       - EventEmitter
 // preprocess    - Function
 // batchInterval - Number
-var List = function (patterns, excludes, emitter, preprocess, batchInterval) {
+var List = function (patterns, excludes, emitter, preprocess, autoWatchBatchDelay) {
   // Store options
   this._patterns = patterns
   this._excludes = excludes
   this._emitter = emitter
   this._preprocess = Promise.promisify(preprocess)
-  this._batchInterval = batchInterval
+  this._autoWatchBatchDelay = autoWatchBatchDelay
 
   // The actual list of files
   this.buckets = new Map()
@@ -71,14 +71,14 @@ var List = function (patterns, excludes, emitter, preprocess, batchInterval) {
   var self = this
 
   // Emit the `file_list_modified` event.
-  // This function is throttled to the value of `batchInterval`
-  // to avoid spamming the listener.
+  // This function is debounced to the value of `autoWatchBatchDelay`
+  // to avoid reloading while files are still being modified.
   function emit () {
     self._emitter.emit('file_list_modified', self.files)
   }
-  var throttledEmit = _.throttle(emit, self._batchInterval, {leading: false})
+  var debouncedEmit = _.debounce(emit, self._autoWatchBatchDelay)
   self._emitModified = function (immediate) {
-    immediate ? emit() : throttledEmit()
+    immediate ? emit() : debouncedEmit()
   }
 }
 
--- a/docs/config/01-configuration-file.md
+++ b/docs/config/01-configuration-file.md
@@ -89,8 +89,9 @@ These are all of the available configuration options.
 
 **Description:** When Karma is watching the files for changes, it tries to batch
 multiple changes into a single run so that the test runner doesn't try to start and restart running
-tests more than it should. The configuration setting tells Karma how long to wait (in milliseconds) after any changes
-have occurred before starting the test process again.
+tests more than it should, or restart while build files are not in a consistent state. The configuration setting
+tells Karma how long to wait (in milliseconds) from the last file change before starting
+the test process again, resetting the timer each time a file changes (i.e. [debouncing](https://davidwalsh.name/javascript-debounce-function)).
 
 
 ## basePath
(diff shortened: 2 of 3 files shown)


dignifiedquire:
Thanks @danielcompton, I agree this change makes sense. There are probably some tests that need updating after this change.

danielcompton:
Yeah there's some tests that fail, and I still need to verify the behaviour is correct.

danielcompton:
(Probably also adding more tests)

wesleycho:
I think all that is needed here is to just adjust the delay in the tests to bounce it back a little further.

danielcompton:
Thanks for the tip, you're right that did fix those tests. 

I modified the `[batch interval](https://github.com/karma-runner/karma/pull/2562/files#diff-a9b72b9949983d5d75e4e99605949cd9R744)` test to test the debouncing, and it seems like that isn't working correctly. I'm not sure if this is because the tests need to mock more of lodash, or if there is something else going on? I'm at about the limits of my JS knowledge, if you can take a quick look to see if there's something obvious I'm not doing, I'd appreciate it :)

lib/file-list.js
```diff
@@ -71,14 +71,14 @@ var List = function (patterns, excludes, emitter, preprocess, batchInterval) {
...
 
   // Emit the `file_list_modified` event.
-  // This function is throttled to the value of `batchInterval`
+  // This function is debounced to the value of `batchInterval`
   // to avoid spamming the listener.
   function emit () {
     self._emitter.emit('file_list_modified', self.files)
   }
-  var throttledEmit = _.throttle(emit, self._batchInterval, {leading: false})
+  var debouncedEmit = _.debounce(emit, self._batchInterval, {leading: false})
```
wesleycho: I don't think `{leading: false}` needs to be passed into the function here - debouncing never fires the leading call.

wesleycho:
Not sure if this helps you, but in another PR for upgrading dependencies (including lodash), I had to make this adjustment: https://github.com/wesleycho/karma/blob/c80a3161dd4c345be2ec00c6691811b5dc9b2a5d/test/unit/file-list.spec.js#L770-L805

lib/file-list.js
danielcompton: Thanks, fixed.

danielcompton:
Phew! I think I've got it all.

In the [old version](https://github.com/karma-runner/karma/blob/2a847c250bb62134d87f5230d97be8483d4a13cf/test/unit/file-list.spec.js#L730-L750), the tests were passing, but they weren't testing quite what it looked like. `list.removeFile`, `list.changeFile, and `list.addFile` all return promises and call `self._emitModified` when a file modification is accepted as valid. 

There is an important difference in behaviour between `list.removeFile` and the other too though. Due to the promise structure of `list.removeFile`, `self._emitModified` is called synchronously. The test code was calling all of these functions one after another without waiting on the result. The only function call that had any effect on the debouncing function running while inside the test scope were the `list.removeFile` calls. You can verify this by commenting out the two `list.removeFile` calls, and watch the test fail.

```
        list.changeFile('/some/b.js')
        // list.removeFile('/some/a.js') // /some/b.js, /a.txt
        // list.removeFile('/a.txt')     // /some/b.js
        list.addFile('/a.txt')        // /some/b.js, /a.txt
        list.addFile('/some/0.js')    // /some/0.js, /some/b.js, /a.txt

        clock.tick(99)
        expect(modified).to.not.have.been.called

        clock.tick(2)
        expect(modified).to.have.been.calledOnce // This will fail
```

In my tests, I extended this test with more calls to modify and change files to extend the debouncing period. However because the promises weren't being waited on, they never ran until after the test block had finished, leading to very confusing behaviour where removing a file makes the tests work, but changing or adding a file doesn't.

I've converted the tests to wait on the promises and added more and they all pass.

danielcompton:
This is ready for review now.

danielcompton:
@wesleycho would you able to take another review please?

dignifiedquire:
Thanks a lot for getting through this and especially updating the tests to correctly test what they should test!

PR #2564, opened 2017-02-10
fix: make window.parent.karma available in debugged context

dryajov:


dryajov:
I signed it!

dignifiedquire:
Thanks :octocat:

Issue #2581, opened 2017-02-21
Git Commit Msg document error


sabi0:
The page at http://karma-runner.github.io/1.0/dev/git-commit-msg.html  says that the commit message body 
> uses the imperative, present tense: “change” not “changed” nor “changes”

However, the example at the top of the page has the body starting with "Added".

dignifiedquire:
Thank you for spotting this. Want to file a PR to update the example?

sabi0:
I think a PR would be an overkill for such a minor fix. Feel free to commit the change yourself.

PR #2583, opened 2017-02-22
docs: Correct tense of sample commit desc

vivganes:
Sample commit desc is in present tense now. 
Closes https://github.com/karma-runner/karma/issues/2581


dignifiedquire:
Thanks :octocat:

PR #2591, opened 2017-03-02
Does not work with Mootools

Ondoher:
The for..in loop was not working with mootools because mootools  adds functions to the array prototype.

checking for own property

### DIFF
--- a/lib/middleware/karma.js
+++ b/lib/middleware/karma.js
@@ -179,6 +179,10 @@ var createKarmaMiddleware = function (
             var filePath = file.path
             var fileExt = path.extname(filePath)
 
+            if (!files.included.hasOwnProperty(i)) {
+              continue
+            }
+
             if (!file.isUrl) {
               filePath = filePathToUrlPath(filePath, basePath, urlRoot, proxyPath)
 


Ondoher:
i signed it

Ondoher:
i signed it

Ondoher:
i signed it

Ondoher:
i signed it!

Ondoher:
I signed it!

Ondoher:
I signed it!

Ondoher:
I signed it!

lib/middleware/karma.js
```diff
@@ -179,6 +179,10 @@ var createKarmaMiddleware = function (
             var filePath = file.path
             var fileExt = path.extname(filePath)
 
+            if (!files.included.hasOwnProperty(i)) {
+              continue
```
dignifiedquire: this `continue` means the things below are skipped in this case, I don't think we want to do that.

Ondoher: Perhaps. But let's say they added a new method called randomize to the Array prototype. I don't see any further processing you would want to do to that method. I'd think it should just be skipped.

Ondoher: What's holding up a merge?

dignifiedquire: I think this: https://github.com/karma-runner/karma/pull/2580/files#r104290773 is a better approach to fixing this issue

Ondoher: My original code changed things to a forEach loop, but I though there may be unwanted side effects or opportunities for bugs--for instance, the loop variable `i` is addressed in the loop. So, I kept it simple and just continued the loop on the next item if it was not an own property.

Ondoher: My original code changed things to a forEach loop, but I though there may be unwanted side effects or opportunities for bugs--for instance, the loop variable `i` is addressed in the loop. So, I kept it simple and just continued the loop on the next item if it was not an own property.

Ondoher: My original code changed things to a forEach loop, but I though there may be unwanted side effects or opportunities for bugs--for instance, the loop variable `i` is addressed in the loop. So, I kept it simple and just continued the loop on the next item if it was not an own property.

dignifiedquire: okay, but shouldn't the check then be the first thing in the loop?

Ondoher: I guess that depends on style. The style here seemed to be to have all the variable declarations at the top of the block, so I added my code underneath. 

None of the existing variable initializations should have a problem dealing with an undefined value, but it could break future code. I could go either way.

PR #2595, opened 2017-03-10
fix(client): add proxy support to stringify

twolfson:
In https://github.com/twolfson/karma-electron/issues/20, we received a bug report with respect to serializing Electron's remote modules. After some digging, it turns out this was being caused by a serialization attempt on `Proxy` wrapped functions:

https://github.com/electron/electron/blob/v1.6.3/lib/renderer/api/remote.js#L170-L190

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/toString

```js
new Proxy(function x () {}, {}).toString()
// VM594:1 Uncaught TypeError: Function.prototype.toString is not generic
//    at Proxy.toString (<anonymous>)
//    at <anonymous>:1:33
```

Unfortunately, there is no patch/workaround for serializing proxies. Although, they can provide a custom `toString()` function if they want:

```
new Proxy(function x () {}, {get: function () { return function () { return x.toString() }; }}).toString()
// "function x() {}"
```

As a result, we should attempt to serialize normally and fallback if we fail. In this PR:

- Added test for proxy behavior
- Added serialization fallback for proxies

Here's proof of it working with Electron:

```
Object{clipboard: Object{availableFormats: Proxy(function remoteMemberFunction(...) { ... }), has: Proxy(function remoteMemberFunction(...) { ... }), 
```

### DIFF
--- a/common/stringify.js
+++ b/common/stringify.js
@@ -19,7 +19,18 @@ var stringify = function stringify (obj, depth) {
     case 'undefined':
       return 'undefined'
     case 'function':
-      return obj.toString().replace(/\{[\s\S]*\}/, '{ ... }')
+      try {
+        // function abc(a, b, c) { /* code goes here */ }
+        //   -> function abc(a, b, c) { ... }
+        return obj.toString().replace(/\{[\s\S]*\}/, '{ ... }')
+      } catch (err) {
+        if (err instanceof TypeError) {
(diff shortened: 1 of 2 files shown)


dignifiedquire:
Thank you :octocat:

PR #2618, opened 2017-03-21
fix: return result of confirm and prompt

voltrevo:
When confirm and prompt were wrapped here https://github.com/karma-runner/karma/commit/3a618b3ce51f7fb8da94636562ba2ed785da0579, returning the results of these methods was forgotten. This fixes that.


dignifiedquire:
Thank you :octocat:

PR #2620, opened 2017-03-21
fix(watcher): use correct file sha in url query string

### LINKED ISSUE #2317: "nocache" option not working ?
### Expected behavior
It should prevent the browser to use cached files marked with `nocache` option set to `true`

### Actual behavior
It is not preventing the browser to use cached files marked with `nocache` option set to `true`

### Enviroment Details

- Karma version (output of `karma --version`): 1.1.2
- Relevant part of your `karma.config.js` file : 
```json
files: [
  ...
  {"pattern": "application.spec.js", "nocache": true},
  {"pattern": "templates.js", "nocache": true}
]
````

1. Setup your karma config file with `nocache: true` on some files/glob
2. Run karma with `singleRun=false`
3. Change something in a test that should fail, the test will not necessary fail. (Not happening all the times)

Also, I noticed when using `PhantomJS` that when not passing the `nocache` option, it will add a token next to the filename. Example : `application.spec.js?a1b2c3d4f5g6`. 

This token will not change unless with restart karma.

Putting the `nocache` option to `true`, the token will be `undefined` (`application.spec.js?undefined`) so maybe it's what is causing that cache issue ?

I would expect instead that there's a token and each time the test suite start it replaces that token with a new one, preventing cached files to be used.

### PR DESCRIPTION
**Pull request by chriscasola**
Use the most recently computed file sha present on the served file
when setting the file sha in the scripts tags added to the context.html.

This fixes an issue where the sha query arg would always be the sha of
the file when karma initially started and would not take into account
changes to the file when the watcher is running.

Closes #2317, #2264

---------------------------------

I think this fixes both issues linked above. From what I could tell from my debugging, the sha on the file object in the `served` files array gets updated each time the file changes and the preprocessors run, but the sha on the `included` files does not. However for cache busting in the file URL path the sha from `included` was always used so the browser would end up requesting the original version of the file and it would be served by the browser cache.

I can add a test for this, but I wanted to get an idea of if this was an acceptable solution.

### CODE REVIEW

**Comment by chriscasola** [comment]
Is there something I can do to work towards getting this merged? The travis CI build seems to be failing because of a lint error in the master branch.

**Comment by dignifiedquire** [comment]
Thank you, could you rebase onto master so CI is happy please?

**Comment by maksimr** [comment]
@chriscasola can you reproduce this problem in the test?
I think it will be great if we catch it in tests.

Thanks

**Comment by maksimr** [comment]
@chriscasola are you sure that this PR fix problem  https://github.com/karma-runner/karma/issues/2317?

Because as I see there a problem with **"nocache"**

```
Setting nocache to false it does : application.spec.js?a1b2c3d4f5g6
Setting nocache to true it does : application.spec.js?undefined <-----
```
https://github.com/karma-runner/karma/issues/2317#issuecomment-240161662

**Comment by maksimr** [comment]
Also, I don't sure that this PR fixes the issue https://github.com/karma-runner/karma/issues/2264
because there a problem with preprocessing files when user set **nocache**

**Comment by chriscasola** [comment]
@maksimr -- I will work on adding a test to replicate the issue and I'll take another look at the issues I referenced. 

The info in #2317 seems to indicate that in watch mode karma is not picking up changes to files, which is exactly the issue I was having that this PR fixes. I don't think what is described there is actually related to the "nocache" option but I will confirm.

As far as #2264 I think you're right that it is unrelated, I'll remove that reference from the commit message.

Thanks for your feedback!

**Comment by maksimr** [comment]
@chriscasola first of all thank you for your help.

About issue #2317 files with property **doNotCache**  shouldn't  have a *sha*(hash) at all. (IMHO)

**Comment by chriscasola** [comment]
@maksimr I'm not really sure what `noCache` should do in that case. According to [this documentation](https://github.com/karma-runner/karma/blob/e5d792dd64af63a4b8e53392c3574dfa87f45c1a/docs/config/02-files.md#nocache) it simply means karma should not cache the file contents in memory, and instead read from disk each time it is requested. I think that is separate from whether karma should prevent the browser from caching it or not.

I think the person who originally filed that issue was trying to use the `noCache` option because they noticed the same bug I did, which was that karma was rerunning tests on save, but not with the latest version of the test files.

**Comment by maksimr** [comment]
@chriscasola Well, Let's fix the problem and wait for the feedback

Thanks!

**Comment by chriscasola** [comment]
@maksimr -- I added a test and rebased

**Comment by maksimr** [comment]
@chriscasola thanks

I had a look and I think that the current solution fixes the circumstances not the source of the problem.

We have a *FileList* class which managed files.
It contains property [*files*](https://github.com/karma-runner/karma/blob/master/lib/file-list.js#L215). Each time when we try take a value of this property we call [getter method](https://github.com/karma-runner/karma/blob/master/lib/file-list.js#L216). Getter always return [new object](https://github.com/karma-runner/karma/blob/master/lib/file-list.js#L254-L257).
Returned object contains two property *served* and *included*.
Served means that file should be served by the karma web server.
Included means that file should be included on the page by tag script.

Also, *FileList* has a method [*changeFile*](https://github.com/karma-runner/karma/blob/master/lib/file-list.js#L341) this method called by [watcher](https://github.com/karma-runner/karma/blob/master/lib/watcher.js) [when file changed](https://github.com/karma-runner/karma/blob/master/lib/watcher.js#L101).
Inside this method, we [find a file](https://github.com/karma-runner/karma/blob/master/lib/file-list.js#L345) by path passed from watcher after that we [run preprocessors](https://github.com/karma-runner/karma/blob/master/lib/file-list.js#L359) on the changed file.

Inside Preprocessor we [set *sha*](https://github.com/karma-runner/karma/blob/master/lib/preprocessor.js#L32) to the file after all preprocessors.

Ok, *FileList* waiting while preprocessor works when preprocessor finish work *FileList* [trigger event file_list_modified](https://github.com/karma-runner/karma/blob/master/lib/file-list.js#L363) and [pass *FileList.files* as first](https://github.com/karma-runner/karma/blob/master/lib/file-list.js#L77) argument.

[Karma server](https://github.com/karma-runner/karma/blob/master/lib/server.js#L345) and [WebServer](https://github.com/karma-runner/karma/blob/master/lib/web-server.js#L45) listen event file_list_modified.

Karma server on file_list_modifie [emit event by socket to the browser](https://github.com/karma-runner/karma/blob/master/lib/server.js#L350) and say browser restart tests.

WebServer on file_list_modifie event [save link on *files* inside *filesPromise*](https://github.com/karma-runner/karma/blob/master/lib/web-server.js#L45).

After that browser mad request to the server for getting files. Here we go to [karma middleware](https://github.com/karma-runner/karma/blob/master/lib/middleware/karma.js#L87)
which handles the request. It takes *filePromise* and [get files](https://github.com/karma-runner/karma/blob/master/lib/middleware/karma.js#L150) from it.
For included fiels it [generate scripts](https://github.com/karma-runner/karma/blob/master/lib/middleware/karma.js#L190-L211) for served files it [generate map](https://github.com/karma-runner/karma/blob/master/lib/middleware/karma.js#L215-L223) which also passed to the client.

As you can see in karma middleware we have another request which uses included files for [example '/context.json'](https://github.com/karma-runner/karma/blob/master/lib/middleware/karma.js#L251-L255).

So first, I didn't see where we can loose the *sha* and get the current problem.
Second I think if *sha* wrong it should be fixed in another place for example in FileList becase we have code in different places which rely on *sha* from included file.

**Comment by chriscasola** [comment]
@maksimr I dug into this more and I think I see what is happening now.

In my project that is using karma but seeing the issue with changes not being picked up by the watch, we have the `files` property of our config set to something like:

```
[
  'src/**/*.js',
  'src/**/*.spec.js',
]
```

Obviously, it seems the second matcher there is redundant and can be removed to fix our issue, but because both of the matchers match files ending in `.spec.js` our test files end up appearing in the `FileList` under multiple patterns.

In [changeFile](https://github.com/karma-runner/karma/blob/master/lib/file-list.js#L345) the version of the file matched by the second matcher is found and preprocessed. Later on when the [files getter](https://github.com/karma-runner/karma/blob/master/lib/file-list.js#L226) is called, the second occurrence of the file, the one that was preprocessed, is put in the `served` list but the `included` list ends up pulling the first occurrence, which has the wrong sha and old file contents.

So in summary, I think the real bug here, is that if a file is matched by more than one file matcher from the karma config, changes are not picked up during the watch.

What would you recommend for a solution here? Ideas that come to mind are:

1. Show a warning if the file matchers match more than one file. This doesn't seem ideal since I don't think it makes sense to force people to write matchers that don't overlap.
1. Always use the file matched by the first matcher
1. Always use the file matched by the last matcher

**Comment by maksimr** [comment]
@chriscasola Thanks! This is very valuable information/research because now it sounds familiar. 
I think we already had such issue and maybe still have.

https://github.com/karma-runner/karma/issues/1499

> What would you recommend for a solution here?
For me, we should have only one file which can be presented in several patterns/buckets. 
I will look at the code.

And yeah, thanks again. Great job. 👍 

P.S. 
Could you create a repo which reproduces the problem If it's not a hassle?

**Comment by chriscasola** [comment]
@maksimr -- I think I see where the issue is and have a solution that ensures we always use the file from the first match. Just writing up a quick test and then I'll update this PR.

**Comment by maksimr** [comment]
Different things that maybe related to this problem:

https://github.com/raspo/karma/commit/353b9cf7e1c9a15e4d5263b4ba47fca37209fb29

**Comment by chriscasola** [comment]
@maksimr I just pushed what I think fixes the root of the problem, let me know what you think.

**Comment by chriscasola** [comment]
@maksimr @dignifiedquire can this be merged?

**Comment by dignifiedquire** [comment]
thank you :octocat:

**Comment by lukaw3d** [comment]
This is a huge regression. There should be an option to disable it

All projects that have a configuration similar to below are broken after this change
```js
files: [
  { pattern: 'src/**/*', included: false, served: true },
  { pattern: 'src/**/*.spec', included: true } // 232 tests
]

> Error 0 of 0 tests executed
```

and worse, projects with the following configuration are not even broken. It's just that most of their tests are not being run anymore
```js
files: [
  { pattern: 'src/**/*', included: false, served: true },
  { pattern: 'src/**/*.spec', included: true }, // 232 tests
  { pattern: 'src2/**/*.spec', included: true } // 81 tests
]

> 81 of 81 tests executed
```


---

And this breaks plugins that append adapters to the end of the files list
```js
files: [
  { pattern: 'src/**/*', included: false, served: true },
  { pattern: 'tests/**/*.spec', included: true },
  { pattern: 'node_modules/**/*', included: false, served: true }
  // <<< Plugin appends an adapter here. It is loaded from within node_modules
]
```

PR #2641, opened 2017-03-31
chore: ensure test failures show correct line number

### LINKED ISSUE #2500: Test failures do not display accurate test line numbers in stack trace
### Expected behaviour
The line number in a test file referenced when a test fails is correct

### Actual behaviour
The line number reported is incorrect, a virtual line number generated from Babel

1.  Go to https://github.com/karma-runner/karma/blob/e36ba6c1d4cad4fc321f3376129b329fe663068d/test/unit/browser.spec.js#L30
2. Change `.to.equal(fullName)` to `.to.equal(fullName) + '1'`
3. Observe the line number reported

This likely is a simple configuration problem with Babel or one of the grunt plugins

chriscasola:
Fixes #2500

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,4 @@
 {
-  "presets": ["es2015"]
+  "presets": ["es2015"],
+  "retainLines": true
 }


dignifiedquire:
Thank you :octocat:

PR #2658, opened 2017-04-11
Remove debug message which shows location of config file

JakeChampion:
I was using `config.LOG_DISABLE` and noticed this message was still being outputted to the console. I figured this message is not actually needed unless an error is thrown, and that case is already catered for in the code.


dignifiedquire:
`LOG_DISABLE` should hide this message, that is probably related to the issues around log levels. No need to remove this debug message.

JakeChampion:
I signed it!

JakeChampion:
Superseded by https://github.com/karma-runner/karma/pull/2837

PR #2666, opened 2017-04-18
Add a little paragraph about scopes

Fenntasy:
This is a proposal to explain a little bit more how to use scopes.

This is how I understood them but I would love to have your opinions on the different ways to use scope.

Should it be a real list across every projects? Project-specific each time? Should it talk about user-centric scopes or technical ones?


Fenntasy:
I signed it!

(even though, that's repulsing just to propose something on a repo :/)

wesleycho:
Thank you for the effort!

I'm not sure if this change would be good to add in this particular markdown file. If there is confusion as to what the scope should entail, we should flesh it out in an issue & determine how best to document it, as opposed to adding uncertainty inside the documentation IMO. This paragraph is a little at odds with the tenor of the document, which is explaining the requirements of the commit message.

PR #2676, opened 2017-04-26
fix(reporters): Revert the backwards-incompatible log priority order changes

### LINKED ISSUE #2582: Undocumented breaking change
### Expected behaviour
#2228 extended `browserConsoleLogOptions.level` to filter messages on the terminal.  This changes the behavior of karma so that under certain circumstances messages created with `console` no longer show up in the terminal without configuration changes.  This should be documented.

### Actual behaviour
There is no documentation.  In my case I spent over 3 hours investigating source code, git issues, versions, etc before I identified the problem was this change.

### Environment Details
- Karma version (output of `karma --version`): 1.5.0

mgol:
Commit 35965d9 changed the order of log priorities, making lots of
configurations no longer print console.log messages. This was a
backwards-incompatible change which this commit reverts.

Tests were added to ensure the proper order is respected.

Fixes #2582

----
Also, another commit in this PR:

test(reporters): Fix the log suppressing tests

The tests ensuring a lower priority doesn't get logged were written in an
incorrect format so even changing the order in the LOG_PRIORITIES array
would not make them fail.

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -20,9 +20,9 @@ exports.LOG_PRIORITIES = [
   exports.LOG_DISABLE,
   exports.LOG_ERROR,
   exports.LOG_WARN,
+  exports.LOG_LOG,
   exports.LOG_INFO,
-  exports.LOG_DEBUG,
-  exports.LOG_LOG
+  exports.LOG_DEBUG
 ]
 
 // Default patterns for the pattern layout.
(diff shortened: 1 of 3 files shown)


blackswanny:
according to Mozilla this is the right place for LOG, but people misuse it
https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/console#Setting_the_Logging_Level

mgol:
@blackswanny Regardless what's the correct way spec-wise this was a breaking change which shouldn't happen before 2.0. This PR restores backwards compatibility.

dignifiedquire:
Thank you @maksimr

mgol:
@dignifiedquire could you release a new version to npm?

dignifiedquire:
yeah sorry will do today

yeronhernandez2-png (approved):

PR #2678, opened 2017-04-27
Use `forEach` in lib/middleware/karma.js

### LINKED ISSUE #2671: Karma is not able to run if any methods in `Array.prototype` are modified beforehand.
### Expected behaviour

Karma should be running.

### Actual behaviour

Not able to parse the paths given in the `config.files` and throw an error:

```javascript
ERROR [karma]: TypeError: Path must be a string. Received undefined
    at assertPath (path.js:7:11)
    at Object.extname (path.js:887:5)
    at d:\Test\test-karma\node_modules\.1.6.0@karma\lib\middleware\karma.js:181:32
    at d:\Test\test-karma\node_modules\.1.6.0@karma\lib\middleware\common.js:117:36
    at d:\Test\test-karma\node_modules\.4.1.11@graceful-fs\graceful-fs.js:78:16
    at FSReqWrap.readFileAfterClose [as oncomplete] (fs.js:439:3)
```

### Environment Details

- Karma version (output of `karma --version`): `v1.6.0`
- Node.js version: `v7.9.0`
```javascript
module.exports = function (config) {
    config.set({
        browsers: ['PhantomJS'],
        files: [
            './test.js'
        ],
        singleRun: true,
        // coverage reporter generates the coverage
        reporters: ['progress'],
        plugins: [
            'karma-mocha',
            'karma-phantomjs-launcher'
        ]
    });
};
```

Run this snippet of script:
```javascript
var Karma = require('karma').Server;
var path = require('path');

var fn = Array.prototype.indexOf;
delete Array.prototype['indexOf'];

Array.prototype['indexOf'] = function () {
    return fn.apply(this, arguments);
};

new Karma({
    configFile: path.resolve(__dirname, 'karma.conf.js')
}).start();
```

### Cause
At [this line](https://github.com/karma-runner/karma/blob/master/lib/middleware/karma.js#L177) in `lib/middleware/karma.js`:
```javascript
for (var i in files.included) 
```
`for...in` is used to traverse an array and this is a known bad practice. Maybe there is a particular reason for using it here?

BenjaminVanRyseghem:
Use `forEach` instead of `for..in` as it's better
practices and fix #2671.

### DIFF
--- a/lib/middleware/karma.js
+++ b/lib/middleware/karma.js
@@ -174,14 +174,9 @@ var createKarmaMiddleware = function (
 
           var scriptTags = []
           var scriptUrls = []
-          for (var i in files.included) {
-            var file = files.included[i]
+          files.included.forEach(function (file) {
             var filePath = file.path
             var fileExt = path.extname(filePath)
-
-            if (!files.included.hasOwnProperty(i)) {
-              continue
-            }
 
             if (!file.isUrl) {
               filePath = filePathToUrlPath(filePath, basePath, urlRoot, proxyPath)
@@ -195,12 +190,12 @@ var createKarmaMiddleware = function (
 
             if (fileExt === '.css') {
               scriptTags.push(util.format(LINK_TAG_CSS, filePath))
-              continue
+              return
             }
 
             if (fileExt === '.html') {
               scriptTags.push(util.format(LINK_TAG_HTML, filePath))
-              continue
+              return
             }
 
             // The script tag to be placed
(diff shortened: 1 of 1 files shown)


BenjaminVanRyseghem:
I signed it!

dignifiedquire:
Thank you but we can not make this change, `.forEach` assumes that `files` is an unmodified array with an intact prototype, but we had issues where that was not the case.

BenjaminVanRyseghem:
Thanks for the pointer.

I'll see if I can get another workaround :smile:

PR #2679, opened 2017-04-28
Update 04-public-api.md

piperchester:
Wondering why other exposed functions use lowercase, and Server uses upper. It seems inconsistent. Users receive a TypeError otherwise.


dignifiedquire:
`Server` is a constructor, and `server` is legacy code, which we forgot to remove in `1.0`. https://github.com/karma-runner/karma/blob/master/lib/index.js#L10

PR #2689, opened 2017-05-06
chore: update dependencies

### LINKED ISSUE #1812: Upgrade to lodash 4.0.0

### LINKED ISSUE #2688: Update all dependencies to their latest
```
 lodash                   ^3.8.0  →      ^4.17.4
 log4js                  ^0.6.31  →       ^1.1.1
 cucumber                 ^1.2.0  →  ^2.0.0-rc.9
 eslint-config-standard   ^7.0.0  →      ^10.2.1
 eslint-plugin-react     ^6.10.0  →       ^7.0.0
 eslint-plugin-standard   ^2.0.1  →       ^3.0.1
 sinon                   ^1.17.5  →       ^2.2.0
```

wesleycho:
- Update dependencies as part of process to drop support for versions of Node.js below 4

Closes #1812
Closes #2688


wesleycho:
I had to make a couple of changes for the upgrade.

The sinon changes were because sinon deprecated the fake call api from within `stub`.
The eslint package additions were caused by the upgrade of the eslint-plugin-react plugin, which required those packages to have to be installed in order to lint - those also resulted in having to make the minor linting change adjustments.
`file-list.spec.js` has a few hacky changes to defer checking until after the `_.defer` calls are able to do their job inside `file-list.js`.

There is still one test that eludes me currently.

wesleycho:
This PR passes all CI checks except for Node 0.10 and 0.12 - it requires #2692 in order to properly pass.

dignifiedquire:
Before we can merge & publish this, the following needs to happen

- remove lodash from helper export, so plugins can't use it anymore: https://github.com/karma-runner/karma/blob/master/lib/helper.js#L171
- go through all plugins in the `karma-runner` org and make sure they do depend on their own lodash version.

wesleycho:
I just went through every repository - I believe we should be good, but you may want to do a second pass.

I also removed lodash from being exported.

PR #2690, opened 2017-05-06
fix(file-list): Stop polluting global environment with core-js

JakeChampion:
When using Karma programatically I came across issues which were due to core-js polluting the environment. It turns out that all Karma needs from core-js (whilst karma supports node 0.x versions) is `Map` and `Set`, switching to the file includes for these minimises the amount of core-js being used :-)

### DIFF
--- a/lib/file-list.js
+++ b/lib/file-list.js
@@ -7,7 +7,8 @@
 // Dependencies
 // ------------
 
-require('core-js')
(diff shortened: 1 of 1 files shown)


JakeChampion:
I signed it!

JakeChampion:
I'm not sure about the npm install failure on windows node 5, can we re-run that CI job?

PR #2691, opened 2017-05-06
feat(ci): disable testing of node versions below 4

wesleycho:
- Remove active testing of node versions below 4


wesleycho:
Whoops, pushed to the wrong remote

Issue #1397, opened 2015-05-08
In Visual Studio, crashed with errno -4082 EBUSY


maccurt:
Sometimes when I save a file in Visual Studio, karma just crashes with a -4082 EBUSY. This is currently only happening to me at work, it never happens on my computer at home. Any advice.  I pondered on if it was the virsus scan locking up the file or something. Is there a way to force Karma to contine on an error? I opened the test file in notepad and saved it and it also crashed, but it took me really saving quick over and over to get it to crash, with Visual Studio it is just doing common work.. I will check my version at home (when I get home)

Visual Studio Premium 2013
Karma version 0.12.31
node v0.12.2

PhantomJS

maccurt:
I am at a loss, so I opend the project in VS2012 and I am noticing it does not happen (or has not happened yet). I will look and see if in VS2013 I have some plug in or extension that is causing this. Perhaps some process has the file in VS2013 (resharper, web essential, some devil spawn)

maccurt:
This might shed some light on the issue: https://github.com/codereflection/Giles/issues/40

maccurt:
I closed this issue, because I believe it is a VS2013 issue

Issue #1213, opened 2014-10-14
Frequent EBUSY error on changed file


madbarron:
When I have the karma server running and I save a tracked file, it often crashes karma with:

```
ERROR [karma]: { [Error: EBUSY, open 'c:\myfile.js']
  errno: 10,
  code: 'EBUSY',
  path: 'c:\\myfile.js' }
Error: EBUSY, open 'c:\myfile.js'
```

I have set `autoWatchBatchDelay: 1000` in my karma.conf.js, but this error occurs before the delay takes place.

I am in Windows 7, using Sublime Text 3, karma 0.12.24, node 0.10.32

AbraaoAlves:
+1
I am in Windows 8.1, using Sublime Text 3, karma 0.12.31, node 0.10.33

Narretz:
I have the same issue. Maybe it was the last Sublime update?

backmask:
Currently, files are loaded as soon as a modification is detected. If a file handle is not released yet, karma crashes (or tests are aborted with 095cfb4dcf6b7520ecbc6e9bcd1d90b7f0fe79f9).

I believe the best fix would be to load modified files after the `autoWatchBatchDelay` It would also have the nice benefit to reload a file once if it is saved multiple times quickly.

madbarron:
Yeah, I expected that karma wouldn't attempt to touch the files until `autoWatchBatchDelay`. Your PR looks like a reasonable workaround.

Issue #959, opened 2014-03-16
Karma crashes on editing files in Visual Studio (due to temp files?)


DanTup:
When I run Karma and edit a js file in Visual Studio, it crashes like this (edit: seems like this is intermittent rather than consistent!):

```
ERROR [karma]: { [Error: ENOENT, stat 'M:\Coding\TestApps\AngularAppTest\Tests\tqxo1gsf.xxl']
  errno: 34,
  code: 'ENOENT',
  path: 'M:\\Coding\\TestApps\\AngularAppTest\\Tests\\tqxo1gsf.xxl' }
Error: ENOENT, stat 'M:\Coding\TestApps\AngularAppTest\Tests\tqxo1gsf.xxl'
INFO [watcher]: Changed file "M:/Coding/TestApps/AngularAppTest/Tests/SampleSpec.js".
```

It seems that when you save in Visual Studio, what happens is:
1. New (temp) file is created with the new contents
2. Old file is deleted
3. Temp file is renamed to the original filename

When I set autoWatchBatchDelay to 1000ms the error still seemed to occur, and appeared to occur immediately (not after 1s), however with it changed to 5000ms, I'm unable to repro (but still trying!), so not completely sure whether it's related to the threshold or not. My machine is very fast and I'm saving to SSDs, so I would be surprised if anything could take 1s, but Ill continue investigating!

(I can't find the code that actually uses autoWatchBatchDelay in GitHub search; so I'm struggling to dig into the code that does this)

PR #2713, opened 2017-05-17
Ywong/fix crash on read file error

ywonglanecove:
When saving a file in IDEs like Visual Studio, Atom, a dozen operations including read/write/delete/rename are performed on the file, there are nearly 50% chance karma will crash because the file is locked

There werer multiple issues raised because of this

https://github.com/karma-runner/karma/issues/1397
https://github.com/karma-runner/karma/issues/1213
https://github.com/karma-runner/karma/issues/959

The solution is retry reading file if fs.readFile fails and abort after 3 retries

### DIFF
--- a/lib/preprocessor.js
+++ b/lib/preprocessor.js
@@ -86,7 +86,7 @@ var createPreprocessor = function (config, basePath, injector) {
           fs.readFile(file.originalPath, handleFile)
           return
         } else {
-          throw err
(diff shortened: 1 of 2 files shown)


ywonglanecove:
I signed it!

ywonglanecove:
Closing because I used my work email address as committer

PR #2714, opened 2017-05-17
fix(preprocessor): retry if fs.readFile fails

ywonglanecove:
When saving a file in IDEs like Visual Studio, Atom, a dozen operations including read/write/delete/rename are performed on the file. When aggressive antivirus like McAfee is running it could make it even worse, there are nearly 50% chance karma will crash because the file is locked

There were multiple issues raised because of this

#1397
#1213
#959

The solution is retry reading file if fs.readFile fails and throw after 3 failed retries

### DIFF
--- a/lib/preprocessor.js
+++ b/lib/preprocessor.js
@@ -86,7 +86,7 @@ var createPreprocessor = function (config, basePath, injector) {
           fs.readFile(file.originalPath, handleFile)
           return
         } else {
-          done()
+          throw err
         }
       }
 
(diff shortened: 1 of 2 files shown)


lib/preprocessor.js
```diff
@@ -121,7 +130,8 @@ var createPreprocessor = function (config, basePath, injector) {
 
         nextPreprocessor(null, thisFileIsBinary ? buffer : buffer.toString())
       })
-    })
+    }
+    return fs.readFile(file.originalPath, handleFile)
```
dignifiedquire: This should call `handleFile` probably?

ywonglanecove: fs.readFile is the correct call as handleFile was the original anonymous callback at line 79

I've renamed handleFile to readFileCallback to make it clearer

dignifiedquire:
Thank you :octocat:

PR #2779, opened 2017-07-18
feat(server): Add public API to force a file refresh

pvdlg:
Adds a public API similar to `server.refreshFiles()` that allows to refresh a specific file.

This is useful for preprocessors that process files with dependencies (sass, js modules, postcss), so they can trigger an update on a file when a change is detected on a dependency.
Contrary to `server.refreshFiles()` it allows to refresh only the file that need to be processed again.


pvdlg:
The rebase is done!

PR #2791, opened 2017-07-31
[add] browser.allResults 

pizzasaurusrex:
saving tests results to the browser variable will allow easier access to unit test results within launcher plug-ins. (specifically karma-sauce-launcher)


pizzasaurusrex:
CLA should be fixed

PR #2837, opened 2017-10-10
fix(config): Call debug log methods after setting the loglevel based upon config/cli-options

JakeChampion:
Currently the debug message about where the configfile was loaded from is always shown, regardless of the user defined loglevel. I've moved those debug messages to be done after the user's defined loglevel has been set which enables the user to choose whether they see this message or not :-)

Alternative fix the problem described in https://github.com/karma-runner/karma/pull/2658

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -357,7 +357,6 @@ var CONFIG_SYNTAX_HELP = '  module.exports = function(config) {\n' +
 var parseConfig = function (configFilePath, cliOptions) {
   var configModule
   if (configFilePath) {
-    log.debug('Loading config %s', configFilePath)
 
     try {
       configModule = require(configFilePath)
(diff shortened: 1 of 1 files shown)


lib/config.js
```diff
@@ -389,12 +388,11 @@ var parseConfig = function (configFilePath, cliOptions) {
       return process.exit(1)
     }
   } else {
-    log.debug('No config file specified.')
     // if no config file path is passed, we define a dummy config module.
     configModule = function () {}
   }
 
-  var config = new Config()
+    var config = new Config()
```
dignifiedquire: this seems unrelated/wrong indentation

JakeChampion: Not quite sure how that happened, I've edited it via the Github editor.

JakeChampion:
@dignifiedquire updated pr in line with your review

dignifiedquire:
Thank you :octocat:

Issue #2824, opened 2017-09-03
Include css file without `.css` extension


josh18:
For example if I have:
```
files: [
    'https://fonts.googleapis.com/icon?family=Material+Icons'
]
```
I get a `Uncaught SyntaxError: Invalid or unexpected token` error due to it being loaded as a script. It looks like it just [matches against the file extension](https://github.com/karma-runner/karma/blob/master/lib/middleware/karma.js#L196) so I'm guessing there is currently no way around this?

PR #2846, opened 2017-10-20
feat(middleware): added manual file type option

josh18:
Adds a manual file type so that users can instruct Karma which method to use when loading a file without an extension (such as Google font files)

Example:
```
files: [
    {
        pattern: 'https://fonts.googleapis.com/icon?family=Material+Icons',
        type: 'css'
    }
]
```

Fix for https://github.com/karma-runner/karma/issues/2824

First time working karma so let me know if there are any changes I need to make.

### DIFF
--- a/docs/config/01-configuration-file.md
+++ b/docs/config/01-configuration-file.md
@@ -425,7 +425,7 @@ httpsServerOptions: {
 
 **CLI:** `--log-level debug`
 
-**Possible values:**
+**Possible Values:**
 
   * `config.LOG_DISABLE`
   * `config.LOG_ERROR`
@@ -578,7 +578,7 @@ Note: Using `'https:'` requires you to specify `httpsServerOptions`.
 
 **Description:** Module used for Karma webserver.
 
-Uses the provided module instead of node's built in `http` or `https` module. The module loaded here must exactly match the interface of node's http module. This can be useful for loading in a module like `node-http2` to allow for http2 support. 
+Uses the provided module instead of node's built in `http` or `https` module. The module loaded here must exactly match the interface of node's http module. This can be useful for loading in a module like `node-http2` to allow for http2 support.
 
 Note: if you're using this to enable `http2` you must also set the `protocol` to `https:` and specify certificates as http2 can only run of https.
 
--- a/docs/config/01-configuration-file.md
+++ b/docs/config/01-configuration-file.md
@@ -425,7 +425,7 @@ httpsServerOptions: {
 
 **CLI:** `--log-level debug`
 
-**Possible Values:**
+**Possible values:**
 
   * `config.LOG_DISABLE`
   * `config.LOG_ERROR`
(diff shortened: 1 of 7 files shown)


josh18:
I signed it!

josh18:
1. Sure thing
2. As this is the default is this necessary? Or are you just thinking of adding it to the validation in (3)?
3. Will do, what would I use to log it? Just something like `log.warn()`?
Another thing I was just thinking of is if it should be `fileExt === '.css' || fileType === 'css'` (can't override file extension) or `(fileExt === '.css' && !fileType) || fileType === 'css'` (fileType always overrides extension)

lib/middleware/karma.js
```diff
@@ -193,12 +194,12 @@ var createKarmaMiddleware = function (
...
+            if (fileExt === '.css' || fileType === 'css') {
               scriptTags.push(util.format(LINK_TAG_CSS, filePath))
               continue
             }
 
-            if (fileExt === '.html') {
+            if (fileExt === '.html' || fileType === 'html') {
               scriptTags.push(util.format(LINK_TAG_HTML, filePath))
               continue
             }
```
appsforartists: On line 208, `scriptType` should be `fileType || SCRIPT_TYPE[fileExt] || 'text/javascript'`.  I think that would let us strip the code out of #2834 and just merge its tests.

appsforartists: GitHub sadly wouldn't let me comment on the correct line.

josh18:
Made some changes, let me know what you think.

- File type overrides extension
- Added script file types (including esModule)
- Added file type validation
- Added file type documentation
  - Renamed one instance of `Possible values` -> `Possible Values` for consistency
  - Removed extra white space (done automatically by my editor, not sure if you want to keep this change)

lib/middleware/karma.js
```diff
@@ -32,9 +33,17 @@ var CROSSORIGIN_ATTRIBUTE = 'crossorigin="anonymous"'
...
+  'js': 'text/javascript',
+  'dart': 'application/dart',
+  'esModule': 'module'
 }
+var FILE_TYPES = [
+  'css',
+  'html',
+  'js',
+  'dart',
+  'esModule'
```
appsforartists: This is total bikeshedding, but this could be more simply called `module`.  It removes the only camel-cased option, and makes the API more consistent with the browser: `type: 'module'`.

docs/config/01-configuration-file.md
```diff
@@ -425,7 +425,7 @@ httpsServerOptions: {
 
 **CLI:** `--log-level debug`
 
-**Possible values:**
+**Possible Values:**
```
appsforartists: Pull requests usually follow the mantra "one idea -> one PR".  This correction should probably have been a separate PR.

josh18: Makes sense, sorry about that. Reverted now. Should I squash all my commits?

lib/middleware/karma.js
josh18: That's fine I was just copying from the previous PR but I think `'module'` is better. Fixed now

dignifiedquire:
Thank you :octocat:

PR #2855, opened 2017-10-25
chore: update @mgol's name in .mailmap

mgol:
In this way my past contributions are mapped correctly.


mgol:
Not sure what's up with the `VALIDATE_COMMIT_MESSAGE` failure (it looks like it cannot find something in Git) but the Node 8 failure seems to be caused by the inclusion of a new built-in experimental `http2` module in Node 8.8.0 that is used instead of the npm `http2` package.

mgol:
I submitted #2857 for the Node 8.8.0 http2 issue.

PR #2857, opened 2017-10-25
test(web-server): fix http2 tests

mgol:
Node 8.8.0 includes a new experimental built-in http2 module which trumps
the external one from npm, breaking Karma tests. This commit switches the
require back to using the npm package.


dignifiedquire:
Thanks :octocat:

PR #2876, opened 2017-11-14
feat: trim whitespace in browser name

dherges:
Solves #2875


EzraBrooks:
Seems like a pretty harmless change, but it looks like a Mocha test timed out on Appveyor. Doesn't seem related to the change, either. Might be worth triggering a retry?

PR #2886, opened 2017-11-22
refactor: run build on post-install instead of pre-publish

### LINKED ISSUE #2884: Karma should compile its static resources on the client if they're not present
### Expected behaviour

The current 1.7.1 release on NPM does not work for my project. It _entirely_ prevents us from running unit tests, as described in #2798. I was hoping I could just pull the repository as a git URL in NPM until the issue was fixed in a stable release.

### Actual behaviour

However, when pulled from GitHub, the project is unbuilt and therefore unusable without running the build scripts. NPM has a post-install step for this reason. It is not sustainable for me to tell every member of my team to run your build scripts just because we have to workaround a fault in the current stable version.

### Environment Details

- Karma version (output of `karma --version`): Irrelevant
- Relevant part of your `karma.config.js` file: Irrelevant

1. Pull the project from a git URL instead of from an npm registry
2. Try to run Karma
3. 404 on `karma.js` because it's not built.

EzraBrooks:
Fixes #2884.

Switching the bundling to happen on the client-side makes Karma development more flexible. For instance, on projects at work we pull npm packages based on git tags, since we don't have enough to warrant running a private npm registry. If we were to fork Karma, it wouldn't work at all due to the bundling being done on pre-publish.

Additionally, it allows developers to easily pull in pre-release versions of Karma to test locally, since they won't have to manually run the build script when installing the module into a test package.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -5,7 +5,6 @@ tmp
 test
 tasks
 docs
-client
 logo
 integration-tests
 
@@ -17,6 +16,8 @@ Karma.sublime-*
 
 static/karma.src.js
 static/karma.wrapper
+static/context.js
+static/karma.js
 test-results.xml
 thesis.pdf
 mocha-watch.sh
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -144,7 +144,6 @@ module.exports = function (grunt) {
     grunt.task.run([
       'npm-contributors',
       'bump:' + (type || 'patch') + ':bump-only',
-      'build',
       'conventionalChangelog',
       'bump-commit',
       'conventionalGithubReleaser',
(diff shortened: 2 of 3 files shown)


EzraBrooks:
I signed it.

EzraBrooks:
Worth noting that I think maybe a more elegant solution would be to compile the client-side resources when Karma spawns the web server for the first time, but this was the easier and quicker way to get my project working again.

dignifiedquire:
In the past postinstall steps have been very problematic with npm unfortunately, so I always try and avoid them. Not being able to use `master` out of the box is not great I agree. Doing the browserify run on first start and then caching the results would be the best way to go I think.

EzraBrooks:
Okay, I’ll close this PR and work on solving it that way instead.

PR #2890, opened 2017-11-25
Document port collision scenario

### LINKED ISSUE #2862: Document port collision scenario
### Expected behaviour

Ability to let karma select a random server port

### Actual behaviour

The server port have to be fixed before deployment.

### Environment Details

- Karma version (output of `karma --version`): 1.7.1

### Why

In a CI server, multiple karma instance could be running concurrently.
There is no way to predict which port will be available.
It would be nice to let karma automatically select a random available port.

### Absence of workaround

I wanted to implement this behaviour using https://github.com/sindresorhus/get-port. 
But since the library returns a promise of port, it seems that I have no way to pass it inside my synchronous `karma.conf.js`.

reda-alaoui:
Fix #2862

### DIFF
--- a/docs/config/01-configuration-file.md
+++ b/docs/config/01-configuration-file.md
@@ -522,6 +522,8 @@ See [plugins] for more information.
 
 **Description:** The port where the web server will be listening.
 
+If the defined port is already in use, karma will automatically increase its value in steps of 1 until a free port is found.
+
 
 ## processKillTimeout
 **Type:** Number


reda-alaoui:
About the CLA issue: I confirm

EzraBrooks:
Looks like the CLA bot is frozen because I squashed the commits after it already complained about mixed user contributions. @dignifiedquire can you merge this?

PR #2916, opened 2018-01-09
Update browserify

benitogf:
image

I'm getting these two warnings, just checking if updating browserify would break anything


benitogf:
the warnings above are not related to karma or browserify, sorry 😃 closing this

PR #2918, opened 2018-01-12
fix(middleware): add file type to absolute urls

josh18:
Currently the file type option does not work with absolute urls, this PR fixes that issue.


josh18:
Updated a test to include absolute urls

dignifiedquire:
Thank you :octocat:

Issue #2930, opened 2018-02-11
Karma 2 no longer correctly reports the original error stack when using source maps


doberkofler:
### Expected behavior
When reporting an error and source maps are available the error stack should point to the original source line and column. The mocha-reporter shows this by using a format like in the following example:
_(webpack:///test/unittest/frontend/ods/convert--test.js:43:9 <- temp/karma_unittests/convert--test.js:712:10)_
```
at Context.<anonymous> (webpack:///test/unittest/frontend/ods/convert--test.js:43:9 <- temp/karma_unittests/convert--test.js:712:10)
```

### Actual behavior
Up to Karma 1.7 this worked as expected but starting with Karma 2.0 the [commit](https://github.com/karma-runner/karma/commit/34dc7d3) made a change in the definition of `URL_REGEXP` that causes the reporter to no longer recognize the lines in the error stack.
In the file `lib/reporter.js` when building `URL_REGEXP` the `urlRoot + '\\/?'` is now added before the `(base/|absolute)` and because `urlRoot` is always automatically post fixed with `/` the resulting regular expression now expects the url always to start with `/base` causing the matching to fail and to no longer correctly report the original file, line and column of the error stack.
```
at Context.<anonymous> (base/temp/karma_unittests/convert--test.js?6c85b6986d1a9cc153d2877ce8637a38a6b666ad:712:10)
```

### Environment Details

- Karma version (output of `karma --version`): 2.0.0
- Relevant part of your `karma.config.js` file
```
// Karma configuration

module.exports = function(config) {
  config.set({
    // base path that will be used to resolve all patterns (eg. files, exclude)
    basePath: '',

	// frameworks to use
    // available frameworks: https://npmjs.org/browse/keyword/karma-adapter
    frameworks: ['mocha', 'chai'],

    // list of files / patterns to load in the browser
    files: [
      'dist/**/*.js'
    ],

    // list of files to exclude
    exclude: [
    ],

    // preprocess matching files before serving them to the browser
    // available preprocessors: https://npmjs.org/browse/keyword/karma-preprocessor
    preprocessors: {
		'dist/**/*.js': ['sourcemap']
    },

    // test results reporter to use
    // possible values: 'dots', 'progress'
    // available reporters: https://npmjs.org/browse/keyword/karma-reporter
    reporters: ['mocha'],

    // reporter options
    mochaReporter: {
      showDiff: true
  	},

    // web server port
    port: 9876,

    // enable / disable colors in the output (reporters and logs)
[...]

PR #2931, opened 2018-02-12
fix(reporter): Fix for issue #2930 causing error stack not to be parsed correctly

doberkofler:
Fixed a problem when config.urlRoot is set to '/' causing the error stack not to be parsed correctly and therefore source maps are not used.

Issue: https://github.com/karma-runner/karma/issues/2930


doberkofler:
CLA: I signed it!

PR #2939, opened 2018-02-24
fix(lib): switch to sync write

BridgeAR:
The async write operation has no callback and errors are not
detected that way. Switch to the sync version to circumvent that
problem.

This will otherwise cause issues from Node.js 10 on. See https://github.com/nodejs/node/pull/18668.


dignifiedquire:
thank you

PR #2947, opened 2018-03-02
Avoid using deprecated Buffer constructor

ChALkeR:
`safe-buffer` is already used and provides `Buffer.alloc` polyfill, so just use `Buffer.alloc` directly to avoid hitting deprecated API.

Refs:
https://nodejs.org/api/deprecations.html#deprecations_dep0005_buffer_constructor

This was overlooked in 3d94b8cf18c695104ca195334dc75ff054c74eec (#2484).

Tracking: https://github.com/nodejs/node/issues/19079


ChALkeR:
I signed it!

PR #2954, opened 2018-03-13
fix: Nicer "disconnect" - no more "Disconnectedundefined"

apepper:
The method "disconnect" sometimes is called without arguments. The emited message should not look broken then.

Previously it looked like this:

disconnected


apepper:
I signed it!

dignifiedquire:
Thanks :octocat:

PR #2985, opened 2018-05-01
refactor(formatters): refactor formatters.js

lusarz:
Refactor `lib/init/formatters` to use ES6

### DIFF
--- a/config.tpl.coffee
+++ b/config.tpl.coffee
@@ -25,7 +25,8 @@ module.exports = (config) ->
 
     # preprocess matching files before serving them to the browser
     # available preprocessors: https://npmjs.org/browse/keyword/karma-preprocessor
-    preprocessors: %PREPROCESSORS%
+    preprocessors: {%PREPROCESSORS%
+    }
 
 
     # test results reporter to use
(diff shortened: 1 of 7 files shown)


config.tpl.coffee
```diff
@@ -25,7 +25,8 @@ module.exports = (config) ->
 
     # preprocess matching files before serving them to the browser
     # available preprocessors: https://npmjs.org/browse/keyword/karma-preprocessor
-    preprocessors: %PREPROCESSORS%
+    preprocessors: {%PREPROCESSORS%
+    }
```
dignifiedquire: why this change?

lib/utils/file-utils.js
```diff
@@ -0,0 +1,17 @@
+const fs = require('graceful-fs')
```
dignifiedquire: please add `'use strict'`

config.tpl.coffee
lusarz: to be more consistent with way `files`, `exclude` are built.

lib/utils/file-utils.js
lusarz: Done

PR #3007, opened 2018-05-17
refactor: migrate init to ES2015

devoto13:


devoto13:
Ops, this seems to be a dupe of #2998. Closing.

PR #3013, opened 2018-05-24
config: setup 'arrow-parens' eslint rule

lusarz:
As discussed with @johnjbarton [here](https://github.com/karma-runner/karma/pull/3012#discussion_r190129318) this rule will force us to use parens in arrow functions.


lusarz:
CI servers fails when `eslint` fails. Current build pass, so we don't need to fix anything. I fixed 4 cases within this pr.

PR #3015, opened 2018-05-24
refactor: migrate Url to ES2015

devoto13:


lib/url.js
```diff
@@ -1,17 +1,16 @@
-// Url
-// ===
-//
-// Url object used for tracking files in `file-list.js`
+/**
```
lusarz: Please add `'use strict'` here, it would allow to pass tests on Node 4. 

Personally I think we can remove this comment.

devoto13: Good catch! Fixed.

PR #3017, opened 2018-05-24
Prepare BaseReporter for conversion to class

devoto13:
Converted to ES2015 syntax and re-arranged code a bit, so conversation to real class won't produce hard to review diff.

In tests import BaseReporter directly instead of using `mocks`. This was unnecessary complexity.

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -1,23 +1,43 @@
-var util = require('util')
+'use strict'
 
-var constants = require('../constants')
-var helper = require('../helper')
+const util = require('util')
+
+const constants = require('../constants')
+const helper = require('../helper')
 
 var BaseReporter = function (formatError, reportSlow, useColors, browserConsoleLogOptions, adapter) {
   this.adapters = [adapter || process.stdout.write.bind(process.stdout)]
 
-  this.onRunStart = function () {
+  this.USE_COLORS = false
+  this.EXCLUSIVELY_USE_COLORS = undefined
+  this.LOG_SINGLE_BROWSER = '%s: %s\n'
+  this.LOG_MULTI_BROWSER = '%s %s: %s\n'
+
+  this.SPEC_FAILURE = '%s %s FAILED' + '\n'
+  this.SPEC_SLOW = '%s SLOW %s: %s\n'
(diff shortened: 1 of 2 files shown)


lib/reporters/base.js
```diff
@@ -1,23 +1,43 @@
...
+const helper = require('../helper')
+
+const BaseReporter = function (formatError, reportSlow, useColors, browserConsoleLogOptions, adapter) {
   this.adapters = [adapter || process.stdout.write.bind(process.stdout)]
 
-  this.onRunStart = function () {
+  this.USE_COLORS = false
+  this.EXCLUSIVELY_USE_COLORS = undefined
+  this.LOG_SINGLE_BROWSER = '%s: %s\n'
+  this.LOG_MULTI_BROWSER = '%s %s: %s\n'
```
lusarz: Seems like `LOG_SINGLE_BROWSER` and `LOG_MULTI_BROWSER` variables are redundant. I think:

    this.writeCommonMsg(util.format(this.LOG_SINGLE_BROWSER, type, log)) 
can be replaced with:

    this.writeCommonMsg(`${type}: ${log}\n`) 

This note may also refer to other variables, what do you think ?

devoto13: Yeah, these variables are base values, which can be redefined by the concrete reporters, so they should be kept. Also as other reporters in a wild may rely on these values removing them may introduce a breaking change.

lib/reporters/base.js
```diff
@@ -1,23 +1,43 @@
...
   }
 
-  this.renderBrowser = function (browser) {
-    var results = browser.lastResult
-    var totalExecuted = results.success + results.failed
-    var msg = util.format('%s: Executed %d of %d', browser, totalExecuted, results.total)
+  this.renderBrowser = (browser) => {
+    const results = browser.lastResult
+    const totalExecuted = results.success + results.failed
+    let msg = util.format('%s: Executed %d of %d', browser, totalExecuted, results.total)
```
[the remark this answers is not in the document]
devoto13: We can remove it from some places, but not completely. Again this is used in  `write` method, which can be used by third-party reporters.

PR #3029, opened 2018-06-05
refactor(server): refactor bundleResource in lib/server.js

lusarz:
Pull request refactor code around `static/context.js` and `static/karma.js` bundling. Main changes:
- move code responsible for bundle resources with `browserify` into `lib/utils/bundle-utils`
- move responsibility to check if bundled files exists into `BundleUtils.bundleResourceIfNotExist` method
- overall refactoring
- adding unit tests for created methods

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -8,6 +8,7 @@ const spawn = require('child_process').spawn
 const tmp = require('tmp')
 const fs = require('fs')
 const path = require('path')
+const BundleUtils = require('./utils/bundle-utils')
 const root = global || window || this
 
 const cfg = require('./config')
@@ -30,20 +31,6 @@ const Browser = require('./browser')
 const BrowserCollection = require('./browser_collection')
 const EmitterWrapper = require('./emitter_wrapper')
 const processWrapper = new EmitterWrapper(process)
-
-const karmaJsPath = path.join(__dirname, '/../static/karma.js')
-const contextJsPath = path.join(__dirname, '/../static/context.js')
-
-function bundleResource (inPath, outPath) {
-  const browserify = require('browserify')
-  return new Promise((resolve, reject) => {
-    browserify(inPath)
-      .bundle()
-      .pipe(fs.createWriteStream(outPath))
-      .once('finish', () => resolve())
-      .once('error', (e) => reject(e))
-  })
-}
 
 function createSocketIoServer (webServer, executor, config) {
   const server = new SocketIO(webServer, {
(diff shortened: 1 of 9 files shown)


lib/server.js
```diff
@@ -31,10 +31,10 @@ const BrowserCollection = require('./browser_collection')
 const EmitterWrapper = require('./emitter_wrapper')
 const processWrapper = new EmitterWrapper(process)
 
-const karmaJsPath = path.join(__dirname, '/../static/karma.js')
-const contextJsPath = path.join(__dirname, '/../static/context.js')
-
 function bundleResource (inPath, outPath) {
+  if (fs.existsSync(outPath)) {
```
lusarz: Moved responsibility to check if `outPath` file already exists to bundleResource method.

lib/server.js
```diff
@@ -152,39 +152,32 @@ class Server extends KarmaEventEmitter {
...
-            process.exitCode = 1
-            process.kill(process.pid, 'SIGINT')
-          }
-        })
-      }
-
-      if (fs.existsSync(karmaJsPath) && fs.existsSync(contextJsPath)) {
-        startWebServer()
-      } else {
-        this.log.info('Front-end scripts not present. Compiling...')
```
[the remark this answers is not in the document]
johnjbarton: So this test does not work?
https://github.com/karma-runner/karma/blob/31d9a08892f51308efbe7c467c379b3f70f1e416/test/unit/server.spec.js#L113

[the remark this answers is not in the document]
johnjbarton: Now that I look more carefully the `else` case test is not correct because `fs` is not injected and the test runs in the same filesystem location every time.  If the file system has the static files from a previous run or from a use of karma the test will pass even if bundling code does not work.

lib/server.js
johnjbarton: Please put this comment in the git commit description.

lib/server.js
```diff
@@ -152,39 +152,32 @@ class Server extends KarmaEventEmitter {
...
-          bundleResource(path.join(__dirname, '/../client/main.js'), karmaJsPath),
-          bundleResource(path.join(__dirname, '/../context/main.js'), contextJsPath)
-        ])
-          .then(startWebServer)
-          .catch((error) => {
-            this.log.error('Front-end script compile failed with error: ' + error)
-            process.exitCode = 1
-            process.kill(process.pid, 'SIGINT')
+      return Promise.all([
+        bundleResource(path.join(__dirname, '/../client/main.js'), path.join(__dirname, '/../static/karma.js')),
```
johnjbarton: I think the code would be clearer if you create  function to compute the paths:
```js
  bundleResource(['/../client/main.js', '/../static/karma.js'].map(absPath));
```
If you also pass the `fs`:
```js
  bundleResource(fs, ['/../client/main.js', '/../static/karma.js'].map(absPath));
```
then we can unit test the `bundleResource()` properly.

lib/server.js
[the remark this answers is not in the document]
lusarz: I've updated it a bit, now everything should be tested.

lib/server.js
lusarz: Maybe a bit other way, but done :)

lib/server.js
lusarz: You mean pull request description ?

PR #3050, opened 2018-06-16
refactor(preprocessor): update lib/preprocessor.js to ES6

lusarz:


lusarz:
@johnjbarton good points about specs, I'll add tests for these files.

PR #3054, opened 2018-06-18
feat(preprocessor): Allow preprocessor to handle binary files

kyliau:
Karma currently detects whether a file is binary, and does not allow binary files to be passed on to preprocessors. This PR removes this restriction, and allows binary files to be handled by preprocessors.

See https://github.com/karma-runner/karma/issues/3046

### DIFF
--- a/lib/preprocessor.js
+++ b/lib/preprocessor.js
@@ -98,12 +98,7 @@ function createPreprocessor (config, basePath, injector) {
         var preprocessorNames = []
         for (var i = 0; i < patterns.length; i++) {
           if (mm(file.originalPath, patterns[i], {dot: true})) {
-            if (thisFileIsBinary) {
-              log.warn('Ignoring preprocessing (%s) %s because it is a binary file.',
-                config[patterns[i]].join(', '), file.originalPath)
-            } else {
-              preprocessorNames = combineLists(preprocessorNames, config[patterns[i]])
-            }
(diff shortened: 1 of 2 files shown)


lib/preprocessor.js
```diff
@@ -125,7 +120,12 @@ function createPreprocessor (config, basePath, injector) {
           }
 
           instances[name] = p
-          preprocessors.push(p)
+          if (!thisFileIsBinary || p.handleBinaryFiles) {
```
lusarz: Do we really need `handleBinaryFiles` flag and this whole if statement ? Maybe we can just allow to preprocess binary files, and if something fails, preprocessor will throw error.

I guess in more than 90% there will be no binary files that matches pattern and should be preprocessed (I may miss something).

Just thinking out loud, and I'm curious others opinion about that.

kyliau: This is just for safe backwards compatibility.
Currently if there is a pattern `foo/*` that matches binary files, it is ignored. Without this flag, the 10% projects that do include binary files in their pattern will break.

lib/preprocessor.js
```diff
@@ -125,7 +120,12 @@ function createPreprocessor (config, basePath, injector) {
...
+            preprocessors.push(p)
+          } else {
+            log.warn('Ignoring preprocessing (%s) %s because it is a binary file.',
```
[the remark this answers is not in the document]
kyliau: done

PR #3152, opened 2018-10-02
refactor(server): replace var with const/let

lusarz:


johnjbarton:
Looks like the cumcumber tests are point to some issue?

lusarz:
Yes, I'll investigate more why. 

Unfortunately now when I run tests on my computer, even on `master` branch I'm getting:
error

PR #3157, opened 2018-10-03
refactor(launcher): use this directly instead assign to self variable in base launcher

lusarz:
This pull request is the first step that should help us to transform `BaseLauncher` to be ES6 class

### DIFF
--- a/lib/launchers/base.js
+++ b/lib/launchers/base.js
@@ -32,7 +32,6 @@ function BaseLauncher (id, emitter) {
   this.state = null
   this.error = null
 
-  const self = this
   let killingPromise
   let previousUrl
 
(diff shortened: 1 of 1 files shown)


lib/launchers/base.js
```diff
@@ -41,7 +40,7 @@ function BaseLauncher (id, emitter) {
 
     this.error = null
     this.state = BEING_CAPTURED
-    this.emit('start', url + '?id=' + this.id + (helper.isDefined(self.displayName) ? '&displayName=' + encodeURIComponent(self.displayName) : ''))
+    this.emit('start', url + '?id=' + this.id + (helper.isDefined(this.displayName) ? '&displayName=' + encodeURIComponent(this.displayName) : ''))
   }
 
   this.kill = function () {
```
devoto13: I think you should also change this.start and this.kill to be arrow functions to be on the safe side.

PR #3187, opened 2018-10-17
refactor(lib): use assert library instead of throwing error directly

lusarz:
NodeJS includes [assertion library](https://nodejs.org/api/assert.html) natively, and it's syntax seems to be more elegant that throwing errors directly.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const path = require('path')
+const assert = require('assert')
 const optimist = require('optimist')
 const fs = require('graceful-fs')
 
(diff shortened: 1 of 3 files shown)


lusarz:
I see message like:
assert
previously it was:
type

I think we can go one step further - wrap assert and create something like:

    assertArrayWhenDefined(config.browsers, 'Invalid configuration: browsers option must be an array')

etc. 

@johnjbarton What do you think ?

johnjbarton:
I think your change is good as is.

lusarz:
Ok.
Maybe a bit offtopic - seems like we have two different ways of processing fatal errors:
- here we throw error out:
https://github.com/karma-runner/karma/blob/385a0f35121b2e5b09503490eb43a9f4f87968c2/lib/config.js#L200
- here we log error and do process.exit : 
https://github.com/karma-runner/karma/blob/385a0f35121b2e5b09503490eb43a9f4f87968c2/lib/config.js#L381-L382

Would be nice to make it more consistent one day :)

PR #3223, opened 2018-12-01
fix: add missing dep flatted (fixes #3222)

### LINKED ISSUE #3222: Cannot find module 'flatted/cjs'
seems it should be in deps,  than in dev deps.

it was introduced in https://github.com/karma-runner/karma/commit/fb05fb134fb67f8879711054b6e0da0febbbdc7b

Aladdin-ADD:
the tests failed on node 6, but cannot be related to this change.


Aladdin-ADD:
I signed it!

Issue #3226, opened 2018-12-03
karma 3.1.3 is not working with single entry file


Havunen:
### Expected behaviour

Karma 3.1.3 should re-process files when there is single entry point and it uses karma-webpack integration

### Actual behaviour

Karma 3.1.3 does not re-process the file and no changes are picked up
Karma 3.1.1 works as expected

segrey:
@Havunen Could you please provider steps to reproduce? A small example project would be great. Also, how do you run your karma tests?

garyvh2:
@segrey here's a quick example https://github.com/garyvh2/karma-reload-problem
Just do the following:
```
git clone https://github.com/garyvh2/karma-reload-problem
npm install
npm test
```
Tests will be runned on Chrome Headless, you will noticed 2 tests being run successfully.

Now, go into `deepCompare.spec.js` and change any of the test to fail, I.E change in `line 10`, false into true, so you can break the inequality test.

You will notice that the tests are automatically run again, but both still successful, now, stop the tests (`ctrl + c`) and run them again without changing anything, now 1 tests is failing and the other one is successful.

If you try this with `Karma 3.1.1` the behavior works as expected, so whenever you perform any change on the tests they are effectively run again, without you having to restart the test server.

Tell me if you need further information.

segrey:
@garyvh2 Thanks for the repro!

matthieu-foucault:
It's been reported in `karma-webpack` too, here's another repro if you're interested: https://github.com/webpack-contrib/karma-webpack/issues/100#issuecomment-444216116

Single-entry setups are quite with `karma-webpack`

PR #3230, opened 2018-12-06
fix(file-list): revert "do not preprocess up-to-date files" (#3226)

segrey:
This reverts commit 5334d1a8
karma-webpack preprocessor updates bundle when needed, but karma's fileList consists of a single entry file which is usually unmodified. In this case, fileList.refresh() should always call preprocessor for the single entry file, avoiding preprocessing of unmodified files is done on karma-webpack side.

### DIFF
--- a/lib/file-list.js
+++ b/lib/file-list.js
@@ -102,12 +102,6 @@ class FileList {
           return Promise.resolve(file)
         }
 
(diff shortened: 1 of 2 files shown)


segrey:
@johnjbarton Thanks for the quick turnaround. Sorry for the regression.

PR #3259, opened 2019-02-02
Debug Child Processes exit signal

webmaster128:
The Child Processes Event: 'exit' has two parameters: `code` and `signal` (https://nodejs.org/api/child_process.html#child_process_event_exit)

> One of the two will always be non-null.

This helps debugging cases where code is null.


webmaster128:
I signed it!

PR #3274, opened 2019-02-20
fix(config): remove ts-node register() call to allow custom configuration

mlegenhausen:
Calling `require('ts-node').register()` to determine if `typescript` is available prevents the custom configuration of the `register` call described in the documentation. I removed the unnecessary `register` call and fixed the missing `module.exports` assignment in the documentation.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -27,7 +27,7 @@ try {
 } catch (e) {}
 
 try {
-  require('ts-node').register()
+  require('ts-node')
   TYPE_SCRIPT_AVAILABLE = true
 } catch (e) {}
 
--- a/docs/config/01-configuration-file.md
+++ b/docs/config/01-configuration-file.md
@@ -64,7 +64,7 @@ require('ts-node').register({
     module: 'commonjs' 
   } 
 });
-require('./karma.conf.ts');
+module.exports = require('./karma.conf.ts');
 ```
 
 ## File Patterns


lib/config.js
```diff
@@ -27,7 +27,7 @@ try {
 } catch (e) {}
 
 try {
-  require('ts-node').register()
+  require('ts-node')
```
johnjbarton: I gather you are saying: we don't need to call register because all we have no options?

trusktr: The fact that Karma does this is highly bothersome: https://github.com/karma-runner/karma/issues/3329

Use of `ts-node` is not a bad idea, but it should be optional, with some way to enable it.

trusktr:
By the way, all of my build and test processes live in an NPM package, so in a new project, I install the package, which gives me out-of-the-box functionality for building with Gulp/Webpack/Babel and testing with Karma/Electron. 

So, none of my projects have any of this build configuration in their repos, only one central package contains these things.

I think you can begin to imagine how the inflexible ts-node `register()` call creates a problem here.

The only easy workaround that I can think of is forking karma to disable the `register()` call (like in that PR).

trusktr:
To get a sense of how I'm using Karma, this is the shared package I use with my projects: https://github.com/trusktr/builder-js-package

In there, you'll see a `config` folder which contains configs for various tools.

It relies on [Builder](https://github.com/FormidableLabs/builder), a tool designed to make it easy to share package scripts across projects.

So, because Karma and it's configuration are located inside `node_modules`, it complicates things (for example, the `karma.config.js` can't make an accurate guess as to which `tsconfig.json` file a project may be using, in order to extend it and disable `allowJs` and `checkJs`.

lib/config.js
mlegenhausen: @trusktr thanks for the example thats exactly the problem I encountered too. You can only call `register()` once in your whole project after that it is not possible to change the `options` given to the `register` call. The `register` call in your `kamra.conf.ts` is simply ignored cause `karma` already called it here. The only way to inject some other configuration is via the `tsconfig.json` which needs to be the top level `tsconfig.json` file. This is not suitable for all projects.

This will be a breaking change cause, if we remove the `register` call we need a `karma.conf.js` that calls `register` for us. Another solution could be to allow the injection of a `tsconfig.json` via a command line arg.

For me I was able to find a `tsconfig.json` configuration that fits `karma` and the rest of the project. But the way to this solution was very cumbersome.

johnjbarton: Please add the breaking change info

https://www.conventionalcommits.org/en/v1.0.0/

and the doc change.

muuvmuuv:
Can we please have this merged. It is blocking a PR in Angular builders that allows custom webpack configuration in Angular.

> https://github.com/just-jeb/angular-builders/pull/824

mlegenhausen:
It seems the problem was fixed via https://github.com/karma-runner/karma/commit/474f4e1caff469cce87f19a11d9179e4e05552f9. Now it should be possible to define a JS configuration that can call ts-node's register function without interfering with the karma register call.

I am currently unable to confirm if this solved the described problem cause I am not working anymore on any karma dependent project.

PR #3280, opened 2019-02-27
fix: remove vulnerable dev dependencies

SteinRobert:
Remove `grunt-cucumberjs` as a dependency.
Move from qunitjs to qunit (same framework new name)

Not sure - it looked to me like `grunt-cucumberjs` is not used.
`qunitjs` is now published as `qunit`. The dependency would be updated to the next working version, not the latest. Is `qunit` actually used in the code? Was is probably replaced by `karma-qunit`?


SteinRobert:
Seems like grunt-cucumberjs is still used - however there is no update available for the vulnerability.

PR #3309, opened 2019-05-15
Cleanup dependencies

devoto13:
- Removed devDependencies, which are not used by the codebase.
- Updated most of the remaining devDependencies to latest versions.


test/e2e/reconnecting.feature
```diff
@@ -12,6 +12,11 @@ Feature: Passing Options
         'karma-jasmine',
         'karma-chrome-launcher'
       ];
+      client = {
+        jasmine: {
+          random: false
```
devoto13: So Jasmine 3 executes tests in random order by default and since this particular spec relies on the tests being executed sequentially I have disabled the random execution feature.

PR #3312, opened 2019-05-18
chore: update eslint packages to latest and fix complaints

devoto13:
Two new rules are:
- require whitespaces inside of the object literals
- no usage of deprecated Node APIs

As for the second rule fixed `assert` module to use *strict* methods, but silenced warnings on deprecated `url` module API, because the new version (WHATWG URL class) is not a trivial replacement. Deprecated API will stay for at least couple of Node versions, so it is not urgent to update it.


devoto13:
Rebased on top of https://github.com/karma-runner/karma/pull/3313. This PR should be merged after https://github.com/karma-runner/karma/pull/3313.

PR #3313, opened 2019-05-18
chore: revert back to Mocha 4

devoto13:
Mocha 5+ dropped support for IE 8 and IE 9. This leads to [failing
BrowserStack tests](https://travis-ci.org/karma-runner/karma/jobs/534037781#L900), so reverting back to version 4 until Karma decides
to drop support for the mentioned browsers.


devoto13:
The failed tests look like flakes... Will re-trigger the build.

PR #3314, opened 2019-05-21
chore(test): fix flaky test cases

devoto13:
See separate commits for details on fixed issues.


devoto13:
@johnjbarton This is ready for review. Please check separate commits to make review easier. It doesn't really make sense to split commits into separate PRs as they target same issue and have a combined effect.

Hopefully this will make tests more stable and unblock my other PRs.

PR #3315, opened 2019-05-27
docs(travis): Correct the docs to also show how to do it on Xenial

peternewman:


peternewman:
Replaced by #3316 as the commit message is wrong.

PR #3316, opened 2019-05-27
docs(travis): Correct the docs to also show how to do it on Xenial

peternewman:


peternewman:
Take two, replacing #3315 with the commit message as specified too.

peternewman:
Thanks for merging. Does this only get onto the website when you cut a new release then? The site still has the old info currently (for both 4.0 and 3.0):
https://karma-runner.github.io/4.0/plus/travis.html

devoto13:
@peternewman Deploying documentation is a manual process currently. But docs in this repo are the source of truth, so it will get to the website at some point.

peternewman:
Great thanks @devoto13 , I assumed it would be GitHub pages autodeploy given the URL, or is it more involved than that? Travis can do some clever stuff now if that would work (aside from the direct deployment via GitHub):
https://docs.travis-ci.com/user/deployment/pages/

devoto13:
Rendered docs live in another repo - https://github.com/karma-runner/karma-runner.github.com, so somebody (maybe you :smile:) will need to go there, run `sync-docs.sh` and submit a PR. Once merged it will be deployed automatically.

PR #3325, opened 2019-06-05
fix(reporter): format stack with 1-based column

### LINKED ISSUE #3324: Incorrect column in stack trace with original file
### Expected behaviour

Stack trace contains locations with 1-based line and 1-based column.

### Actual behaviour

Stack trace contains locations with 1-based line and 0-based column.

### Environment Details

- Karma version (output of `karma --version`): 4.1.0
- Relevant part of your `karma.config.js` file: see the project

1. Clone https://github.com/segrey/karma-webpack-example
2. Run tests with `npm test`
3. Observe line
```
 at UserContext.<anonymous> (webpack:///test/specs/utils.spec.js:3:4 <- index.js:72:5)
```

The location in the original file (`utils.spec.js:3:4`) points to the correct 1-based line, but has wrong 1-based column (in fact it's 0-based column). Since lines and columns in stack traces are expected to be both 1-based, it should be `utils.spec.js:3:5`.

Please note that the location in the transpiled file (`index.js:72:5`) has correct 1-based column.

segrey:
Columns in original stack are 1-based, but
SourceMapConsumer.prototype.originalPositionFor(generatedPosition)
accepts 0-based column and returns 0-based column too.
This change converts columns from 1-based to 0-based forth and back.

Closes #3324

### DIFF
--- a/lib/reporter.js
+++ b/lib/reporter.js
@@ -60,11 +60,13 @@ function createErrorFormatter (config, emitter, SourceMapConsumer) {
         const bias = column ? SourceMapConsumer.GREATEST_LOWER_BOUND : SourceMapConsumer.LEAST_UPPER_BOUND
 
         try {
-          const original = getSourceMapConsumer(file.sourceMap).originalPositionFor({ line, column: (column || 0), bias })
+          const zeroBasedColumn = Math.max(0, (column || 1) - 1)
+          const original = getSourceMapConsumer(file.sourceMap).originalPositionFor({ line, column: zeroBasedColumn, bias })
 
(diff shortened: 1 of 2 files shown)


segrey:
The Travis CI build was failed for "Node.js 8" configuration in the same way as it fails for karma master (https://travis-ci.org/karma-runner/karma). So, seems this failed check can be skipped.

segrey:
@johnjbarton Could you please take a look into this PR? It helps IDE to annotate correctly method call where assertion is failed.
(The failed Travis build is not caused by this change.)

segrey:
Thanks!

PR #3370, opened 2019-09-27
chore: update dependencies

TimothyGu:
This updates pretty much all of the dependencies to the latest versions. Some packages are not upgraded:

- cucumber looks like the work might be more involved
- latest husky requires a more recent Git version than Ubuntu Trusty used on Travis CI currently

Additional changes were made to accomodate for the ESLint and isbinaryfile upgrades. test/e2e/tag.feature was changed to reduce some flakiness in the E2E test.


devoto13:
This PR is pretty outdated with the based branch, also many changes from it has already landed. Please open new smaller PRs if you think something has been missed.

PR #3497, opened 2020-05-05
Improve file type detection

falsandtru:
Suppress verbose warnings.

>06 05 2020 07:43:55.461:WARN [middleware:karma]: Invalid file type (js?flags=gated&features=default), defaulting to js.
06 05 2020 07:43:55.463:WARN [middleware:karma]: Invalid file type (js?config=TeX-AMS_CHTML,Safe), defaulting to js.

### DIFF
--- a/lib/helper.js
+++ b/lib/helper.js
@@ -141,6 +141,10 @@ const replaceWinPath = (path) => {
 
 exports.normalizeWinPath = process.platform === 'win32' ? replaceWinPath : _.identity
 
+exports.getFileType = (file) => {
+  return file.type || path.extname(file.path.split(/[?#]/, 1)[0] || '').substring(1)
+}
+
 exports.mkdirIfNotExists = (directory, done) => {
   // TODO(vojta): handle if it's a file
   /* eslint-disable handle-callback-err */
--- a/lib/middleware/karma.js
+++ b/lib/middleware/karma.js
@@ -11,7 +11,6 @@
  * - setting propert caching headers
  */
 
-const path = require('path')
 const url = require('url')
 const helper = require('../helper')
 
@@ -164,7 +163,7 @@ function createKarmaMiddleware (
           const scriptTags = []
           for (const file of files.included) {
             let filePath = file.path
-            const fileType = file.type || path.extname(filePath).substring(1)
+            const fileType = helper.getFileType(file)
 
             if (helper.isDefined(fileType) && !FILE_TYPES.includes(fileType)) {
               log.warn(`Invalid file type (${fileType}), defaulting to js.`)
--- a/lib/middleware/karma.js
+++ b/lib/middleware/karma.js
@@ -162,31 +162,31 @@ function createKarmaMiddleware (
 
           const scriptTags = []
           for (const file of files.included) {
-            let filePath = file.path
+            const urlPath = file.isUrl
+              ? file.path
+              : requestUrl === '/context.html'
+                ? filePathToUrlPath(file.path, basePath, urlRoot, proxyPath) + '?' + file.sha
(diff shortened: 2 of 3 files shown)


falsandtru:
@googlebot I signed it!

falsandtru:
Looks like a validation error of cla.

> The command "./scripts/validate-commit-msg.sh $COMMIT_TO_VALIDATE" exited with 1.

devoto13:
Thanks for the PR @falsandtru. The issue is that the commit message does not follow the [repository rules](http://karma-runner.github.io/4.0/dev/git-commit-msg.html]. This change will also need unit tests.

But before getting to it, I don't quite understand how you end up with query params added to your scripts. Can you share an example config and elaborate on your use case please?

falsandtru:
OK, I'll fix the commit message. My config is here.

https://github.com/falsandtru/securemark/blob/master/karma.conf.js

falsandtru:
Can you merge?

devoto13:
@falsandtru We still need unit tests to ensure future updates will not break this logic.

I think the easiest way is to extract `detectFileType(pathOrUrl)` function at the bottom of the same file and add `describe('detectFileType', () => { ... })` block with the relevant tests to the `test/unit/middleware/karma.spec.js` file.

falsandtru:
done

lib/helper.js
```diff
@@ -141,6 +141,10 @@ const replaceWinPath = (path) => {
 
 exports.normalizeWinPath = process.platform === 'win32' ? replaceWinPath : _.identity
 
+exports.getFileType = (file) => {
+  return file.type || path.extname(file.path.split(/[?#]/, 1)[0] || '').substring(1)
```
devoto13, line 145: Is there any specific reason why we need ` || ''` here? 
Otherwise LGTM.

devoto13:
Maybe also change commit message to something like "correctly detect file type for URLs with query params and fragment identifiers", so it is more clear what exactly is fixed.

lib/helper.js
devoto13: Also it may be better to first take extensions and then split as file name may contain `?` or `#` (as it was before). E.g. `myfile?.js` or `myfile#.js` are valid filenames on Linux.

Hm, actually `path.extname('http://example.com/mylib.js?load=test.md')` will fail result in `md`, not `js`. Maybe we need a more robust logic here...

falsandtru: > Is there any specific reason why we need || '' here?

String#split method can return an empty array.

> Probably we need a more robust logic here...

Yes. I'm expecting that someone write a stricter implementation later.

falsandtru:
I got an error.

> input: fix(middleware): detect file type for URLs with query params and fragment identifiers
✖   header must not be longer than 72 characters, current length is 85 [header-max-length]

lib/helper.js
falsandtru: Hm, looks like the source code treats paths only as URLs.

devoto13: > Yes. I'm expecting that someone write a stricter implementation later.

Well, current implementation breaks file type detection for `myfile?.js` and `myfile#.js` in favour of URLs. I don't think it is appropriate to break one corner case in favour of another.

Here is a better idea. Instead of having single method to handle both file path and URL we move the file type detection to constructors in `file.js` and `url.js` correspondingly. This way we can provide appropriate implementation depending on what we're dealing with.

E.g. 

`path.extname(filePath).substring(1)` for file paths
`path.extname(new URL('https://example.com/myfile.js?test=param').pathname).substring(1)` for URLs

falsandtru: Not sure. You should do such refactoring yourself.

devoto13: Okey, I'll try and submit an alternative PR in the coming days.

PR #3500, opened 2020-05-07
refactor(test): move execKarma into the World

devoto13:
Also generalize a step to accept any CLI arguments instead of only log-level. This should make this step more flexible. After the recent changes the removed step was not very clear as to which command (foreground or background) it applies to and is fully covered with expanded "I {command} Karma ..." step.

### DIFF
--- a/test/e2e/support/world.js
+++ b/test/e2e/support/world.js
@@ -1,4 +1,4 @@
-const { spawn } = require('child_process')
+const { exec, spawn } = require('child_process')
 const fs = require('fs')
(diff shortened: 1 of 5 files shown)


test/e2e/support/world.js
```diff
@@ -42,6 +42,7 @@ class World {
       frameworks: ['jasmine'],
       basePath: this.workDir,
       colors: false,
+      logLevel: 'warn',
```
devoto13, line 45: This value was previously set in `execKarma`.

PR #3504, opened 2020-05-07
refactor(test): remove no debug matching option

devoto13:
There is no need to filter out [DEBUG] logs as tests are run with logLevel = "warn" by default.


devoto13:
It was there from when this test was added in 2013, so hard to say exactly why it was needed. Maybe tests were run with log level = debug at that time. Now tests are run with log level = warn, so it does nothing. Test still checks what it is supposed to check - run fails without client arguments and passes with them. Thus showing that arguments are passed to the browser.

PR #3512, opened 2020-05-15
fix(dependencies): update production dependencies

devoto13:
Most of the updates are within allowed SemVer range. `tmp` and `rimraf` dropped support for ancient Node versions, which are not supported by Karma anyways.


devoto13:
Changed commit scope to `fix`, so it triggers a release and appears on the changelog.

PR #3517, opened 2020-05-18
chore: update ESLint packages

devoto13:
Most of the changes are automatic fixes from `npm run lint:fix`. The only non-trivial change is to `.hasOwnProperty()`, which is now reported by recommended ESLint rules. See [no-prototype-builtins](https://eslint.org/docs/rules/no-prototype-builtins) for more details.


devoto13:
Yeah, Don't like it either, but AFAICT there is no better alternative without changing the semantics of existing checks.

I was also thinking to replace `loggerCache` with the `Map`. Will submit it as a separate PR.

The other two occurrences are technically valid (although very unlikely). E.g. users can potentially add `hasOwnProperties: 42` to the `proxies` object in their configuration.

devoto13:
@johnjbarton Is there something you expect from me before merging or just want to give it more thought?

PR #3527, opened 2020-05-25
fix(proxy): use keepAlive agent

matz3:
Using an agent with keepAlive improves performance as the socket is
reused across multiple requests.

### DIFF
--- a/lib/middleware/proxy.js
+++ b/lib/middleware/proxy.js
@@ -1,4 +1,6 @@
 const url = require('url')
+const { Agent: httpAgent } = require('http')
+const { Agent: httpsAgent } = require('https')
 const httpProxy = require('http-proxy')
 const _ = require('lodash')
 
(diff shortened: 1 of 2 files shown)


lib/middleware/proxy.js
```diff
@@ -42,11 +43,13 @@ function parseProxyConfig (proxies, config) {
...
     }
     const changeOrigin = proxyConfiguration.changeOrigin || false
+    const keepAliveAgent = new http.Agent({ keepAlive: true })
     const proxy = httpProxy.createProxyServer({
       target: { host: hostname, port, https, protocol },
       xfwd: true,
       changeOrigin: changeOrigin,
-      secure: config.proxyValidateSSL
+      secure: config.proxyValidateSSL,
+      agent: keepAliveAgent
```
devoto13, line 52: Does this work if target URL is HTTPS?

matz3: Good catch. I completely missed that. Should be solved now.

devoto13: Another thing is that we probably also need to call `.destroy()` on the agent when proxy server is shut down to make sure sockets are not leaking.

(Although currently Karma is not doing a good job at cleaning up resources after itself it would be good to not add more such places.)

PS I'll take a more thorough look in the coming days as I'm not very familiar with this part of the codebase.

matz3: Thanks! I'll have a look into ensuring a proper cleanup

matz3: I added destroying the agents on exit, but need to check how this can be properly unit tested.

devoto13: Thanks! IMO We can ignore this test for now.

matz3:
Thanks for your review! I've squashed the commits and adopted the message.

PR #3531, opened 2020-06-07
fix: patch karma to allow loading virtual packages

### LINKED ISSUE #3530: Incorrect generated file paths when using Yarn 2 for virtual packages
Yarn 2 will create virtual packages for those dependencies that have peer dependencies. The Karma plugins that we use such as `karma-requirejs` have a peer dependency on `karma`. The virtual package path that Yarn 2 generates is:

`/Users/shahn/my-project-dir/.yarn/$$virtual/karma-requirejs-virtual-d16b0e1d06/0/cache/karma-requirejs-npm-0.2.5-fb217d2fa2-3.zip/node_modules/karma-requirejs/`

So for example the `karma-requirejs` plugin will try to load the `lib/adapter.js` file which would resolve to this path:

`/Users/shahn/my-project-dir/.yarn/$$virtual/karma-requirejs-virtual-d16b0e1d06/0/cache/karma-requirejs-npm-0.2.5-fb217d2fa2-3.zip/node_modules/karma-requirejs/lib/adapter.js`

Problem is that the Karma middleware will try to generate the HTML file with all these paths but it uses the `replace()` function in JS and the double dollar sign `$$` is a [special pattern](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_string_as_a_parameter). It means replace with `$`. So the paths generated to put in the HTML code is:

`/Users/shahn/my-project-dir/.yarn/$virtual/karma-requirejs-virtual-d16b0e1d06/0/cache/karma-requirejs-npm-0.2.5-fb217d2fa2-3.zip/node_modules/karma-requirejs/lib/adapter.js`

This errors out with 404s.

There are several ways to go about fixing this but I figured I just pass in an arrow function instead and that would turn off the special patterns.

snasirca:
Fixes #3530

### DIFF
--- a/lib/middleware/karma.js
+++ b/lib/middleware/karma.js
@@ -220,10 +220,10 @@ function createKarmaMiddleware (
           }) : []
 
           return data
-            .replace('%SCRIPTS%', scriptTags.join('\n'))
+            .replace('%SCRIPTS%', () => scriptTags.join('\n'))
             .replace('%CLIENT_CONFIG%', 'window.__karma__.config = ' + JSON.stringify(client) + ';\n')
             .replace('%SCRIPT_URL_ARRAY%', 'window.__karma__.scriptUrls = ' + JSON.stringify(scriptUrls) + ';\n')
-            .replace('%MAPPINGS%', 'window.__karma__.files = {\n' + mappings.join(',\n') + '\n};\n')
+            .replace('%MAPPINGS%', () => 'window.__karma__.files = {\n' + mappings.join(',\n') + '\n};\n')
             .replace('\n%X_UA_COMPATIBLE%', getXUACompatibleMetaElement(request.url))
         })
       })
--- a/test/unit/middleware/karma.spec.js
+++ b/test/unit/middleware/karma.spec.js
@@ -214,6 +214,21 @@ describe('middleware.karma', () => {
     callHandlerWith('/__karma__/context.html')
   })
 
+  it('should serve context.html without using special patterns when replacing script tags', (done) => {
+    includedFiles([
+      new MockFile('/.yarn/$$virtual/first.js', 'sha123'),
+      new MockFile('/.yarn/$$virtual/second.dart', 'sha456')
+    ])
+
+    response.once('end', () => {
+      expect(nextSpy).not.to.have.been.called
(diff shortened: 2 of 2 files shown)


snasirca:
@googlebot I signed it!

snasirca:
What would be the best way to add regression coverage to this change?

devoto13:
Thanks for fixing this!

A unit test with the problematic path similar to [this](https://github.com/karma-runner/karma/blob/master/test/unit/middleware/karma.spec.js#L202) should be enough.

acls:
I just ran into this issue yesterday and would like to get this merged as soon a possible. I can write the unit test if needed.

snasirca:
@acls @devoto13 I've added test coverage. Can you check if this is sufficient?

snasirca:
@devoto13 can you review this?

lib/middleware/karma.js
```diff
@@ -220,10 +220,10 @@ function createKarmaMiddleware (
           }) : []
 
           return data
-            .replace('%SCRIPTS%', scriptTags.join('\n'))
+            .replace('%SCRIPTS%', () => scriptTags.join('\n'))
             .replace('%CLIENT_CONFIG%', 'window.__karma__.config = ' + JSON.stringify(client) + ';\n')
             .replace('%SCRIPT_URL_ARRAY%', 'window.__karma__.scriptUrls = ' + JSON.stringify(scriptUrls) + ';\n')
```
devoto13, line 225: I believe this should be changed to function as well to cover [this branch](https://github.com/karma-runner/karma/blob/master/lib/middleware/karma.js#L107). Please change it and add a corresponding test.

crownofapollo:
@snasirca wondering about the status of this. I'm running into the same issue and found this potential fix.

snasirca:
> @snasirca wondering about the status of this. I'm running into the same issue and found this potential fix.

Sorry about the delay. This hasn't dropped off of radar. I'm on vacation and back by September 20. That'll be the earliest I can address the code review feedback.

elpddev:
@snasirca Thank you for the hard work. I can confirm this is working for our project.

johnjbarton:
This fix was subsequently merged in #3663

PR #3547, opened 2020-08-28
chore(test): fix client tests in older IE versions

devoto13:
By locking sinon to the older version. The problem was that new version of sinon brought newer versions of several dependencies, which use ES2016 Map, which is not available in the older IE versions. The solution is fragile as we can't reliably lock transitive dependencies, but should do for now to make master green.

As a proper solution we should consider cutting a new major release, where we would drop support for ancient browsers and more clearly document which browsers karma aims to support.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2291,9 +2291,9 @@
       }
     },
     "@sinonjs/commons": {
-      "version": "1.8.1",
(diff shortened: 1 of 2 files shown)


devoto13:
Example of passing tests is [here](https://travis-ci.org/github/karma-runner/karma/jobs/721941372#L256). GitHub is pulling info from the failed Travis build from another PR.

PR #3555, opened 2020-09-02
fix: remove broken link from docs - 06-angularjs.md

rmjoia:
The link [Full Spectrum Testing with AngularJS and Karma](http://www.yearofmoo.com/2013/01/full-spectrum-testing-with-angularjs-and-karma.html) is no longer working, http://www.yearofmoo.com cannot be reached.


devoto13 (approved):
@rmjoia Thanks for fixing this!

PR #3559, opened 2020-09-03
fix: revert source-map update

devoto13:
As it contains a breaking change per #3557.


dlawrence2060:
How long for this to be reviewed?

johnjbarton (approved):

PR #3567, opened 2020-10-08
fix: patch karma to allow loading virtual packages

### LINKED ISSUE #3530: Incorrect generated file paths when using Yarn 2 for virtual packages

handerss-tibco:
Rebase and additional branch coverage of #3531 which fixes #3530. Credit to @snasirca.

Not sure if its correct to create a new PR to karma-runner, but GH did not allow me to pick the nulogy fork as a base repo for this PR.

### DIFF
--- a/lib/middleware/karma.js
+++ b/lib/middleware/karma.js
@@ -220,10 +220,10 @@ function createKarmaMiddleware (
           }) : []
 
           return data
-            .replace('%SCRIPTS%', scriptTags.join('\n'))
+            .replace('%SCRIPTS%', () => scriptTags.join('\n'))
             .replace('%CLIENT_CONFIG%', 'window.__karma__.config = ' + JSON.stringify(client) + ';\n')
-            .replace('%SCRIPT_URL_ARRAY%', 'window.__karma__.scriptUrls = ' + JSON.stringify(scriptUrls) + ';\n')
-            .replace('%MAPPINGS%', 'window.__karma__.files = {\n' + mappings.join(',\n') + '\n};\n')
+            .replace('%SCRIPT_URL_ARRAY%', () => 'window.__karma__.scriptUrls = ' + JSON.stringify(scriptUrls) + ';\n')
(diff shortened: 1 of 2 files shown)


handerss-tibco:
Not sure if I can sign a Contributor License Agreements, will close for now.

devoto13:
@handerss-tibco If it is because of @snasirca's commit in the PR, I would suggest to make a new commit with your work only on top of the previous PR. Then we can get it in as two separate contributions.

handerss-tibco:
@devoto13 No, I don't think I can sign away rights to code I have written without approval from the company unfortunately.

PR #3592, opened 2020-12-17
feat: remove support for running dart code in the browser

devoto13:
As Dartium browser is no more, there are no browsers left, which support running Dart applications natively and therefore we can remove such support from Karma to reduce the maintenance effort.

BREAKING CHANGE: Using Karma to run Dart code in the browser is no longer supported. Use your favorite Dart-to-JS compiler instead.

`dart` file type has been removed without a replacement.

`customFileHandlers` DI token has been removed. Use [`middleware`](http://karma-runner.github.io/5.2/config/configuration-file.html#middleware) to achieve similar functionality.

`customScriptTypes` DI token has been removed. It had no effect, so no replacement is provided.


johnjbarton (approved):
thanks!

PR #3602, opened 2020-12-23
chore(build): remove obsolete Grunt tasks

devoto13:
These tasks are now handled by the semantic-release or other scripts and therefore we can clean them up.


johnjbarton (approved):
thanks!

PR #3611, opened 2021-01-06
feat(client): update banner with connection, test status, ping times

johnjbarton:
The banner is visible in videos so we can have some client side
info on the state if it breaks.

### DIFF
--- a/client/karma.js
+++ b/client/karma.js
@@ -2,7 +2,8 @@ var stringify = require('../common/stringify')
 var constant = require('./constants')
 var util = require('../common/util')
 
-function Karma (socket, iframe, opener, navigator, location, document) {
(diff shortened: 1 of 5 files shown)


client/karma.js
```diff
@@ -190,6 +191,7 @@ function Karma (socket, iframe, opener, navigator, location, document) {
     }
 
     socket.emit('karma_error', message)
+    self.updater.updateTestStatus(`karam_error ${message}`)
```
jginsburgn, line 194: Typo. Please fix.

johnjbarton: done

jginsburgn (approved):

devoto13 (approved):
I'm on vacation this week, so haven't looked too much into the details, but overall LGTM

PR #3616, opened 2021-01-14
fix(server): refactor file-list event interface

johnjbarton:
Remove the confusing promise container in favor of a current-file-list state.
The custom-then() function around a promise is hard to understand and reason about.
A simple object with a files property expresses the 'current files list' clearer.


lib/server.js
```diff
@@ -79,7 +79,7 @@ class Server extends KarmaEventEmitter {
       webServer: ['factory', createWebServer],
       serveFile: ['factory', createServeFile],
       serveStaticFile: ['factory', createServeStaticFile],
-      filesPromise: ['factory', createFilesPromise],
+      currentFiles: ['factory', createCurrentFiles],
```
devoto13, line 82: If we're making the breaking change here anyways. Maybe we can remove this provider altogether and ask consumers to inject `fileList` and use `fileList.files` property directly instead? Or are there any particular use cases for this wrapper?

devoto13: Ah, because `fileList.files` is a getter! Maybe we can make it a regular property, which is updated eagerly upon a change instead of lazy computation in a getter? Then we can implement the above simplification.

johnjbarton: I hadn't thought this was part of the public API, but I suppose it is.  I will revert the deletion of the function and emit a deprecation warning when it is called.

Having a `FileList` with a mutated field (getter then an object) is pretty horrible and my overall goal is to provide first class support for 'bundling'. There the `.included` list is mutated independent of the `FileList` patterns. 

In the presence of bundling the `FileList` is representing two different things:
* a list of configured patterns expanded into a list of Files
* a list of Files to be included in the test html.

So maybe it would be better to have another class represent the files to be included.   A bundler would consume one instance an produce another.  This class would be the value in the injector.

devoto13: > Having a FileList with a mutated field (getter then an object) is pretty horrible

Well, I don't see how it is different to the extra object with a mutable property, which can be modified by plugins exactly the same way as the property of the `FileList`. IMO it is cleaner to get rid of extra logic and use `FileList.files` directly.

What I'm thinking here is that instead of extra provider subscribing to an event and fetching files from the computable getter we compute the same list eagerly upon `FileList` modification and store it in the `FileList` property. Exactly the same logic, but computed list is stored in the different place.

> my overall goal is to provide first class support for 'bundling'

Can you explain a bit more the problem you want to solve and what you want to achieve eventually?

johnjbarton: `FileList.files`:  yes I can see that reasoning about the content of `.files` will be easier with eager computation.

bundling: I don't know how other teams do bundling for tests, but our bundler is confusing because the `FileList.files` is both an input configuration and a computed output, specifying bundle files to be `.included` in the `context.html`. To improve the system I propose to split the pipeline and add a phase for bundling.  The bundling phase will sit between the config/preprocessing and the file service. It will be triggered by `file_list_modified` and it will emit `bundle_list_modified` with an argument `FileList` which will contain bundles rather than config/preprocessed results.

With this design the `FileList` data structure has two important and different instances, one for inputs and one for outputs.  I think they should both be dependency injection values or both parameters of emit event, not a mix like now.

WDYT?  Is that enough to get the idea?

johnjbarton: > What I'm thinking here is that instead of extra provider subscribing to an event and fetching files from the computable getter we compute the same list eagerly upon FileList modification and store it in the FileList property.

The configuration, preprocessor, and bundled `FileList.files` are pipeline results so I think passing them *only* via events makes more sense than storing them on properties.  On properties we can't access them unless we know they are ready.

I don't think that promises/async functions are a good match either because the pipeline can be entered at any time by the `watch` feature. So the downstream components have to be ready to reprocess on future events.  Wrapping an unending loop around a promise is extra confusion compared to simply having an event handler. 

So: get rid of the provider and the getter. Compute the `.files` when ready, fire event, trigger handler, compute `.files` when ready, ...

Oh, I can see now one issue I forgot earlier: the `.files` property is not really `FileList` thing so much as computed result from the `FileList` values. One computation is "extract `.included`/`.served` preprocessed files into lists". Another computation is "bundle `.included`/`.served` preprocessed files for file service".  So maybe the property should not be `FileList.files` but instances of a new class.

devoto13: > WDYT? Is that enough to get the idea?

Yes, this makes some sense!

You said "provide first class support for 'bundling'", but from the description, I believe exactly what you're after is a convenient way of adding transitive dependencies from the preprocessor, aren't you?
So we should keep something like what `karma-esm` plugin from https://github.com/karma-runner/karma/issues/3335 is doing in mind.

Contrary, one of the popular bundlers ([`karma-webpack`](https://github.com/ryanclark/karma-webpack/blob/master/src/karma-webpack.js#L428) takes a different approach: it uses a custom middleware to serve any custom resources, which may be loaded by the test page, so there is no need to modify `served`/`included` at all.

> On properties we can't access them unless we know they are ready.

I don't really see a fundamental difference here. While a bundling process is running you can continue using an outdated value from a previous event... or an outdated value, which you can fetch from a property anytime. Having an event is a must, as consumers should know when a bundle is ready. Having a property may provide extra convenience for components, which just need the latest list (e.g. middleware).

> Oh, I can see now one issue I forgot earlier: the .files property is not really FileList thing so much as computed result from the FileList values. One computation is "extract .included/.served preprocessed files into lists". Another computation is "bundle .included/.served preprocessed files for file service". So maybe the property should not be FileList.files but instances of a new class.

This sounds like a pretty massive breaking change. I wonder if we can solve the problem by making preprocessors more powerful (so they can also register new files). Then we can adjust [this logic](https://github.com/karma-runner/karma/blob/master/lib/file-list.js#L90) and the feature can be fully backward compatible!

---

I'm not sure how your bundler work, but I guess another feature which will be needed for the first-class bundlers support is to allow preprocessors to register/unregister new patterns for watching in a similar fashion 🤔

johnjbarton: Thanks for the good feedback. 

The change I have in mind is pretty small in terms of code, and by default it wouldn't do much: insert a post-preprocessing step between `FileList` and `middleware/karma`.  So `FileList` works as now and preprocesses file-by-file; and `middleware.karma` works as now, writing `.html` based on `.included`.  By default the bundling step just copies `FileList.files` so nothing changes. A bundler would map the input `FileList.files` to a new list of `.included` files (the bundles) so `middleware.karma` writes the `.html` in terms of bundles.

An esm-driven solution would not be better or worse because the issue with esm-driven testing is around `.served`: the `FileList` patterns don't support the dynamic import approach.  As far as I can tell the esm-driven testing issue is simply a matter of better pattern support. If an esm module requests path/foo.js then path/foo.js has to be in the file system and thus has to be expressible as pattern.  Maybe there is a config-file-not-watched-and-updated issue too.

devoto13: Okay, I guess this should be safe enough. When you say *the bundling step* do you think of a new type of a plugin? Or just another object in DI, which can be replaced by a plugin?

> As far as I can tell the esm-driven testing issue is simply a matter of better pattern support. If an esm module requests path/foo.js then path/foo.js has to be in the file system and thus has to be expressible as pattern. Maybe there is a config-file-not-watched-and-updated issue too.

I guess the main point of that feature request was to avoid the need to specify the patterns manually. I think bundling plugin can partially solve the use case by traversing the dependency tree and adding missing patterns as `.served`. But I agree, we should probably handle that use case separately.

devoto13: Nevermind, I got it now.

devoto13 (approved):
LGTM once CI is happy

johnjbarton:
I added the bundler here. I will try to separate the addition from the refactoring.

johnjbarton:
To avoid introducing a change to middleware/karma and then just turn around and change it again, I combined the refactoring with the bundler introduction in #3638

PR #3618, opened 2021-01-14
chore(test): run unit tests without grunt wrapper

devoto13:


package.json
```diff
@@ -496,7 +493,7 @@
   "scripts": {
     "lint": "eslint . --ext js --ignore-pattern *.tpl.js",
     "lint:fix": "eslint . --ext js --ignore-pattern *.tpl.js --fix",
-    "test:unit": "grunt",
+    "test:unit": "mocha 'test/unit/**/*.spec.js'",
```
XhmikosR, line 496: the single quotes should be double quotes (escaped) otherwise this won't work in all shells.

devoto13: Good point, thanks!

johnjbarton (approved):

Issue #3619, opened 2021-01-14
Karma error after update to version 6.0.0


SymbioticKilla:
Hi,

I get following error after update:

```
ERROR [karma-server]: Server start failed on port 9876: Error: No provider for "customFileHandlers"! (Resolving: framework:@angular-devkit/build-angular -> customFileHandlers)   
ERROR [karma-server]: Error while calling exit event listeners
TypeError: Cannot read property 'name' of undefined
    at TfsReporter.onExit (C:\Dev\***\spa\node_modules\karma-tfs-reporter\index.js:53:38)
    at Server.<anonymous> (C:\Dev\***\spa\node_modules\karma\lib\events.js:40:26)
    at Server.emit (events.js:327:22)
    at Server.emitExitAsync (C:\Dev\***\spa\node_modules\karma\lib\server.js:146:12)
    at Server._close (C:\Dev\***\spa\node_modules\karma\lib\server.js:413:10)
    at Server.start (C:\Dev\***\spa\node_modules\karma\lib\server.js:120:12)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)

```

alecgibson:
This is probably a `karma-webpack` issue: https://github.com/ryanclark/karma-webpack/issues/462

devoto13:
Note that 6.0.0 is a major release, which has breaking changes. It is not compatible with the current release of `@angular-devkit/build-angular` package, hence the error. You should have received a warning about incompatible peer dependency from NPM when you updated to it.

The support for Karma 6.0.0 has already been added to Angular CLI (https://github.com/angular/angular-cli/pull/19784) and will probably be released in 11.1.0. Until then, please stick with Karma 5.x.

SymbioticKilla:
@devoto13 Thanks! Will it be back ported to 10.2?

devoto13:
I don’t think so. AFAIK only critical fixes get backported.

SymbioticKilla:
Ist it possible to back port CVE Fixes to Karma 5?

devoto13:
Unfortunately we don't have a process to backport changes at the moment... I guess we can either try to make a patch release on 5.x branch or restore `customFileHandlers` provider with the deprecation warning on the 6.x branch (so people can use Karma 6.x with older Angular CLI with the peer dependency warning). 

@johnjbarton Which option do you think is better?

PR #3624, opened 2021-01-19
fix: restore `customFileHandlers` provider

devoto13:
The removal of `customFileHandlers` turned out to be more disruptive change than anticipated as this provider is still used by several popular plugins: https://github.com/karma-runner/karma/issues/3619, https://github.com/ryanclark/karma-webpack/issues/462, https://github.com/angular/angular-cli/issues/19815. Hence we restore the provider and print a deprecation warning to make upgrading easier. This should give more time for the plugin authors to release new versions and users to adopt these versions.

How it looks now:

image

### DIFF
--- a/lib/web-server.js
+++ b/lib/web-server.js
@@ -16,6 +16,25 @@ const proxyMiddleware = require('./middleware/proxy')
 
 const log = require('./logger').create('web-server')
 
+function createCustomHandler (customFileHandlers, config) {
+  let warningDone = false
+
+  return function (request, response, next) {
+    const handler = customFileHandlers.find((handler) => handler.urlRegex.test(request.url))
+
+    if (customFileHandlers.length > 0 && !warningDone) {
+      warningDone = true
+      log.warn('The `customFileHandlers` is deprecated and will be removed in Karma 7. Please upgrade plugins relying on this provider.')
+    }
+
+    return handler
+      ? handler.handler(request, response, 'fake/static', 'fake/adapter', config.basePath, 'fake/root')
+      : next()
+  }
+}
+
+createCustomHandler.$inject = ['customFileHandlers', 'config']
+
 function createFilesPromise (emitter, fileList) {
   // Set an empty list of files to avoid race issues with
   // file_list_modified not having been emitted yet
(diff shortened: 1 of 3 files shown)


lib/web-server.js
```diff
@@ -16,6 +16,25 @@ const proxyMiddleware = require('./middleware/proxy')
 
 const log = require('./logger').create('web-server')
 
+function createCustomHandler (customFileHandlers, config) {
+  let warningDone = false
+
+  return function (request, response, next) {
+    const handler = customFileHandlers.find((handler) => handler.urlRegex.test(request.url))
+
+    if (customFileHandlers.length > 0 && !warningDone) {
```
devoto13, line 25: The awkward logic is here because handlers are added dynamically (e.g. https://github.com/angular/angular-cli/pull/19784/files#diff-7e99b8c9ae4a43ae87d525fccfd19e3c7449e3f12b3d0e07aedf208a8a451f3cL223) and we can't rely on the presence of the provider either, because it is always there. 

So I resorted to checking that there is at least one custom handler is present when handling a request and printing the warning only once to avoid spamming the logs. 

The check is not perfect as the warning is printed only when this handler is reached but this should be good enough in practice.

alan-agius4:
JFYI, Angular CLI 11.0 doesn't support Karma 6, thus the error is expected. Karma 6 is supported in Angular 11.1.

devoto13:
@alan-agius4 Yes, I know. But people still upgrade to Karma 6 and get the error. I think it is better to have the warning to give the ecosystem time to adapt. This also affects older versions of `karma-webpack` plugin (and there is no stable version with the fix). Also some people want to upgrade to fix some CVEs (https://github.com/karma-runner/karma/issues/3619#issuecomment-760730884).

alan-agius4:
@devoto13, Agreed! 

Although warnings as typically ignored. Strictly speaking in Angular CLI, they already get a warning of unmeet peer dependencies.

SymbioticKilla:
@devoto13 Thank you very much! Really a good decision.

test/unit/web-server.spec.js
```diff
@@ -180,6 +182,23 @@ describe('web-server', () => {
         })
     })
 
+    it('should load custom handlers', () => {
+      servedFiles(new Set())
+
+      // TODO(vojta): change this, only keeping because karma-dart is relying on it
```
johnjbarton, line 188: Please change or remove this comment as karma-dart is no longer the blocker.

johnjbarton (commented):
Maybe we need an Issues: Release 7 blockers to remember to follow through.

johnjbarton (approved):

test/unit/web-server.spec.js
devoto13: Fixed and added a point to https://github.com/karma-runner/karma/issues/3503.

PR #3626, opened 2021-01-20
fix(server): set maxHttpBufferSize to the socket.io v2 default

### LINKED ISSUE #3621: v6 coverage runs can fail dues to maxHttpBufferSize change in socket.io v3
Hi guys,

Our build starts to fail when we upgrade karma to v6.0.0. All tests run without problem, but Karma thinks the browser disconnects right after executing the tests (or the browser actually disconnects, but that doesn't make Karma yell in v5.2).

I tracked it down and reproduced with a single test running in `.only` (we're using Mocha as a framework) so I suppose it's not a problem with the actual tests, but maybe related to the actual amount of JS code we're oading in the browser. Again, in v5.2 they run fine.

I'm attaching [karma.v5.log](https://github.com/karma-runner/karma/files/5821996/karma.v5.log) and [karma.v6.log](https://github.com/karma-runner/karma/files/5821998/karma.v6.log) in DEBUG level that show the exact same run, one with 5.2.3 and the other with 6.0.0. The only step in-between these two runs is the upgrade of Karma.

A few additional information on our setup (and you have the full `config` object at the beginning of the log file)
* The actual application is using AngularJS
* We use Mocha as our test framework
* We run tests in Chrome headless only, using [karma-chrome-launcher](https://www.npmjs.com/package/karma-chrome-launcher)
* We use [karma-coverage-istanbul-instrumenter](https://www.npmjs.com/package/karma-coverage-istanbul-instrumenter) and [karma-coverage-istanbul-reporter](https://www.npmjs.com/package/karma-coverage-istanbul-reporter) to generate coverage report information
* We preprocess HTML templates to make them AngularJS modules (yeah... more JS code loaded :p)

I'm kinda stuck now, not knowing where to go to track the problem even further. I will provide any additional information you may require but would appreciate help from Karma's community. In addition, I volounteer to make a PR if we eventually find a problem in Karma itself !

Thanks,
Regards,

David

johnjbarton:
Fixes #3621

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -43,7 +43,9 @@ function createSocketIoServer (webServer, executor, config) {
     transports: config.transports,
     forceJSONP: config.forceJSONP,
     // Default is 5000 in socket.io v2.x and v3.x.
-    pingTimeout: config.pingTimeout || 5000
+    pingTimeout: config.pingTimeout || 5000,
+    // Default in v2 is 1e8 and coverage results can fail at 1e6
+    maxHttpBufferSize: 1e8
   })
 
   // hack to overcome circular dependency


devoto13 (approved):
Great :+1:

ddolcimascolo:
Thx @johnjbarton

Issue #3631, opened 2021-01-25
Process exits without exceptions or logs when using `karma.config.parseConfig`


npetruzzelli:
I'm using `karma.config.parseConfig` to create a configuration using the public API:
https://karma-runner.github.io/6.0/dev/public-api.html

And my code has no way of being notified of errors. For example, the process exits silently if you pass a relative path to `parseConfig`. No exceptions are thrown and the logger doesn't output any messages to the console. Neither me (via the console) nor my application (via thrown exceptions) has any way of knowing that something went wrong. 

I am aware that using the CLI, it would use `path.resolve` to create an absolute path, but that is not the issue. I want to be informed about problems with the input being provided so that I can then tell the user or take some kind of action based on the kind of error it is.

**EXAMPLE**
```javascript
console.log('START OF FILE')
const karma = require('karma')

// When passing a relative path, the code attempts to locate the file relative to 
// `node_modules/karma/lib/conf.js`, not relative to your file or the current 
// working directory, but the error is caught and not re-thrown.
const myConfig = karma.config.parseConfig("./path/to/karma.conf.js")

console.log('END OF FILE') // The process exits before this statement is executed.
```

**RELATED CODE**

https://github.com/karma-runner/karma/blob/e246461c528dd84690b6d086b948986046f28ce7/lib/config.js#L6-L7

https://github.com/karma-runner/karma/blob/e246461c528dd84690b6d086b948986046f28ce7/lib/config.js#L354-L424

**ROUGH START**

Log4js isn't setup until the function is almost done, this might explain why logging doesn't seem to be enabled. I'm not familiar with that library, so I am not sure what a good solution would look like, but a possible work around for critical errors that normally exit the process might look like the following:

```diff
- function parseConfig (configFilePath, cliOptions) {
+ function parseConfig (configFilePath, cliOptions, functionOptions) {
+   // TODO: Can optional chaining be used in this repository?
+   const throwErrors = functionOptions && functionOptions.throwErrors === true
   let configModule
   if (configFilePath) {
     try {
       configModule = require(configFilePath)
       if (typeof configModule === 'object' && typeof configModule.default !== 'undefined') {
         configModule = configModule.default
       }
     } catch (e) {
       log.error('Error in config file!\n  ' + e.stack || e)
 
       const extension = path.extname(configFilePath)
       if (extension === '.coffee' && !COFFEE_SCRIPT_AVAILABLE) {
[...]

PR #3635, opened 2021-01-26
feat(config): improve `karma.config.parseConfig` error handling

npetruzzelli:
This PR add the ability of Public API users of `parseConfig` to opt-in to throwing errors when a critical failure is encountered.

The current behavior is to exit the process immediately, which denies the API consumer the ability to log or respond to the error at their application's level. 

Related Issue: #3631


npetruzzelli:
Before I clean up comments and attempt to write tests for the new code, I was hoping to get feedback on this first attempt.

npetruzzelli:
@googlebot I signed it!

lib/utils/log-utils.js
```diff
@@ -0,0 +1,206 @@
+
```
devoto13, line 1: Thanks for all this work, @npetruzzelli! But I think you heavily overdo it. We don't really need a new powerful logging utility here. It is enough to make one simple helper function in the same file (similar to what I presented in the issue) and use this utility in 3 places in `parseConfig` function where we currently have `log.error() + process.exit(1)` calls.

This warning logic is only needed temporarily and it is written to be removed, so it does not need to be flexible or reusable or anything like that.

npetruzzelli: Got it. At the very least I learned a bit more about the currently logging works. I'll update the PR using your comment in the issue as a template. I'll needed to change it a little bit to account for:

-   the one instance where multiple arguments are logged
-   the import failure logging multiple times before exit

npetruzzelli:
I wasn't sure if you intended for `fail` to be used outside of `parseConfig`. If that was the case, the signature would need only a minor change.

```javascript
function fail (logArgs, failOptions) {
    const _logArgs = Array.isArray(logArgs) ? logArgs : [logArgs]
    log.error(..._logArgs )
    if (failOptions && failOptions.throwErrors === true) {
    // ...
}
// in `parseConfig`
fail (message, parseOptions)
fail ([message, e], parseOptions)



// -- OR --

function fail (logArgs, throwErrors) {
    const _logArgs = Array.isArray(logArgs) ? logArgs : [logArgs]
    log.error(..._logArgs )
    if (throwErrors === true) {
    // ...
}
```

lib/config.js
```diff
@@ -360,21 +379,22 @@ function parseConfig (configFilePath, cliOptions) {
...
       } else if (extension === '.ls' && !LIVE_SCRIPT_AVAILABLE) {
-        log.error('You need to install LiveScript.\n  npm install LiveScript --save-dev')
+        errorMessageSegments.push('You need to install LiveScript.\n  npm install LiveScript --save-dev')
       } else if (extension === '.ts' && !TYPE_SCRIPT_AVAILABLE) {
-        log.error('You need to install TypeScript.\n  npm install typescript ts-node --save-dev')
+        errorMessageSegments.push('You need to install TypeScript.\n  npm install typescript ts-node --save-dev')
       }
-      return process.exit(1)
+      const configModuleImportErrorMessage = errorMessageSegments.join('\n')
+      return fail(configModuleImportErrorMessage)
```
devoto13, line 394: I think "You need to install ..." logs are more of a hint on where the error could have come from, so we can skip them from the thrown error and only keep them as regular logs. With this assumption, we can move the actual error log and `process.exit(1)` together and avoid `errorMessageSegments`.

```suggestion
      return fail('Error in config file!\n  ' + e.stack || e)
```

devoto13 (commented):
Please also update the usage of `parseConfig` in Karma's source code, so it does not rely on the deprecated behavior (last point in https://github.com/karma-runner/karma/issues/3631#issuecomment-767006344).

And add a unit test for `parseConfig` function with `throwErrors: true`.

Otherwise, it looks good :+1:
Thanks!

devoto13:
> I wasn't sure if you intended for fail to be used outside of parseConfig. If that was the case, the signature would need only a minor change.

Good to know! Let's keep it simple here and we can adjust the signature in the next PR (if/when necessary).

lib/config.js
npetruzzelli: > Please also update the usage of `parseConfig` in Karma's source code, so it does not rely on the deprecated behavior (last point in [#3631 (comment)](https://github.com/karma-runner/karma/issues/3631#issuecomment-767006344)).

```javascript
    let config
    try {
      // A)
      // Did you want this and an option passed to it as the third argument?
      config = cfg.parseConfig(cliOptions.configFile, cliOptions)

      // B)
      // or did you want the new behavior enabled immediatly since the catch
      // block will done or exit anyway?
      config = cfg.parseConfig(cliOptions.configFile, cliOptions, { throwErrors: true })
    } catch (parseConfigError) {
      (done || process.exit)(1)
    }
```

Did you want the server code to enable the new behavior?

---

> ... we can move the actual error log and process.exit(1) together and get rid of errorMessageSegments variable.

> And add a unit test for `parseConfig` function with `throwErrors: true`.

Will do!

devoto13: B), so we behave the same as before (exit the process), but no longer rely on the deprecated `process.exit()` call in the `parseConfig` method and don't need to change our code when we remove it in the next major release.

Two notes:
- We should probably `return` in the catch block to prevent any further execution in case custom `done` is passed in.
- It may be cleaner to use a default argument instead of inline fallback: `constructor (cliOptions, done = process.exit)`.

npetruzzelli: > * It may be cleaner to use a default argument instead of inline fallback: `constructor (cliOptions, done = process.exit)`.

The previous behavior was "if `done` is **falsey**, then call process.exit()."

`constructor (cliOptions, done = process.exit)` changes it to "if `done` is **`undefined`**, then call process.exit()."

**Are you ok with this?**

```javascript
  constructor (cliOptions, done = process.exit) {
    super()
    logger.setupFromConfig(cliOptions)

    this.log = logger.create('karma-server')

    this.loadErrors = []

    let config
    try {
      config = cfg.parseConfig(cliOptions.configFile, cliOptions, { throwErrors: true })
    } catch (parseConfigError) {
      done(1)
      return
    }

    this.log.debug('Final config', util.inspect(config, false, /** depth **/ null))

    let modules = [{
      helper: ['value', helper],
      logger: ['value', logger],
      done: ['value', done], // PREVIOUSLY:    done: ['value', done || process.exit],
```

---

Neither case guards against `done` being a non-callable/non-function value. That kind of change is probably out of scope for this PR, and that is if you determine it to be necessary at all.

devoto13: You're right. Technically falling back on `undefined` instead of a falsy value is a breaking change. So let's play it safe and restore the original behavior with the TODO to change it in the next major.

Besides this, it LGTM! Thanks again for doing this work :+1:

----

> Neither case guards against done being a non-callable/non-function value. That kind of change is probably out of scope for this PR, and that is if you determine it to be necessary at all.

Yeah, it's out of scope for this PR and I doubt it's worth the effort in general.

npetruzzelli: With that in mind, should I make the following change (Option A) as well?

```diff
     let config
     try {
       config = cfg.parseConfig(cliOptions.configFile, cliOptions, { throwErrors: true })
     } catch (parseConfigError) {
-      done(1)
+      process.exit(1) // ------ // Option A
+      (done || process.exit)(1) // Option B
       return
     }
```
Previously, `parseConfig` only exited the process here. It did not attempt to call `done`.

devoto13: I think (B) is fine. `done` was already called this way for other errors, so it should be fine. Even though behavior is different now.

devoto13 (approved):

devoto13:
@johnjbarton I think this is ready to land. Can you take a look?

johnjbarton:
Thanks!  For future reference, after a few commits, esp ones that revert changes, its probably better to squash the commits on your branch and push -f to the PR so all we see are the diffs that matter.

lib/config.js
```diff
@@ -351,7 +351,26 @@ const CONFIG_SYNTAX_HELP = '  module.exports = function(config) {\n' +
   '    });\n' +
   '  };\n'
 
-function parseConfig (configFilePath, cliOptions) {
+function parseConfig (configFilePath, cliOptions, parseOptions) {
+  function fail () {
+    log.error(...arguments)
```
XhmikosR, line 356: Is this supported on IE?

devoto13: This is server-side code, so IE is not relevant here.

PR #3637, opened 2021-01-26
fix(context): do not error when karma is navigating in iframe

radnan:
reset karmaNavigating flag only after iframe is loaded. otherwise we run in to a race condition where the flag is reset before URL changes and onbeforeunload is called afterwards making the handler think the navigation was not part of context change.

### DIFF
--- a/client/karma.js
+++ b/client/karma.js
@@ -128,8 +128,10 @@ function Karma (updater, socket, iframe, opener, navigator, location, document) 
       }
     // run in iframe
     } else {
+      iframe.onload = () => {
+        karmaNavigating = false
+      }
       iframe.src = policy.createURL(url)
-      karmaNavigating = false
     }
   }
 
--- a/static/karma.js
+++ b/static/karma.js
@@ -138,8 +138,10 @@ function Karma (updater, socket, iframe, opener, navigator, location, document) 
       }
     // run in iframe
     } else {
+      iframe.onload = () => {
+        karmaNavigating = false
+      }
       iframe.src = policy.createURL(url)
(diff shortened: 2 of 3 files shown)


client/karma.js
```diff
@@ -128,8 +128,10 @@ function Karma (updater, socket, iframe, opener, navigator, location, document)
       }
     // run in iframe
     } else {
+      iframe.onload = () => {
+        karmaNavigating = false
+      }
       iframe.src = policy.createURL(url)
```
johnjbarton, line 134: Immediately before this statement, `karmaNavigating` should be `true`. This statement sets the iframe URL.  Any action after this statement that causes an unload event is an error.  Thus we immediately set `karmaNavigating` to be `false`.

client/karma.js
```diff
@@ -128,8 +128,10 @@ function Karma (updater, socket, iframe, opener, navigator, location, document)
       }
     // run in iframe
     } else {
+      iframe.onload = () => {
+        karmaNavigating = false
```
johnjbarton, line 132: This change would delay the check on unload events until the test was completely loaded.  Thus code that executes during the load sequence and navigates the page will be allowed to execute, contrary to our goal of preventing navigation during testing.

Rather than a race is it possible that your test is attempting to navigate during test load?

client/karma.js
radnan: Looks like we were using a really old version of chrome (71) where the order wasn't deterministic for some reason. Meaning it was actually triggering the unload after the flag was being set to false. Fixed it on our side by upgrading chrome. Thanks!

PR #3638, opened 2021-01-27
feat(bundler): add a bundler layer

### LINKED ISSUE #3633: Feature Request: Introduce a "bundle" step in the file processing pipeline.
Tests and app code get through a pipeline in Karma:

1. They are determined (wildcard resolution).
2. They are loaded into memory (and cache).
3. Watching is set up (if enabled).
4. They undergo preprocessing.
5. They are inserted in the respective `context.html` template expansion.
6. They are served.

Preprocessing takes in a file and modifies it (i.e. input: 1 file; output: 1 file). However, this pipeline misses a step that offers a lot of potential: bundling. After preprocessing (between 3 and 4), bundling takes many files and outputs many as well.

Bundling could be used for:

- Splitting large script files. This has been a need in our tests on Karma, as frequently big scripts mess up with the event loop and memory.
- Packing/chunking scripts for lazy loading (https://webpack.js.org/guides/code-splitting/).

johnjbarton:
The input config supplies 'patterns' which FileList expands into Files.
These files are preprocessed and then 'file-list-modified' is fired.
The event offers a pair of file lists, {included, served}.
The bundle layer maps this pair to a new pair where fewer files are included.
By default the bundler copies the inputs to outputs.

The bundled files are stored on currentWebFiles which replaces
the FileList.files in the middleware/karma.

Fixes #3633


lib/server.js
```diff
@@ -82,7 +83,10 @@ class Server extends KarmaEventEmitter {
       webServer: ['factory', createWebServer],
       serveFile: ['factory', createServeFile],
       serveStaticFile: ['factory', createServeStaticFile],
+      // Obsolete, do not use
```
jginsburgn, line 86: Maybe we can display a warning when people use this?

jginsburgn (approved):

codymikol:
I actually was just about to make this feature request and to my surprise here it is 💯

PR #3640, opened 2021-01-28
fix(server): clean up vestigial code from proxy

jginsburgn:
When using the proxy feature of Karma, the target value can include the [scheme](https://tools.ietf.org/html/std66#section-3.1). It was used to determine the `https` variable to be sent to the [`http-proxy`](https://www.npmjs.com/package/http-proxy) `.createProxyServer` method. However, it is now disregarded by that package. Therefore, this PR cleans it up.

### DIFF
--- a/lib/middleware/proxy.js
+++ b/lib/middleware/proxy.js
@@ -1,6 +1,4 @@
 const url = require('url')
-const { Agent: httpAgent } = require('http')
-const { Agent: httpsAgent } = require('https')
 const httpProxy = require('http-proxy')
 const _ = require('lodash')
 
@@ -34,24 +32,17 @@ function parseProxyConfig (proxies, config) {
 
     const hostname = proxyDetails.hostname || config.hostname
     const protocol = proxyDetails.protocol || config.protocol
-    const https = proxyDetails.protocol === 'https:'
-    let port
-    if (proxyDetails.port) {
-      port = proxyDetails.port
-    } else if (proxyDetails.protocol) {
-      port = https ? '443' : '80'
-    } else {
-      port = config.port
+    const defaultPorts = {
+      'http:': '80',
+      'https:': '443'
     }
+    const port = proxyDetails.port || defaultPorts[proxyDetails.protocol] || config.port
     const changeOrigin = proxyConfiguration.changeOrigin || false
-    const Agent = https ? httpsAgent : httpAgent
-    const agent = new Agent({ keepAlive: true })
     const proxy = httpProxy.createProxyServer({
-      target: { host: hostname, port, https, protocol },
+      target: { host: hostname, port, protocol },
       xfwd: true,
       changeOrigin: changeOrigin,
-      secure: config.proxyValidateSSL,
-      agent
+      secure: config.proxyValidateSSL
     })
 
     ;['proxyReq', 'proxyRes'].forEach(function (name) {
@@ -71,7 +62,7 @@ function parseProxyConfig (proxies, config) {
       res.destroy()
     })
 
-    return { path: proxyPath, baseUrl: pathname, host: hostname, port, https, proxy, agent }
+    return { path: proxyPath, baseUrl: pathname, host: hostname, port, proxy }
   }), 'path').reverse()
 }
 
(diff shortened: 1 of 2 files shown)


lib/middleware/proxy.js
```diff
@@ -34,7 +34,7 @@ function parseProxyConfig (proxies, config) {
 
     const hostname = proxyDetails.hostname || config.hostname
     const protocol = proxyDetails.protocol || config.protocol
-    const https = proxyDetails.protocol === 'https:'
+    const https = protocol === 'https:'
```
johnjbarton, line 37: The `proxyDetails` result from parsing the `proxyUrl`.  How then can `proxyDetails.protocol` be nullish? Seems to me that a better fix is to delete `|| config.protocol` and the variable `protocol` because it provides no value. Unless it does: what is the use case?

johnjbarton (changes requested):

lib/middleware/proxy.js
jginsburgn: Well. I can assure you that it can be nullish/falsy, because some tests that I have, were solved by this change.

devoto13:
The fix looks correct to me. An example would be a path-only proxy when Karma is configured to run over https. E.g.

```
{
  protocol: 'https',
  proxies: {
    '/img/': '/base/test/images/'
  }
  ...
}
```

See *Loading assets* section (near the end): http://karma-runner.github.io/6.0/config/files.html

Can you add a test cased based on the above example, please?

johnjbarton:
Yes the test would clarify the use case

jginsburgn:
Upon trying to add a test for the use case that @devoto13 mentioned, I figured that it makes no sense to have a comms protocol (e.g. `https` or `http`) when proxying requests between paths. That is because the target path is in the local filesystem and no connection should be done.

jginsburgn:
Investigating more, the NPM module `http-proxy` that we are using, does not make use of the `https` variable that we were sending it. Also, it disregards the `agent` configuration. Therefore, I cleaned this logic up and its corresponding tests.

devoto13:
> Investigating more, the NPM module http-proxy that we are using, does not make use of the https variable that we were sending it. Also, it disregards the agent configuration. Therefore, I cleaned this logic up and its corresponding tests.

While `target.https` is indeed seems to be leftover from long time ago and can be cleaned up, the `agent` property is a [valid option](https://www.npmjs.com/package/http-proxy#options) and is needed for https://github.com/karma-runner/karma/commit/b77f94c1c95d0f717866eae42d478d09544b6d9d#diff-4bc20220c45a6d4611b038e5f55155e3fb59d9f9151c466efccab75fb96183f9 as far as I can tell. What makes you think that `http-proxy` disregards it?

jginsburgn:
Only the native raw agents are used, according to my understanding.
https://github.com/http-party/node-http-proxy/blob/9b96cd725127a024dabebec6c7ea8c807272223d/lib/http-proxy/passes/web-incoming.js#L11

devoto13:
The naming is confusing. 

I haven't really debugged it, but from a quick code scan the `agent` option is copied into `requestOptions` [here](https://github.com/http-party/node-http-proxy/blob/9b96cd725127a024dabebec6c7ea8c807272223d/lib/http-proxy/index.js#L50) and passed through [here](https://github.com/http-party/node-http-proxy/blob/9b96cd725127a024dabebec6c7ea8c807272223d/lib/http-proxy/index.js#L81) and [here](https://github.com/http-party/node-http-proxy/blob/9b96cd725127a024dabebec6c7ea8c807272223d/lib/http-proxy/passes/web-incoming.js#L127) to [here](https://github.com/http-party/node-http-proxy/blob/9b96cd725127a024dabebec6c7ea8c807272223d/lib/http-proxy/common.js#L62) and eventually into `agent` option of the [`request()` method](https://nodejs.org/api/http.html#http_http_request_options_callback) from either `http` or `https` module, which you referenced in you comment.

jginsburgn:
@devoto13 you're right! I missed that path :/. Let me fix this CL.

devoto13 (approved):
LGTM

PR #3642, opened 2021-01-31
fix: ensure that Karma supports running tests on IE 11

devoto13:
They were failing because of unsupported arrow function syntax in two places:
- in karma.spec.js file (only affecting Karma's own tests)
- in socket.io.js file installed from NPM, hence the dependency update (this affected Karma consumers)

Enabled BrowserStack tests on IE 11 to prevent regressions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4297,31 +4297,31 @@
       }
     },
     "engine.io": {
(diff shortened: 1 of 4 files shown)


johnjbarton (approved):
Excellent work, thanks!

PR #3644, opened 2021-01-31
chore: remove unnecessary `init` scripts

devoto13:
These are not necessary anymore and can be safely cleaned up.


devoto13:
Nevermind

PR #3645, opened 2021-02-01
chore(test): adjust test to pass in Safari browser

devoto13:
In Safari stringified Proxy object has ProxyObject as a name, but in other browsers it does not.

Enabled Safari tests on BrowserStack to prevent future regressions.

### DIFF
--- a/test/client/karma.conf.js
+++ b/test/client/karma.conf.js
@@ -15,12 +15,12 @@ const launchers = {
     os: 'Windows',
     os_version: '10'
   },
-  // bs_safari: {
-  //   base: 'BrowserStack',
-  //   browser: 'Safari',
-  //   os: 'OS X',
-  //   os_version: 'Big Sur'
-  // },
+  bs_safari: {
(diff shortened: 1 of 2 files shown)


johnjbarton (approved):
Thanks!

One thing I would like to understand is why these tests did not fail in a way that we could see the results in Travis? Was that all about the client syntax errors?

devoto13:
@johnjbarton The main problem was a syntax error in `socket.io.js` client code. That's why a browser was not able to connect back to the server and communicate any information.

Re how to debug it. While a browser session is active on BrowserStack, one can go into their UI, and click "Start interactive session" (or something) on the job page where one can interact with the browser as it was on their own machine (open dev tools, etc) and figure out if there are any errors in the browser.

devoto13:
Yeah, and this specific change is because I switched from Safari 9.0 (ancient version) to the latest Safari 14.0, which supports `Proxy`. So this test didn't run before, but it does now.

PR #3647, opened 2021-02-03
Bug fix: Report launcher process error when the process exit event is not emitted

chrisbottin:
This PR is to ensure the launcher process error is reported in the logs when the process exit event is not emitted which can happen in some cases.

As per described on this page https://nodejs.org/api/child_process.html#child_process_event_error, "The 'exit' event may or may not fire after an error has occurred. When listening to both the 'exit' and 'error' events, guard against accidentally invoking handler functions multiple times."

**Without this fix**, if the browser binary file is missing, Karma will report the following **if only in Debug mode**:

```
DEBUG [launcher]: Process {{browser_name}} exited with code -1
```
This is not really helpful to the users and if debug log level is not set, they won't see this message.

**With this fix**, if the browser binary file is missing, Karma will **always** report the following:

```
ERROR [launcher]: Cannot start {{browser_name}}
        Can not find the binary {{path_to_binary}}
        Please set env variable {{env_cmd}}
```
This error is clear to the users and always visible

### DIFF
--- a/lib/launchers/process.js
+++ b/lib/launchers/process.js
@@ -93,6 +93,7 @@ function ProcessLauncher (spawn, tempDir, timer, processKillTimeout) {
       } else {
         errorOutput += err.toString()
       }
+      self._onProcessExit(-1, null, errorOutput)
     })
 
     self._process.stderr.on('data', function (errBuff) {
@@ -101,6 +102,10 @@ function ProcessLauncher (spawn, tempDir, timer, processKillTimeout) {
   }
 
   this._onProcessExit = function (code, signal, errorOutput) {
+    if (!self._process) {
+      // Both exit and error events trigger _onProcessExit(), but we only need one cleanup.
+      return
+    }
     log.debug(`Process ${self.name} exited with code ${code} and signal ${signal}`)
 
     let error = null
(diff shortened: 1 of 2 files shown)


lib/launchers/process.js
```diff
@@ -101,6 +102,9 @@ function ProcessLauncher (spawn, tempDir, timer, processKillTimeout) {
   }
 
   this._onProcessExit = function (code, signal, errorOutput) {
+    if (!self._process) {
```
chrisbottin, line 105: This prevents executing this method twice in case both error and exit event are emitted to comply with https://nodejs.org/api/child_process.html#child_process_event_error "The 'exit' event may or may not fire after an error has occurred. When listening to both the 'exit' and 'error' events, guard against accidentally invoking handler functions multiple times."

chrisbottin:
@googlebot I signed it!

lib/launchers/process.js
johnjbarton: Please move this information into a comment in the code, eg

// Both `exit` and `error` events trigger _onProcessExit(), but we only need one cleanup.

johnjbarton (changes requested):
Thanks, one fix up please.

lib/launchers/process.js
chrisbottin: Thanks, comment added.

johnjbarton (approved):

chrisbottin:
@johnjbarton I see the build on master is failing commit message validation due to the fact the pull request ID was appended to the commit message on merge.

Original Commit Message -  65 chars: `fix: report launcher process error when exit event is not emitted`
Merged Commit Message - 73 chars: `fix: report launcher process error when exit event is not emitted (#3647)`
Commit Message Length Limit set at 72 chars on branches and master

What can we do about this?

chrisbottin:
@johnjbarton Potential fix => https://github.com/karma-runner/karma/pull/3650

Issue #3498, opened 2020-05-06
Feature Request/Proposal: Flag for specifying plugin directories/plug n' play support


njradford:
I believe Plugn'Play in Yarn breaks this due to the relative path lookup from node_modules to obtain plugins. If I'm thinking about it correctly, this could be fixed with an option to specify explicitly how to obtain the plugin modules. Or a flag to turn off the relative import. If this sounds right/sense-making/useful I'd be happy to implement a PR.

devoto13:
I'm not sure that Karma can support plugin auto-loading with plug'n'play package manager. So you'll probably have to explicitly require and add all plugins to the config file, e.g.

```
plugins: [
  require('karma-chorme-launcher'),
  require('karma-jasmine')
]
```

Or is there a generic API, which allows to iterate through all available dependencies?

njradford:
I had no idea you could wire them up directly in that way. Thanks @devoto13.

PR #3649, opened 2021-02-06
docs(plugins): add more information about plugins

### LINKED ISSUE #1247: Correct inline plugin example?
On [the Plugins page of the documentation](https://github.com/karma-runner/karma/blob/master/docs/config/05-plugins.md), the following example is presented:

``` javascript
plugins: [
  // Karma will require() these plugins
  'karma-jasmine',
  'karma-chrome-launcher'

  // inlined plugins
  {'framework:xyz', ['factory', factoryFn]},
  require('./plugin-required-from-config')
]
```

Although I haven't needed to write my own inline plugins yet, I doubt the example provided is correct, since it's not even valid JavaScript syntax. Should it be `{ 'framework:xyz' : [ 'factory', factoryFn ] }` instead? A little explanation on that point would be appreciated.

devoto13:
Changes:

- Promote `require('karma-plugin')` form over `'karma-plugin'` form. Former makes it more clear that plugin is imported from an NPM package and it is a regular JS object, there is no magic behind it. This is inspired by #3498 where user is not aware that it is even possible. This also should make it easier with plug'n'play package managers (like Yarn 2).
- Explain that `plugins` array does not activate plugins, but only registers them to clarify https://github.com/karma-runner/karma/issues/1247#issuecomment-312267740.
- Explain the plugin structure, DI and how to build a new plugin.
- Re-arrange "Developing plugins" page to make it easier to add more information about every plugin type. Adding actual information should be done in the separate PRs though.

Fixes #1247

### DIFF
--- a/docs/config/05-plugins.md
+++ b/docs/config/05-plugins.md
@@ -1,11 +1,12 @@
-Karma can be easily extended through plugins.
-In fact, all the existing preprocessors, reporters, browser launchers and frameworks are also plugins.
+Karma can be easily extended through plugins. In fact, all the existing preprocessors, reporters, browser launchers and frameworks are plugins. 
 
-## Installation
+You can install [existing plugins] from NPM or you can write [your own plugins][developing plugins] for Karma.
 
-Karma plugins are NPM modules, so the recommended way to install them are as project dependencies in your `package.json`:
+## Installing Plugins
 
-```javascript
+The recommended way to install plugins is to add them as project dependencies in your `package.json`:
+
+```json
 {
   "devDependencies": {
     "karma": "~0.10",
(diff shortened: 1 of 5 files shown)


docs/dev/05-plugins.md
```diff
@@ -1,61 +1,126 @@
...
+
+// The Karma config is unchaged and is omitted for brevity. See above example
+// for the plugin usage.
+```
+
+Note: Currently, Karma uses [node-di] library as a DI implementation. The library is more powerful than what's documented above, however, the DI implementation may change in the future, so we recommend not to rely on the node-di implementation details.
+
+## Preparing a plugin for publishing
+
+The above examples used a so-called inline plugin form. It is convenient for examples or a tiny ad-hoc plugin, but when publishing a plugin for others to use, it is usually better to export an object from an NPM package to hide extra complexity from the users of the plugins. This also gives a possibility to register multiple plugins by importing a single NPM package.
```
johnjbarton, line 83: I understand that you are trying to give a simple introduction by using the compact inline form, but it means later you need to issue a correction.  I think using the prefered form first with more explanation and no backtracking is better.  In particular I don't like the mixing of consumer and plugin code.

docs/config/05-plugins.md
```diff
@@ -22,26 +23,35 @@ Therefore, a simple way to install a plugin is:
...
+    'karma-chrome-launcher',
+    './my-fancy-plugin'
+  ]
+})
 ```
 
-There are already many [existing plugins]. Of course, you can write [your own plugins] too!
+## Activating Plugins
+
+Adding a plugin to the `plugins` array only makes Karma aware of the plugin, but it does not activate it. Depending on the plugin type you'll need to add a plugin name into `frameworks`, `reporters`, `preprocessors`, `middleware` or `browsers` configuration key to activate it. For the detailed information refer to the corresponding plugin documentation or check out [Developing plugins][developing plugins] guide for more in-depth explanation of how plugins work.
```
johnjbarton, line 54: This is a common source of errors, not doing all of the steps. Can we have a preview earlier, just before "installing" which calls out the three steps?  Then even better, each subsection of the three steps can call out:
`Installing plugins is the first of three steps in their use....`
`Loading plugins is the second of three steps in their use...`
`Activating plugins is the third and final step in their uses...`

johnjbarton (approved):
Awesome, thanks! 

Offered a couple of  suggestions.

docs/dev/05-plugins.md
devoto13: I've updated examples to use a separate file for the plugin code. I still demonstrate the usage of the plugin from the local file, so readers can copy-paste code and run without dealing with multiple NPM packages. Also removed the section on *publishing* as it is no longer needed. Hope this is enough.

docs/config/05-plugins.md
devoto13: I agree with the point, but the current prose is not written as a step-by-step guide and I hesitate to rewrite the whole article to fix this comment. How about doing this in a separate PR later on?

docs/dev/05-plugins.md
```diff
@@ -1,61 +1,126 @@
...
+## Dependency injection
 
-## Launchers
-- example plugins: [karma-chrome-launcher], [karma-sauce-launcher]
-- use naming convention is `karma-*-launcher`
-- use NPM keywords `karma-plugin`, `karma-launcher`
+Karma is assembled using [*dependency injection*](https://en.wikipedia.org/wiki/Dependency_injection). It is important to understand this concept to be able to develop plugins.
 
-## Preprocessors
+On the very high level you can think of Karma as an object where each key (a *DI token*) is mapped to a certain Karma object (a *service*). For example, `config` DI token maps to `Config` instance, which holds current Karma configuration. Plugins can request (or *inject*) various Karma objects by specifying a corresponding DI token. Upon injection a plugin can interact with injected services to implement their functionality.
```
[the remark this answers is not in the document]
devoto13: Can you clarify what exactly do you mean by that?

docs/config/05-plugins.md
johnjbarton: Perhaps my suggestion was not clear.  I'm just suggesting that we have 4 sentences that help users realize that one step is not enough.

docs/dev/05-plugins.md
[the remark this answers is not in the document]
johnjbarton: Just a suggestion that, along with the other great content here, a mention of how the DI table is populated would help demystify the DI.

johnjbarton (approved):
such a great improvement let's get it in

PR #3650, opened 2021-02-08
fix(commitlint): skip task on master

chrisbottin:
The rule between master and branch should be different to take the PR reference in consideration when merging.

---

This PR should be fixing the current build failure on master due to the latest commit message `fix: report launcher process error when exit event is not emitted (#3647)` being more than 72 characters (https://travis-ci.org/github/karma-runner/karma/jobs/757457475).

---

### Problem
The problem is that the same `header-max-length` rule is applied to master and the PR branches and when a PR branch is merged the commit message has the PR reference appended to it making the length longer and potentially breaking the `commitlint` on master

### Proposed Solution

Using different `header-max-length` rules for master and branches.

* `header-max-length` rule on master set to 80 characters to **temporary fix** the build on master and **then submit another PR to  set the header-max-length rule on master set to 72 characters.**
* `header-max-length` rule on branches set to 64 characters (72 - 8 characters for the PR reference)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,9 @@ jobs:
     - node_js: "14"
       script:
         - npm run init
-        - commitlint-travis
+        - if [[ "$TRAVIS_BRANCH" != "master" ]]; then
+            commitlint-travis;
+          fi;
         - npm run lint
         - npm run build:check
         - npm run test:unit


devoto13:
Hm, maybe it is better to simply disable commit message validation on `master`? 🤔

chrisbottin:
> Hm, maybe it is better to simply disable commit message validation on `master`? 🤔

@devoto13 yes we could do that but do you mean permanently? If we are happy with having no `commitlint` running on master then we can just disable it (I can submit another PR) ... otherwise we do need this PR to have different `header-max-length` rules.

Let me know

devoto13:
Yes, permanently. If you think about it, the check is not really actionable: one can't revert or fix the commit in `master` branch. But it's just my opinion. It's up to @johnjbarton to decide which approach to use.

johnjbarton:
Yes @devoto13  is right, I like the no-master solution better.  @chrisbottin can you do it?

chrisbottin:
@johnjbarton done as requested.

johnjbarton (approved):
Thanks, sorry for the delay lost track of this one.

PR #3659, opened 2021-02-26
feat(plugins): add support wildcard config for scoped package plugin

### LINKED ISSUE #3656: Load scoped package plugins by default
### Expected behavior
Karma should be able to load plugins with `@*/karma-*` in their name.

### Actual behavior
Karma will only load plugins with names that start with `karma-*`.
It's incompatible with the [scoped](https://docs.npmjs.com/cli/v7/using-npm/scope) package plugins.
[When finding plugins, Karma doesn't care about subdirectories](https://github.com/karma-runner/karma/blob/99908c3f5b5e48bc2e2158d090ca2e9d514fcb03/lib/plugin.js#L39), so even if you set the `plugins` pattern to like `@*/karma-*`, it will not match anything.

hdmr14:
If you wish load all scoped package plugins by `@*/karma-*` like `karma-*`, karma doesn't handle it.
It'll be modified to `RegExp("^@.*/karma-*")` internally, and karma doesn't care scoped package's sub directories.

In this PR, add features below.

- support plugin name with multiple `*` (replace ALL `*` to `.*`)
  - `@*/karma-*` is now valid
  - `karma-*` also work same as before
  - **however, if anyone already force using multiple `*` (such as `karma-*-foo-.*`), there is a risk of incompatibility**
- when finding plugin, get scoped package name now correctly (before: `@foo`, now: `@foo/bar`)

Fixes #3656

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -36,13 +36,14 @@ function resolve (plugins, emitter) {
 
       log.debug(`Loading ${plugin} from ${pluginDirectory}`)
       fs.readdirSync(pluginDirectory)
-        .flatMap((e) => {
+        .map((e) => {
           const modulePath = path.join(pluginDirectory, e)
           if (e[0] === '@') {
             return fs.readdirSync(modulePath).map((e) => path.join(modulePath, e))
           }
           return modulePath
         })
+        .reduce((a, x) => a.concat(x), [])
         .map((modulePath) => path.relative(pluginDirectory, modulePath))
         .filter((moduleName) => !IGNORED_PACKAGES.includes(moduleName) && regexp.test(moduleName))
         .forEach((pluginName) => requirePlugin(path.join(pluginDirectory, pluginName)))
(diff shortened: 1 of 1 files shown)


hdmr14:
@googlebot I signed it!

devoto13 (commented):
Thanks for the PR, @hdmr14!

It turned out that we didn't have any tests for this functionality, so I've implemented some in https://github.com/devoto13/karma/commit/ee942c5a18a786a0163a26248c277ab9f42b2012 (going to submit those as a separate PR once yours is merged). 

What I found out is that `flatMap` function is [not available in Node 10](https://travis-ci.org/github/karma-runner/karma/jobs/761487241#L279), which Karma still supports. Can you please update your implementation to make sure that it supports Node 10 as well?

Besides this LGTM

hdmr14:
@devoto13 Thanks for review and implement the tests!

PR now supports Node 10(replaced `flatMap` to `map` and `reduce`).

devoto13 (approved):
LGTM

Tests will be added in the follow-up PR: https://github.com/karma-runner/karma/pull/3661

hdmr14:
@googlebot I signed it!

PR #3660, opened 2021-02-28
feat: support asynchronous `config.set()` call in karma.conf.js

### LINKED ISSUE #1259: Calling config.set asynchronously not working
I'm trying to call `config.set` asynchronously in my configuration function. Doing this because we use a tool that finds an available port asynchronously, so we don't have all the configuration at the time the call to our configuration function is made. It appears as though the function never gets called.

npetruzzelli:
With this change, Karma allows to return a `Promise` from a configuration function directly or make such function `async` and thus allows to perform some asynchronous initialization/fetch configuration from somewhere. When the configuration function returns a Promise, Karma will wait until the promise is resolved before it proceeds with the next steps.

Example usage:

```js
module.exports = async (config) => {
  const karmaConfig = await getKarmaConfig("dev");
  
  config.set({
    ...karmaConfig
  });
};
```

For Karma's end-users the change is fully backward compatible. For Karma's programmatic users the change is backward compatible, but the current behavior of `parseConfig()`, `Server`, `runner`, and `stopper` is deprecated and will change in the next major release. New behavior (which supports asynchronous configuration function) can be enabled with a flag. Please refer to the [public API documentation](http://karma-runner.github.io/latest/dev/public-api.html) for more details.

Fixes #1259

### DIFF
--- a/docs/dev/04-public-api.md
+++ b/docs/dev/04-public-api.md
@@ -192,19 +192,6 @@ use of promises.
 const cfg = require('karma').config;
 const path = require('path');
 // Read karma.conf.js, but override port with 1337
-const karmaConfig = cfg.parseConfig(
-  path.resolve('./karma.conf.js'),
-  { port: 1337 },
-  { throwErrors: true }
-)
-```
-
-##### New Behavior with Promise Support
-
-```javascript
-const cfg = require('karma').config;
-const path = require('path');
-// Read karma.conf.js, but override port with 1337
 cfg.parseConfig(
   path.resolve('./karma.conf.js'),
   { port: 1337 },
--- a/docs/dev/04-public-api.md
+++ b/docs/dev/04-public-api.md
@@ -401,7 +401,7 @@ then `parseOptions.throwErrors = true` allows the promise to be rejected
 instead of exiting the process.
 
 
-## `constants`
+## `karma.constants`
 
 ### `constants.VERSION`
 
--- a/docs/dev/04-public-api.md
+++ b/docs/dev/04-public-api.md
@@ -4,7 +4,7 @@ You can, however, call Karma programmatically from your node module. Here is the
 
 ## karma.Server(options, [callback=process.exit])
 
-### Constructor
+### `constructor`
 
 -   **Returns:** `Server` instance.
 
@@ -48,7 +48,7 @@ parseConfig(
 );
 ```
 
-### **server.start()**
+### `server.start()`
 
 Equivalent of `karma start`.
 
@@ -56,7 +56,7 @@ Equivalent of `karma start`.
 server.start()
 ```
 
-### **server.refreshFiles()**
+### `server.refreshFiles()`
 
 Trigger a file list refresh. Returns a promise.
 
@@ -64,7 +64,7 @@ Trigger a file list refresh. Returns a promise.
 server.refreshFiles()
 ```
 
-### **server.refreshFile(path)**
+### `server.refreshFile(path)`
 
 Trigger a file refresh. Returns a promise.
 
@@ -148,7 +148,7 @@ This event gets triggered whenever all the browsers, which belong to a test run,
 
 ## karma.runner
 
-### **runner.run(options, [callback=process.exit])**
+### `runner.run(options, [callback=process.exit])`
 
 -   **Returns:** `EventEmitter`
 
@@ -211,7 +211,7 @@ runner.run({port: 9876}).on('progress', function(data) {
 
 ## karma.stopper
 
-### **stopper.stop(options, [callback=process.exit])**
+### `stopper.stop(options, [callback=process.exit])`
 
 This function will signal a running server to stop. The equivalent of
 `karma stop`.
@@ -264,7 +264,7 @@ the error callback.
 
 ## karma.config
 
-### config.parseConfig([configFilePath], [cliOptions], [parseOptions])
+### `config.parseConfig([configFilePath], [cliOptions], [parseOptions])`
 
 This function will load given config file and returns a filled config object.
 This can be useful if you want to integrate karma into another tool and want to load
@@ -401,60 +401,60 @@ then `parseOptions.throwErrors = true` allows the promise to be rejected
 instead of exiting the process.
 
 
-## karma.constants
+## `constants`
 
-### **constants.VERSION**
+### `constants.VERSION`
 
 The current version of karma
 
-### **constants.DEFAULT_PORT**
+### `constants.DEFAULT_PORT`
 
 The default port used for the karma server
 
-### **constants.DEFAULT_HOSTNAME**
+### `constants.DEFAULT_HOSTNAME`
 
 The default hostname used for the karma server
 
-### **constants.DEFAULT_LISTEN_ADDR**
+### `constants.DEFAULT_LISTEN_ADDR`
 
 The default address use for the karma server to listen on
 
-### **constants.LOG_DISABLE**
+### `constants.LOG_DISABLE`
 
 The value for disabling logs
 
-### **constants.LOG_ERROR**
+### `constants.LOG_ERROR`
 
 The value for the log `error` level
 
-### **constants.LOG_WARN**
+### `constants.LOG_WARN`
 
 The value for the log `warn` level
 
-### **constants.LOG_INFO**
+### `constants.LOG_INFO`
 
 The value for the log `info` level
 
-### **constants.LOG_DEBUG**
+### `constants.LOG_DEBUG`
 
 The value for the log `debug` level
 
-### **constants.LOG_PRIORITIES**
+### `constants.LOG_PRIORITIES`
 
 An array of log levels in descending order, i.e. `LOG_DISABLE`, `LOG_ERROR`, `LOG_WARN`, `LOG_INFO`, and `LOG_DEBUG`
 
-### **constants.COLOR_PATTERN**
+### `constants.COLOR_PATTERN`
 
 The default color pattern for log output
 
-### **constants.NO_COLOR_PATTERN**
+### `constants.NO_COLOR_PATTERN`
 
 The default pattern for log output without color
 
-### **constants.CONSOLE_APPENDER**
+### `constants.CONSOLE_APPENDER`
 
 The default console appender
 
-### **constants.EXIT_CODE**
+### `constants.EXIT_CODE`
 
 The exit code
--- a/docs/dev/04-public-api.md
+++ b/docs/dev/04-public-api.md
@@ -6,6 +6,17 @@ You can, however, call Karma programmatically from your node module. Here is the
 
 ### Constructor
 
+-   **Returns:** `Server` instance.
+
+#### Usage
+
+Notice the capital 'S' on  `require('karma').Server`.
+
+##### Deprecated Behavior
+
+The following still works, but the way it behaves is deprecated and will be
+changed in a future major version.
+
 ```javascript
 var Server = require('karma').Server
 var karmaConfig = {port: 9876}
@@ -15,7 +26,27 @@ var server = new Server(karmaConfig, function(exitCode) {
 })
 ```
 
-Notice the capital 'S' on  `require('karma').Server`.
+##### New Behavior
+
+```javascript
+const karma = require('karma')
+const parseConfig = karma.config.parseConfig
+const Server = karma.Server
+
+parseConfig(
+  null,
+  { port: 9876 },
+  { promiseConfig: true, throwErrors: true }
+).then(
+  (karmaConfig) => {
+    const server = new Server(karmaConfig, function doneCallback(exitCode) {
+      console.log('Karma has exited with ' + exitCode)
+      process.exit(exitCode)
+    })
+  },
+  (rejectReason) => { /* respond to the rejection reason error */ }
+);
+```
 
 ### **server.start()**
 
--- a/docs/dev/04-public-api.md
+++ b/docs/dev/04-public-api.md
@@ -156,19 +156,158 @@ stopper.stop({port: 9876}, function(exitCode) {
 })
 ```
 
-## karma.config.parseConfig([configFilePath], [cliOptions])
+## karma.config
+
+### config.parseConfig([configFilePath], [cliOptions], [parseOptions])
 
 This function will load given config file and returns a filled config object.
 This can be useful if you want to integrate karma into another tool and want to load
 the karma config while honoring the karma defaults. For example, the [stryker-karma-runner](https://github.com/stryker-mutator/stryker-karma-runner)
 uses this to load your karma configuration and use that in the stryker configuration.
 
+#### Usage
+
+##### Deprecated Behavior
+
+The following still works, but the way it behaves is deprecated and will be
+changed in a future major version.
+
 ```javascript
 const cfg = require('karma').config;
 const path = require('path');
 // Read karma.conf.js, but override port with 1337
-const karmaConfig = cfg.parseConfig(path.resolve('./karma.conf.js'), { port: 1337 } );
+const karmaConfig = cfg.parseConfig(
+  path.resolve('./karma.conf.js'),
+  { port: 1337 }
+);
 ```
+
+The new behavior in the future will involve throwing exceptions instead of
+exiting the process and aynchronous config files will be supported through the
+use of promises.
+
+##### New Behavior
+
+```javascript
+const cfg = require('karma').config;
+const path = require('path');
+// Read karma.conf.js, but override port with 1337
+const karmaConfig = cfg.parseConfig(
+  path.resolve('./karma.conf.js'),
+  { port: 1337 },
+  { throwErrors: true }
+)
+```
+
+##### New Behavior with Promise Support
+
+```javascript
+const cfg = require('karma').config;
+const path = require('path');
+// Read karma.conf.js, but override port with 1337
+cfg.parseConfig(
+  path.resolve('./karma.conf.js'),
+  { port: 1337 },
+  { promiseConfig: true, throwErrors: true }
+).then(
+  (karmaConfig) => { /* use the config with the public API */ },
+  (rejectReason) => { /* respond to the rejection reason error */ }
+);
+```
+
+
+#### `configFilePath` argument
+
+-   **Type:** String | `null` | `undefined`
+-   **Default Value:** `undefined`
+
+A string representing a file system path pointing to the config file whose
+default export is a function that will be used to set Karma configuration
+options. This function will be passed an instance of the `Config` class as its
+first argument. If this option is not provided, then only the options provided
+by the `cliOptions` argument will be set.
+
+-   JavaScript must use CommonJS modules.
+-   ECMAScript modules are not currently supported by Karma when using
+    JavaScript.
+    -   Other formats, such as TypeScript, may support ECMAScript modules.
+
+
+#### `cliOptions` argument
+
+-   **Type:** Object | `null` | `undefined`
+-   **Default Value:** `undefined`
+
+An object whose values will take priority over options set in the config file.
+The config object passed to function exported by the config file will already
+have these options applied. Any changes the config file makes to these options
+will effectively be ignored in the final configuration.
+
+Supports all the same options as the config file and is applied using the same
+`config.set()` method.
+
+The expected source of this argument is parsed command line options, but 
+programatic users may construct this object or leave it out entirely.
+
+
+#### `parseOptions` argument
+
+-   **Type:** Object | `null` | `undefined`
+-   **Default Value:** `undefined`
+
+`parseOptions` is an object whose properties are configuration options that
+allow additional control over parsing and opt-in access to new behaviors or
+features.
+
+These options are only related to parsing configuration files and object and are
+not related to the configuration of Karma itself.
+
+
+##### `parseOptions.promiseConfig` option
+
+-   **Type:** Boolean
+-   **Default Value:** `false`
+
+When `parseOptions.promiseConfig === true`, then `parseConfig` will return a
+promise instead of a configuration object.
+
+When this option is `true`, then the function exported by the config file may
+return a promise. The resolution of that promise indicates that all asynchronous
+activity has been completed. Internally, the resolved/fulfilled value is 
+ignored. As with synchronous usage, all changes to the config object must be
+done with the `config.set()` method.
+
+If the function exported by the config file does not return a promise, then
+parsing is completed and an immediately fulfilled promise is returned.
+
+Whether the function exported by the config file returns a promise or not, the
+promise returned by `parseConfig()` will resolve with a parsed configuration
+object, an instance of the `Config` class, as the value.
+
+_**In most cases, `parseOptions.throwErrors = true` should also be set. This
+disables process exiting and allows errors to result in rejected promises.**_
+
+
+##### `parseOptions.throwErrors` option
+
+-   **Type:** Boolean
+-   **Default Value:** `false`
+
+In the past, `parseConfig()` would call `process.exit(exitCode)` when it
+encountered a critical failure. This meant that your own code had no way of
+responding to failures before the Node.js process exited.
+
+By passing `parseOptions.throwErrors = true`, `parseConfig()` will disable
+process exiting.
+
+For synchronous usage, it will throw an exception instead of exiting the
+process. Your code can then catch the exception and respond how ever it needs
+to.
+
+If the asynchronous API (`parseOptions.promiseConfig = true`) is being used,
+then `parseOptions.throwErrors = true` allows the promise to be rejected
+instead of exiting the process.
+
 
 ## karma.constants
 
--- a/docs/dev/04-public-api.md
+++ b/docs/dev/04-public-api.md
@@ -162,8 +162,7 @@ stopper.stop({port: 9876}, function(exitCode) {
 
 This function will load given config file and returns a filled config object.
 This can be useful if you want to integrate karma into another tool and want to load
-the karma config while honoring the karma defaults. For example, the [stryker-karma-runner](https://github.com/stryker-mutator/stryker-karma-runner)
-uses this to load your karma configuration and use that in the stryker configuration.
+the karma config while honoring the karma defaults.
 
 #### Usage
 
--- a/docs/dev/04-public-api.md
+++ b/docs/dev/04-public-api.md
@@ -150,7 +150,16 @@ This event gets triggered whenever all the browsers, which belong to a test run,
 
 ### **runner.run(options, [callback=process.exit])**
 
+-   **Returns:** `EventEmitter`
+
 The equivalent of `karma run`.
+
+#### Usage
+
+##### Deprecated Behavior
+
+The following still works, but the way it behaves is deprecated and will be
+changed in a future major version.
 
 ```javascript
 var runner = require('karma').runner
@@ -159,6 +168,35 @@ runner.run({port: 9876}, function(exitCode) {
   process.exit(exitCode)
 })
 ```
+
+##### New Behavior
+
+```javascript
+const karma = require('karma')
+
+karma.config.parseConfig(
+  null,
+  { port: 9876 },
+  { promiseConfig: true, throwErrors: true }
+).then(
+  (karmaConfig) => {
+    karma.runner.run(karmaConfig, function doneCallback(exitCode, possibleErrorCode) {
+      console.log('Karma has exited with ' + exitCode)
+      process.exit(exitCode)
+    })
+  },
+  (rejectReason) => { /* respond to the rejection reason error */ }
+);
+```
+
+#### `callback` argument
+
+The callback receives the exit code as the first argument.
+
+If there is an error, the error code will be provided as the second parameter to
+the error callback.
+
+#### runner Events
 
 `runner.run()` returns an `EventEmitter` which emits a `progress` event passing
 the reporter output as a `Buffer` object.
@@ -175,7 +213,15 @@ runner.run({port: 9876}).on('progress', function(data) {
 
 ### **stopper.stop(options, [callback=process.exit])**
 
-This function will signal a running server to stop.  The equivalent of `karma stop`.  
+This function will signal a running server to stop. The equivalent of
+`karma stop`.
+
+#### Usage
+
+##### Deprecated Behavior
+
+The following still works, but the way it behaves is deprecated and will be
+changed in a future major version.
 
 ```javascript
 var stopper = require('karma').stopper
@@ -186,6 +232,35 @@ stopper.stop({port: 9876}, function(exitCode) {
   process.exit(exitCode)
 })
 ```
+
+##### New Behavior
+
+```javascript
+const karma = require('karma')
+
+karma.config.parseConfig(
+  null,
+  { port: 9876 },
+  { promiseConfig: true, throwErrors: true }
+).then(
+  (karmaConfig) => {
+    karma.stopper.stop(karmaConfig, function doneCallback(exitCode, possibleErrorCode) {
+      if (exitCode === 0) {
+        console.log('Server stop as initiated')
+      }
+      process.exit(exitCode)
+    })
+  },
+  (rejectReason) => { /* respond to the rejection reason error */ }
+);
+```
+
+#### `callback` argument
+
+The callback receives the exit code as the first argument.
+
+If there is an error, the error code will be provided as the second parameter to
+the error callback.
 
 ## karma.config
 
@@ -383,7 +458,3 @@ The default console appender
 ### **constants.EXIT_CODE**
 
 The exit code
-
-## Callback function notes
-
-- If there is an error, the error code will be provided as the second parameter to the error callback.
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -304,8 +304,9 @@ exports.run = () => {
           require('./stopper').stop(config)
           break
       }
-    }, function onKarmaConfigRejected (reason) {
-      console.error(reason) // TODO: configure a CLI Logger?
+    }, function onKarmaConfigRejected (/* ignoredReason */) {
+      // The reject reason isn't used to log a message since parseConfig already
+      // calls a configured logger method with an almost identical message.
 
       // The `run` function is a private application, not a public API. We don't
       // need to worry about process.exit vs throw vs promise rejection here.
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -7,6 +7,7 @@ const fs = require('graceful-fs')
 const Server = require('./server')
 const helper = require('./helper')
 const constant = require('./constants')
+const cfg = require('./config')
 
 function processArgs (argv, options, fs, path) {
   Object.getOwnPropertyNames(argv).forEach(function (name) {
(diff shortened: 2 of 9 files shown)


npetruzzelli:
@devoto13 - I'm not sure if I will be able to see this PR through to completion (quickly), but I was able to at least start the work so that the conversation/review can start.

I can't help but to "throw my hat into the ring" as it were. I initially opened #3631 / #3635 because I was working on a personal module that would support ES Modules and Promises. It still needs a lot of work, including the addition of tests (which is why it is only v `0.1.0`). It is significantly overbuilt for what Karma would want or need, but I'll link it here in case you want to have a look: 
https://github.com/promise-file/promise-file/tree/v0.1.0/packages/promise-file-karma-config

npetruzzelli:
Would this need the ability for users to pass their own promise implementation, or will the default Promise meet the needs of all environments we expect this to serve in?

npetruzzelli:
ASIDE: Should the `process` and `parseClientArgs` functions from `cli.js` be made a part of the public API for anyone using `parseConfig` and the public API methods directly? The intent being that they may want to parse command line arguments in a way that is consistent with Karma (this deserves its own issue for discussion prior to a PR and is not in the scope of this PR.)

npetruzzelli:
It still needs new tests to be written, but I am marking this as ready for review to address feedback prior to putting time into tests so that I can avoid writing tests that may be made obsolete by review feedback.

devoto13:
> I'm not sure if I will be able to see this PR through to completion (quickly), but I was able to at least start the work so that the conversation/review can start.

No worries, take your time. And thanks for the work you're putting into implementing this!

> Would this need the ability for users to pass their own promise implementation, or will the default Promise meet the needs of all environments we expect this to serve in?

Default implementation is enough.

> Should the process and parseClientArgs functions from cli.js be made a part of the public API for anyone using parseConfig and the public API methods directly? The intent being that they may want to parse command line arguments in a way that is consistent with Karma (this deserves its own issue for discussion prior to a PR and is not in the scope of this PR.)

I'm not sure how useful is this. Better have strong real use case for this and discuss it before doing anything.

> It still needs new tests to be written, but I am marking this as ready for review to address feedback prior to putting time into tests so that I can avoid writing tests that may be made obsolete by review feedback.

I'll take a look right away.

lib/config.js
```diff
@@ -352,11 +352,17 @@ const CONFIG_SYNTAX_HELP = '  module.exports = function(config) {\n' +
   '  };\n'
 
 function parseConfig (configFilePath, cliOptions, parseOptions) {
+  const promiseConfig = parseOptions && parseOptions.promiseConfig === true
+  const throwErrors = parseOptions && parseOptions.throwErrors === true
   function fail () {
     log.error(...arguments)
-    if (parseOptions && parseOptions.throwErrors === true) {
+    if (throwErrors === true || promiseConfig === true) {
```
devoto13, line 359: IMO it is cleaner to drop `|| promiseConfig === true` part, so flags are independent and one does not imply the other.

lib/cli.js
```diff
@@ -278,26 +279,36 @@ exports.process = () => {
...
-      break
-    case 'init':
-      require('./init').init(config)
-      break
-    case 'completion':
-      require('./completion').completion(config)
-      break
+  const cliOptions = exports.process()
+  const cmd = cliOptions.cmd // prevent config from changing the command
+  const cmdSupportsConfigPromise =
```
devoto13, line 284: nit: cmdSupportsConfigPromise -> needsConfig

It's not so much about the support, but more like whether this command needs to parse the config file.

lib/cli.js
```diff
@@ -278,26 +279,36 @@ exports.process = () => {
...
+    switch (cmd) {
+      case 'init':
+        require('./init').init(cliOptions)
+        break
+      case 'completion':
+        require('./completion').completion(cliOptions)
+        break
+    }
+  } else {
+    cfg.parseConfig(cliOptions.configFile, cliOptions, { promiseConfig: true })
```
devoto13, line 296: Let's set `throwErrors: true`, add a `.catch()` and `process.exit(1)` in the catch block. The idea is that `throwErrors: false` is deprecated and we want to remove it to cleanup the parseConfig() implementation in the next major. Hence we should not rely onto the deprecated behaviour ourselves.

lib/server.js
```diff
@@ -55,22 +55,36 @@ function createSocketIoServer (webServer, executor, config) {
...
-      config = cfg.parseConfig(cliOptions.configFile, cliOptions, { throwErrors: true })
-    } catch (parseConfigError) {
-      // TODO: change how `done` falls back to exit in next major version
-      //  SEE: https://github.com/karma-runner/karma/pull/3635#discussion_r565399378
-      (done || process.exit)(1)
-      return
+    if (cliOptionsOrConfig instanceof cfg.Config) {
+      config = cliOptionsOrConfig
+    } else {
+      try {
```
devoto13, line 73: Can you also log a deprecation warning in the `else` branches in server, runner and stopper?

Something like:

Passing CLI options to the {component} is deprecated and will be removed in the next major release. Use `parseConfig(..., {throwError: true, promiseConfig: true})` to prepare an instance of the `Config` class and pass it instead of raw CLI options.

devoto13 (commented):
It looks great @npetruzzelli! Thanks a lot for doing this.

I left couple of tiny comments. Let me know if they make sense to you.

I've also checked point (4). Looking at the code, I don't think we need to do anything about it.

lib/config.js
npetruzzelli: That's fine. Throwing vs rejecting is still mutually exclusive, but it can easily be worked with.

My instinct is to avoid writing the same error twice.  do I self contain the thing? or do I avoid maintaining the same thing in multiple locations? Here is what I'm thinking:

```js
  function fail () {
    log.error(...arguments)
    const errorMessage = Array.from(arguments).join(' ')
    const err = new Error(errorMessage)
    if (promiseConfig === true) {
      return Promise.reject(err)
    } else if (throwErrors === true) {
      throw err
    } else {
```

lib/cli.js
npetruzzelli: I'll see your nit pick and raise you one! This seems fine to me, but it would make more sense to me to explicitly say what needs the configuration.

How about `cmdNeedsConfig`?

devoto13: Sounds good!

Consider also swapping if/else branches to avoid inverting the condition.

lib/config.js
devoto13: I think the logic should be like this:

```
if (throwErrors) {
  if (promiseConfig) return Promise.reject(err)
  else throw err
} else {
  process.exit(1)
}
```

E.g. promiseConfig == true should not imply throwErrors == true (which controls process.exit behavior). Does it make sense?

npetruzzelli: I am ok with using that logic, since it accomplishes the same thing. The difference in the end result is requiring `throwErrors` to be set in order to use promises.

> promiseConfig == true should not imply throwErrors == true

I agree. I was trying to go for that but my thinking was following a different path.

<details>

Throwing and promising are both things that should disable `process.exit`.

Promises work on fulfillment and rejection. If we don't reject with errors, then we are breaking the expectation the consumer has of using promises.

I guess you could say I am concerned about scenarios where `{throwsErrors: false /* or undefined */, promiseConfig: true}` occur. Instead of a rejection, the process could exit on failure. When using promises, I would not expect that kind of behavior. Since promising configs is a new feature, it seems intuitive to me to avoid a `process.exit` entirely with it.

I might be getting hung up on the choice of words here. `throwsErrrors` vs a hypothetical `dontExit`, which I probably wouldn't be as worried about as it would be `throwsErrors` with a different name.

Since `throwsErrors` is logic that will be cleaned up in a future major release, I'm ok with using the code you suggested above. It absolutely meets the internal and public API goals of this PR.

---

ASIDE: I can see `promiseConfig` staying in the future/cleaned up code (in some form) so that public API users can chose between sync and async. Users would have to accept that sync won't support async configs. (Not sure how I feel about that.) An alternative would be to offer 2 functions (sync and async) and remove the option entirely from the public API.

```js
function parseConfigBase(configFilePath, cliOptions, parseOptions, promiseConfig) {
  // The same function we have now, only not exported
  // Including code review changes
}

// This would be the sync function for backwards compatibility
// otherwise I would have suggested that this be `parseConfigSync` or `parseConfig.sync`
function parseConfig(configFilePath, cliOptions, parseOptions) {
  return parseConfigBase(configFilePath, cliOptions, parseOptions, false)
}

// This name makes it clear that async uses promises instead of the callback pattern
parseConfig.promise = function promiseParsedConfig(configFilePath, cliOptions, parseOptions) {
  return parseConfigBase(configFilePath, cliOptions, parseOptions, true)
}
```

This aside is more thinking aloud for the future and should not affect the direction taken by this PR.

</details>

TL;DR : I'll make the change you suggested.

devoto13: Thanks for detailing it out. The alternative design makes total sense as well.

However, I expect both of these flags to be removed in the future for cleaner code and API simplicity over a hypothetical use case and that's why I suggested this change.

lib/cli.js
npetruzzelli: Are we catching just the parseConfig? or the entire switch?

<details>
<summary>Option 1: Catch Parse Config and All Commands</summary>

```js
exports.run = () => {
  const cliOptions = exports.process()
  const cmd = cliOptions.cmd // prevent config from changing the command
  const cmdNeedsConfig = cmd === 'start' || cmd === 'run' || cmd === 'stop'
  function failCommand (err) {
    console.error(err) // TODO: configure a CLI Logger?

    // The `run` function is a private application, not a public API. We don't
    // need to worry about process.exit vs throw vs promise rejection here.
    process.exit(1)
  }
  if (cmdNeedsConfig) {
    cfg.parseConfig(
      cliOptions.configFile,
      cliOptions,
      {
        promiseConfig: true,
        throwErrors: true
      }
    ).then(function onKarmaConfigFulfilled (config) {
      switch (cmd) {
        case 'start':
          new Server(config).start()
          break
        case 'run':
          require('./runner')
            .run(config)
            .on('progress', printRunnerProgress)
          break
        case 'stop':
          require('./stopper').stop(config)
          break
      }
    }).catch(function onCommandFailed (reason) {
      return failCommand(reason)
    })
  } else {
    try {
      switch (cmd) {
        case 'init':
          require('./init').init(cliOptions)
          break
        case 'completion':
          require('./completion').completion(cliOptions)
          break
      }
    } catch (commandException) {
      return failCommand(commandException)
    }
  }
}
```

</details>

<details>
<summary>Option 2: Catch Parse Config Only</summary>

```js
exports.run = () => {
  const cliOptions = exports.process()
  const cmd = cliOptions.cmd // prevent config from changing the command
  const cmdNeedsConfig = cmd === 'start' || cmd === 'run' || cmd === 'stop'
  if (cmdNeedsConfig) {
    cfg.parseConfig(
      cliOptions.configFile,
      cliOptions,
      {
        promiseConfig: true,
        throwErrors: true
      }
    ).then(function onKarmaConfigFulfilled (config) {
      switch (cmd) {
        case 'start':
          new Server(config).start()
          break
        case 'run':
          require('./runner')
            .run(config)
            .on('progress', printRunnerProgress)
          break
        case 'stop':
          require('./stopper').stop(config)
          break
      }
    }, function onKarmaConfigRejected (reason) {
      console.error(reason) // TODO: configure a CLI Logger?

      // The `run` function is a private application, not a public API. We don't
      // need to worry about process.exit vs throw vs promise rejection here.
      process.exit(1)
    })
  } else {
    switch (cmd) {
      case 'init':
        require('./init').init(cliOptions)
        break
      case 'completion':
        require('./completion').completion(cliOptions)
        break
    }
  }
}
```

</details>

devoto13: Option 2.

> console.error(reason) // TODO: configure a CLI Logger?

Oh, now I remember why I've added item (4) to the list! Because we need to call `setupFromConfig()` to configure logger before we actually attempt to parse config. So we should call it here and guard calls in server/runner/stopper by the check for the backward compatibility.

And we should also expose it on the public API for API consumers. Maybe under a different name. I'll get back on this part.

lib/config.js
```diff
@@ -352,11 +352,17 @@ const CONFIG_SYNTAX_HELP = '  module.exports = function(config) {\n' +
   '  };\n'
 
 function parseConfig (configFilePath, cliOptions, parseOptions) {
+  const promiseConfig = parseOptions && parseOptions.promiseConfig === true
```
npetruzzelli, line 355: Should this change to:

```js
const throwErrors = parseOptions && parseOptions.throwErrors === true
const promiseConfig = throwErrors && parseOptions && parseOptions.promiseConfig === true
```
to have the rest of the function behave the same way as `fail()`?

lib/config.js
```diff
@@ -411,34 +417,59 @@ function parseConfig (configFilePath, cliOptions, parseOptions) {
...
-  return normalizeConfig(config, configFilePath)
+    return normalizeConfig(config, configFilePath)
+  }
+  const returnIsThenable = configModuleReturn != null && typeof configModuleReturn === 'object' && typeof configModuleReturn.then === 'function'
+  if (returnIsThenable) {
+    if (promiseConfig !== true) {
+      const errorMessage =
+        'The `parseOptions.promiseConfig` option must be set to `true` to ' +
+        'enable promise return values from configuration files. ' +
+        'Example: `parseConfig(path, cliOptions, { promiseConfig: true })`'
```
npetruzzelli, line 456: Does this message need to include `throwErrors: true`?

devoto13: Yes, let's do it to show the recommended set of flags to the user.

lib/config.js
devoto13: No, why?

`throwErrors` chooses between `process.exit()` and throw/reject.  
`promiseConfig` decides whether the function returns synchronously or returns a Promise.

lib/config.js
```diff
@@ -411,34 +417,59 @@ function parseConfig (configFilePath, cliOptions, parseOptions) {
...
+      return fail(errorMessage)
+    }
+    return configModuleReturn.then(function onKarmaConfigModuleFulfilled (/* ignoredResolutionValue */) {
```
devoto13, line 459:
```suggestion
    return Promise.resolve(configModuleReturn).then(function onKarmaConfigModuleFulfilled (/* ignoredResolutionValue */) {
```

as otherwise, it crashes when `promiseConfig`, but config is not async.

lib/config.js
devoto13: Actually, nevermind. Let's not change it to stay focused on the error message.

lib/cli.js
devoto13: I actually came up with a better idea. When `promiseConfig == true`, make a call to `setupFromConfig()` at the beginning of `parseConfig()` function. This way we still get logs, if something fails during config parsing and we don't need to expose any new public API.

And guard calls in server/runner/stopper by the `instanceof Config` check for the backward compatibility. They will be cleaned up in the next major together with the other old parsing logic.

lib/config.js
npetruzzelli: I understand (and mostly support) the logic of option A (`throwErrors`) controls process exiting and option B (`promiseConfig`) controls sync vs async.

My concern is that behavior would be surprising in the following scenario:
`{throwErrors: false, promiseConfig: true}`

-   (expected) critical/internal failures exit the process
-   (expected) non-critical failures return rejected promises
-   (unexpected, multiple behaviors) function exported by the config file:
    -   exceptions thrown are treated as critical failures and exit the process
    -   if the promise is rejected, we return that rejected promise instead of exiting the process

<details>

Other branches for reference:

-   `if (promiseConfig === false && configFileReturnsPromise === true)`
    -   critical failure (side thought: **should returned promises be silently ignored?** The existing sync behavior ignores it anyway. My current code tries to detect intent, which may be feature creep.)
-   `if (promiseConfig === true && configFileReturnsPromise === true)`
    -   success results in a promise that resolves to a `Config` instance
    -   if the promise returned by the config file is rejected, that rejection is carried through to the promise that `parseConfig` returns
-   `else if(promiseConfig === true && configFileReturnsPromise === false)`
    -   config is finalized and a resolved promise is returned with the `Config` instance as the fulfilled value.
    -   any errors thrown in `finalizeConfig` (and by extension `normalizeConfig`) are returned as a promise rejected with the exception as the reason
-   `if (promiseConfig === false && configFileReturnsPromise === false)`
    -   normal/existing sync behavior

</details>

I feel like the function exported by the config file should be treated one way or the other, but not both.

---

**Option 1**: My first approach to this was to have critical failures return a rejected promise with `promiseConfig: true` regardless of the value of `throwErrors`. This way thrown exceptions and rejected promises both return a rejected promise. It is up to the consumer of the public API to catch that rejection. Additionally, it made sense to me (personally) to not exit the process when using a method that you expect to return a promise.

This was rejected in a previous code review because one option (`promiseConfig`) implied the other (`throwErrors`) and it didn't seem like a clean solution.

---

**Option 2**: One way of approaching this is detecting `{throwErrors: false, promiseConfig: true}` and critically failing with a message stating the `throwsErrors` option must be `true` in order to use `promiseConfig`.

---

**Option 3**: Another approach could be to catch the promise rejection internally and exit the process, using the rejection reason to build the failure message.

---

**Option 4**: Leave the code as is

---

I'd personally like to avoid option 4, but I also don't want it to jeopardize the rest of the PR as it otherwise works well and the implementation is expected to change in the next major release anyway.

lib/config.js
npetruzzelli: This should be handled by: 
https://github.com/karma-runner/karma/blob/5d92865dd3d61c9570bd85102e56352d9dc0715c/lib/config.js#L450-L451

and then by:
https://github.com/karma-runner/karma/blob/5d92865dd3d61c9570bd85102e56352d9dc0715c/lib/config.js#L462-L472

Though I may have made this more complicated than it needed to be (as I just mentioned in a different comment). In hindsight, I think there may be a more clear way.

```diff
-  const returnIsThenable = configModuleReturn != null && typeof configModuleReturn === 'object' && typeof configModuleReturn.then === 'function'
-  if (returnIsThenable) {
-    if (promiseConfig !== true) {
-      const errorMessage =
-        'The `parseOptions.promiseConfig` option must be set to `true` to ' +
-        'enable promise return values from configuration files. ' +
-        'Example: `parseConfig(path, cliOptions, { promiseConfig: true })`'
-      return fail(errorMessage)
-    }
-    return configModuleReturn.then(function onKarmaConfigModuleFulfilled (/* ignoredResolutionValue */) {
-      return finalizeConfig(config)
-    })
-  } else {
-    if (promiseConfig) {
+  if (promiseConfig) {
+    const returnIsThenable = configModuleReturn != null && typeof configModuleReturn === 'object' && typeof configModuleReturn.then === 'function'
+    if (returnIsThenable) {
+      return configModuleReturn.then(function onKarmaConfigModuleFulfilled (/* ignoredResolutionValue */) {
+        return finalizeConfig(config)
+      })
+    } else {
       try {
         return Promise.resolve(finalizeConfig(config))
       } catch (exception) {
         return Promise.reject(exception)
       }
     } else {
       return finalizeConfig(config)
     }
   }
 }
```

npetruzzelli: -   the config file function _is not required to return a promise_ in order for `parseConfig` to return a promise.
-   if a promise is not returned by the function, the config is handled the same way as synchronous parsing and an immediately resolved promise is returned.

A promise resolving just tells us that the asynchronous code is done. We don't use the return/fulfillment values otherwise since we already have the config object and expect it to be modified via `config.set`.

---

I guess this raises another question. Assumptions I made in the first and second passes at the code may not be in line with what you were expecting.

Do we want to require the config file to have a promise as a return value?

devoto13: Re original suggestion, I guess I did something wrong. I've retested now and everything seems to work as intended. So never mind. Thanks for the explanation.

> Though I may have made this more complicated than it needed to be (as I just mentioned in a different comment). In hindsight, I think there may be a more clear way.

No, I think the current code is good and it helps the user to resolve the problem.

> Do we want to require the config file to have a promise as a return value?

No, we don't want to require it to return a promise. The synchronous config should work with the `promiseConfig == true`.

I'll resolve this thread as there is nothing to do here and respond in another one in a minute.

lib/config.js
devoto13: I think we should go with option 3. The only change we need to do is below and that everything is consistent, right?

```diff
    return configModuleReturn.then(
      () => finalizeConfig(config),
      (e) => fail('Error in config file!\n', e)
    )
```

devoto13: Oh, and the below as well, to capture the `promiseConfig == true` with the sync config function.

```
try {
        return Promise.resolve(finalizeConfig(config))
      } catch (e) {
        return fail('Error in config file!\n', e)
      }
```

lib/cli.js
npetruzzelli: I'm leaning towards Option 2 or Option 4

What do you think?

<details>
<summary>Option 1: What?</summary>

I don't know exactly how log4js works, so I'm not 100% sold on what I am about to suggest, but I took your previous comment and thought of the following:

Instead of configuring it at the start of the `parseConfig` function, why not configure it at the start of the `fail` function, since that is the point that we know that the other logger setup code won't be run.

```js
// This guarding may not be needed. It doesn't look like the edge case is
// possible with the _current_ code.
let loggerHasBeenSetup = false

/**
 * Allow future calls to `parseConfig` to configure the logger based on a
 * successfully created `Config` instance.
 */
function resetParseConfigLogger () {
  loggerHasBeenSetup = false
}

/**
 * Setup up the logger once in the life time of `parseConfig`. If options aren't
 * provided, it uses the same defaults as the a `Config` class instance.
 *
 * This prevents the edge case of the logger being set up at the end of
 * `finalizeConfig` but before `normalizeConfig`, and then an error occurs
 * which, if interpreted as a critical failure, would then try to configure the
 * logger again within the `fail` function.
 */
function setupLoggerInParseConfig (loggerOptions) {
  if (!loggerHasBeenSetup) {
    if (!loggerOptions) {
      loggerOptions = {}
    }

    logger.setupFromConfig(
      {
        colors: loggerOptions.colors,
        logLevel: loggerOptions.logLevel
      },
      loggerOptions.appenders
    )
    loggerHasBeenSetup = true
  }
}
```

```diff

 function parseConfig (configFilePath, cliOptions, parseOptions) {
   const promiseConfig = parseOptions && parseOptions.promiseConfig === true
   const throwErrors = parseOptions && parseOptions.throwErrors === true
   function fail () {
+    setupLoggerInParseConfig()
     log.error(...arguments)
+
+    // This execution is about to end, so enable future executions to configure
+    // the logger.
+    resetParseConfigLogger()
     // ...

   }
   // ...
   
   function finalizeConfig (config) {
     // ...

     // configure the logger as soon as we can
-    logger.setup(config.logLevel, config.colors, config.loggers)
+    setupLoggerInParseConfig({
+      appenders: config.loggers,
+      colors: config.colors,
+      logLevel: config.logLevel
+    })

     log.debug(configFilePath ? `Loading config ${configFilePath}` : 'No config file specified.')

     return normalizeConfig(config, configFilePath)
   }
   // ...
 }
```

A successfully set up config prevents the logger from being configured again unless parse config is called again and then it encounters an error. This may be a breaking. I wrote it and I find it confusing.

The fact that I don't 100% remember what I was going for isn't a good sign for this option. Maybe the coffee hadn't set in yet.

---

</details>

<details>
<summary>Option 2: Configure logger in <code>fail()</code></summary>

```diff
   function fail () {
+    // `setupFromConfig` provides defaults for `colors` and `logLevel`.
+    // `setup` provides defaults for `appenders`
+    // The first argument MUST BE an object
+    logger.setupFromConfig({})
     log.error(...arguments)
     if (throwErrors) {
```

---

</details>


<details>
<summary>Option 3: When <code>promiseConfig == true</code> OR <code>throwErrors == true</code>, make a call to <code>setupFromConfig()</code> at the beginning of <code>parseConfig()</code> function.</summary>

The thought here being that non-promise failures will want a configured logger too.

```diff
 function parseConfig (configFilePath, cliOptions, parseOptions) {
   const promiseConfig = parseOptions && parseOptions.promiseConfig === true
   const throwErrors = parseOptions && parseOptions.throwErrors === true
+  const shouldSetupLoggerEarly = promiseConfig || throwErrors
+  if (shouldSetupLoggerEarly) {
+    // `setupFromConfig` provides defaults for `colors` and `logLevel`.
+    // `setup` provides defaults for `appenders`
+    // The first argument MUST BE an object
+    logger.setupFromConfig({})
+  }
   function fail () {
```

---

</details>


<details>
<summary>Option 4: When <code>promiseConfig == true</code>, make a call to <code>setupFromConfig()</code> at the beginning of <code>parseConfig()</code> function.</summary>

A first pass at what you mentioned above without modification.

```diff
 function parseConfig (configFilePath, cliOptions, parseOptions) {
   const promiseConfig = parseOptions && parseOptions.promiseConfig === true
   const throwErrors = parseOptions && parseOptions.throwErrors === true
+  const shouldSetupLoggerEarly = promiseConfig
+  if (shouldSetupLoggerEarly) {
+    // `setupFromConfig` provides defaults for `colors` and `logLevel`.
+    // `setup` provides defaults for `appenders`
+    // The first argument MUST BE an object
+    logger.setupFromConfig({})
+  }
   function fail () {
```

---

</details>

devoto13: Both 2 and 4 look good to me. Let's go with the 4 as it is nicer from a backward compatibility persepective.

npetruzzelli:
If everything looks good, then next tasks are:

-   if needed, update existing tests
-   add meaningful tests for the new code

lib/cli.js
devoto13: > And guard calls in server/runner/stopper by the instanceof Config check for the backward compatibility. They will be cleaned up in the next major together with the rest of the old parsing logic.

Can you also do this please? Just move them into if/else we alreayd have for `parseConfig()`.

lib/cli.js
```diff
@@ -278,26 +279,47 @@ exports.process = () => {
...
+          require('./runner')
+            .run(config)
+            .on('progress', printRunnerProgress)
+          break
+        case 'stop':
+          require('./stopper').stop(config)
+          break
+      }
+    }, function onKarmaConfigRejected (reason) {
+      console.error(reason) // TODO: configure a CLI Logger?
```
npetruzzelli, line 308: Should this be removed? The way `fail` turned out, a logger is already called with a message that is almost the same a `reason`.

https://github.com/karma-runner/karma/blob/c4fd16e88ccdc79410ed579a32add267b6789e30/lib/config.js#L364-L371

devoto13: Yes, please!

devoto13:
@npetruzzelli Yes, this looks good now. We can proceed with adding new tests and we should also update relevant docs. In particular at https://github.com/karma-runner/karma/blob/master/docs/dev/04-public-api.md.

lib/cli.js
npetruzzelli: Just to make sure I understand correctly, you are looking for something like this in each file:


```diff
 class Server extends KarmaEventEmitter {
   constructor (cliOptionsOrConfig, done) {
     super()
     cliOptionsOrConfig = cliOptionsOrConfig || {}
-    logger.setupFromConfig({
-      colors: cliOptionsOrConfig.colors,
-      logLevel: cliOptionsOrConfig.logLevel
-    })
     this.log = logger.create('karma-server')
     done = helper.isFunction(done) ? done : process.exit
     this.loadErrors = []
 
     let config
     if (cliOptionsOrConfig instanceof cfg.Config) {
       config = cliOptionsOrConfig
     } else {
+      logger.setupFromConfig({
+        colors: cliOptionsOrConfig.colors,
+        logLevel: cliOptionsOrConfig.logLevel
+      })
       const deprecatedCliOptionsMessage =
         'Passing raw CLI options to `new Server(config, done)` is ' +
         'deprecated. Use ' +
         '`parseConfig(configFilePath, cliOptions, {promiseConfig: true, throwErrors: true})` ' +
         'to prepare a processed `Config` instance and pass that as the ' +
         '`config` argument instead.'
       this.log.warn(deprecatedCliOptionsMessage)
       try {
         config = cfg.parseConfig(
           cliOptionsOrConfig.configFile,
           cliOptionsOrConfig,
           {
             promiseConfig: false,
             throwErrors: true
           }
         )
       } catch (parseConfigError) {
         // TODO: change how `done` falls back to exit in next major version
         //  SEE: https://github.com/karma-runner/karma/pull/3635#discussion_r565399378
         done(1)
         return
       }
     }
```

devoto13: Exactly!

npetruzzelli:
> @npetruzzelli Yes, this looks good now. We can proceed with adding new tests and we should also update relevant docs. In particular at https://github.com/karma-runner/karma/blob/master/docs/dev/04-public-api.md.

I did a first pass at the parseConfig section and noticed something not relevant to this PR that should still probably be updated.

https://github.com/karma-runner/karma/blob/d3ff91a5a2f7ff84141096bb47aa31ff73f81cca/docs/dev/04-public-api.md#L163-L164

1.  The URL is no longer valid. The new location of the package appears to be:  
    <https://github.com/stryker-mutator/stryker/tree/master/packages/karma-runner>
2.  This project appears to no longer use `parseConfig`. a search for "parseConfig" in the repo yielded 0 results. It appears they pass an object with just the `configFile` property now.  
    https://github.com/stryker-mutator/stryker/blob/dce98cb944a7d8f8c8f67571f2e3209278b6c48a/packages/karma-runner/src/starters/karma-starter.ts#L7

Would you like me to remove or otherwise modify the sentence using **stryker-karma-runner** (now **@stryker-mutator/karma-runner**) as an example?

devoto13:
> Would you like me to remove or otherwise modify the sentence using stryker-karma-runner (now @stryker-mutator/karma-runner) as an example?

I fine removing it in this PR.

New documentation looks great by the way! I think we should also update examples for server/runner/stopper to show usage with both flags on, mention that they will become the defaults in the next major and refer users to the `parseConfig` documentation for more details on the flags.

docs/dev/04-public-api.md
```diff
@@ -156,20 +156,159 @@ stopper.stop({port: 9876}, function(exitCode) {
...
+  path.resolve('./karma.conf.js'),
+  { port: 1337 }
+);
+```
+
+The new behavior in the future will involve throwing exceptions instead of
+exiting the process and aynchronous config files will be supported through the
+use of promises.
+
+##### New Behavior
```
devoto13, line 189: nit: Merge "New Behavior" and "New Behavior with Promise Support" into a single section with a single example and both flags on to not overload reader with a lot of options.

npetruzzelli:
I could probably dedicate an entire PR to more updates for the docs, but I think I've reached a good stopping point for this PR.

If all outstanding items have been resolved, I will begin looking at tests.

lib/runner.js
```diff
@@ -32,14 +32,42 @@ function parseExitCode (buffer, defaultExitCode, failOnEmptyTestSuite) {
...
+function run (cliOptionsOrConfig, done) {
+  cliOptionsOrConfig = cliOptionsOrConfig || {}
   done = helper.isFunction(done) ? done : process.exit
-  config = cfg.parseConfig(config.configFile, config)
 
+  let config
+  if (cliOptionsOrConfig instanceof cfg.Config) {
+    config = cliOptionsOrConfig
+  } else {
+    // TODO: Should `const log = logger.create('runner')` be moved into the
```
devoto13, line 43: nit: I'm fine with either, including letting it be inconsistent. So choose whichever option and remove this TODO.

Otherwise, LGTM to start working on tests.

npetruzzelli: I'll remove the comment and begin work on the tests.

<details>

The code in question is the use of this function

https://github.com/karma-runner/karma/blob/d3ff91a5a2f7ff84141096bb47aa31ff73f81cca/lib/logger.js#L88-L106

My preference is to let the code speak for/document itself. In this case, there is nothing instance specific about the result of `logger.create`. My call would normally be to use it once at the module level, so `server.js` and `stopper.js` would be updated.

I don't expect the change would break anything, but in the spirit of expecting the unexpected, I'm going to leave it alone for now, delete the comment, and let it remain inconsistent. This PR should remained focused on promise support. Changing the logger statement, however minor it appears to be, falls under the category of general maintenance or logger improvements. As long as someone is aware of it, then the comment being removed accomplished its goal.

Since it pulls from a cache AND none of karma's internal code appears to use the second argument, there is effectively no difference between using it at the constructor/function level vs once at the module level.

</details>

devoto13: I agree that it should not affect anything. Feel free to send a separate PR to make it consistent. Maybe there are other places where we have a similar inconsistency. Having the call at the module level makes sense to me.

npetruzzelli:
I may have over extended in some places and not gone far enough in others, but the first round of tests is ready for review.

I think I was able to cover most of the branches.

lib/cli.js
```diff
@@ -278,27 +279,65 @@ exports.process = () => {
...
-      break
-    case 'completion':
-      require('./completion').completion(config)
-      break
+  const cliOptions = exports.process()
+  const cmd = cliOptions.cmd // prevent config from changing the command
+  const cmdNeedsConfig = cmd === 'start' || cmd === 'run' || cmd === 'stop'
+  let runPromise
+  if (cmdNeedsConfig) {
+    runPromise = cfg.parseConfig(
```
devoto13, line 287: I don't like all the noise we have added to this function only for testing. Let's instead make the function `async`, use `await cfg.parseConfig(` on this line and cleanup all other extra code. You don't seem to use any of the return values anyways.

devoto13 (commented):
Overall looks good! Left couple of comments to cleanup.

test/unit/config.spec.js
```diff
@@ -5,6 +5,24 @@ const loadFile = require('mocks').loadFile
...
+ * If an object is "thenable", then it is considered to be a promise
+ * implementation. It does not have to be an instance of ECMAScript's own
+ * `Promise` class to be considered a promise.
+ *
+ * @param {*} value
+ * @returns {boolean}
+ *
+ * @see {@link https://promisesaplus.com/}
+ */
+function isPromiseLike (value) {
```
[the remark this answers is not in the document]
npetruzzelli: Good point. parseConfig needs this logic because we don't know what kind of promise like object will be returned from the config file. Internally, Karma uses normal promises, so instanceOf will absolutely work for testing. I'll update this with the next push.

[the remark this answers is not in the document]
npetruzzelli: Keep the second [message] argument? Since a helper function isn't being used, the matcher's default message may be enough.

```js

        expect(noConfigFilePromise).to.be.an.instanceof(
          Promise,
          'Expected parseConfig to return a promise when no config file path is provided.'
        )
        expect(syncConfigPromise).to.be.an.instanceof(
          Promise,
          'Expected parseConfig to return a promise when the config file DOES NOT return a promise.'
        )
        expect(asyncConfigPromise).to.be.an.instanceof(
          Promise,
          'Expected parseConfig to return a promise when the config file returns a promise.'
        )
```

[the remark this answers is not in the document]
npetruzzelli: I made this initial change, but it appears to be failing. Even though a promise object is returned, both the matcher and the manual comparison of `noConfigFilePromise instanceof Promise` fails. I think I ran into something similar to this when working with jest. I could be misremembering, but it had to do with with a virtual machine or some other form of isolating the module under test. In this case Promise used by the module is different constructor than the Promise used by the test spec.

This is probably a side effect of `const loadFile = require('mocks').loadFile`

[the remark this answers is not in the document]
npetruzzelli: We could work around it with: 

```js
expect(Object.getPrototypeOf(noConfigFilePromise).constructor.name).to.eql('Prototype')
```

But I think this is a poor practice, as it is vulnerable to any class, function, method, or object property with the name "Promise". It isn't necessarily ECMAScript's Promise implementation.

test/unit/config.spec.js
```diff
@@ -30,6 +48,21 @@ describe('config', () => {
   const wrapCfg = function (cfg) {
     return (config) => config.set(cfg)
   }
+  function wrapCfgResolvedPromise (cfg) {
```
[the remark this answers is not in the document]
npetruzzelli: No we don't.

<details>

I thought it was going to be used more than once like the other wrapper. It could be inlined like some of the other mocks. That said, it is also a reasonable enough wrapper with the potential to see more use if more async mocks are needed in the future.

</details>

At the very least I will change it to match your suggestion. I'm fine with moving it directly to the mock's property as well.

lib/cli.js
npetruzzelli: I wasn't pleased with it as well. It started to feel like bloat. Using async requires minimal changes and cleans up a number of statements, especially for `init` and `completion`. I'll make the update with the next push.

npetruzzelli:
@devoto13 - While working on updates for this round of review, I was running into an issue where random tests in `config.spec.js` would log `WARN [config]: urlRoot normalized to "/"`.

<details>

-   Sometimes it happens, sometimes it doesn't (making it difficult to replicate)
-   Sometimes it was the same test, sometimes it wasn't
-   It always only happened once per run
-   It was never any of the `parseConfig` tests
-   Using `.only` to reduce the number of tests made it much less likely to occur, though using `.only` on the top level `describe` didn't seem to affect it
-   The only console message that interrupts mocha's status reporter is: `(node:21192) Warning: Setting the NODE_TLS_REJECT_UNAUTHORIZED environment variable to '0' makes TLS connections and HTTPS requests insecure by disabling certificate verification.` which has been the norm for me since I started contributing.

</details>

I haven't been able to replicate the problem as of the latest commit. It may not be an issue anymore, but I wanted to make you aware of it.

test/unit/config.spec.js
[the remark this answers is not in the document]
devoto13: Looks good!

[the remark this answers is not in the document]
devoto13: > Keep the second [message] argument? Since a helper function isn't being used, the matcher's default message may be enough.

Let's keep them.

devoto13 (approved):
I've updated the title and description to provide some useful information to future readers. It looks good to go now.

@npetruzzelli Thanks for doing all the hard work! 💪

johnjbarton (approved):
awesome, thanks

npetruzzelli:
You're most welcome! Glad to help!

PR #3662, opened 2021-03-07
fix: ignore dollar replacement sequences in file paths

charlessuh:
When using `string.replace` as a templating engine, care must be taken to escape dollar signs as they can be interpreted as replacement patterns: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_a_parameter

Note yarn uses a special `$$virtual` folder in [some circumstances](https://yarnpkg.com/configuration/yarnrc#virtualFolder), so this can break setups involving yarn.

### DIFF
--- a/lib/middleware/karma.js
+++ b/lib/middleware/karma.js
@@ -31,6 +31,11 @@ const FILE_TYPES = [
 ]
 
 function filePathToUrlPath (filePath, basePath, urlRoot, proxyPath) {
+  // Escape all dollar signs as it can be interpreted as replacement
+  // sequences ($&, $`, $', $n, $nn) when doing the final replacement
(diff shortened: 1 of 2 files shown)


charlessuh:
@googlebot I signed it!

devoto13:
Thanks for a PR @charlessuh! However, I think the approach from #3531 is better as it targets the root cause of the problem and does not create an implicit dependency between the generation of URLs and how we format them into the page body.

If you're interested in this fix, maybe you can build on top of that PR and fix the comment there, so we can get it in?

charlessuh:
Ah, awesome.

PR #3666, opened 2021-03-18
fix(client): do not send when frameworks report errors

jginsburgn:
Calling `complete` on Karma's client when `error` is called, leads to potentially having multiple `complete`. Besides, it is the job of the testing framework adapter to call `complete`, even if there was an error, when the tests are done.


devoto13:
The comment message above this method says:

```
// error during js file loading (most likely syntax error)
// we are not going to execute at all. `window.onerror` callback.
```

We must send a `complete` event to the server in this case because if there was a syntax error in the test code, there is no guarantee, that the test framework didn't crash. Then we'll never receive a `complete` event.

Please provide a minimal reproduction where the current logic causes the double `complete` event and let's look into specifics. There is probably a fix necessary, but the current code does not look like a correct fix to me.

PR #3668, opened 2021-03-28
fix: fix running tests in IE9

### LINKED ISSUE #3665: Karma 6.x fails to load in IE9
Updating from karma 5.2.3 to 6.2.0 I find that the tests no longer load in IE9, for example via SauceLabs on the <https://github.com/js-reporters/js-reporters> repository:

```
IE 10.0 (Windows 7): Executed 23 of 23 SUCCESS (0.232 secs / 0.029 secs)
IE 10.0 (Windows 7): Executed 23 of 23 SUCCESS (0.232 secs / 0.029 secs)
Firefox 45.0 (Windows 7): Executed 23 of 23 SUCCESS (0.199 secs / 0.061 secs)
Edge 15.15063 (Windows 10): Executed 23 of 23 SUCCESS (0.312 secs / 0.075 secs)
IE 11.0 (Windows 7): Executed 23 of 23 SUCCESS (0.21 secs / 0.034 secs)
Chrome 58.0.3029.81 (Windows 7): Executed 23 of 23 SUCCESS (0.216 secs / 0.045 secs)
Edge 88.0.705.50 (Windows 10): Executed 23 of 23 SUCCESS (0.186 secs / 0.024 secs)
14 03 2021 02:04:20.287:WARN [launcher]: internet explorer 9  on SauceLabs have not captured in 60000 ms, killing.
14 03 2021 02:05:23.193:WARN [launcher]: internet explorer 9  on SauceLabs have not captured in 60000 ms, killing.
14 03 2021 02:05:26.784:ERROR [launcher]: internet explorer 9  on SauceLabs failed 2 times (timeout). Giving up.
```

In the video recording, I see that the "Karma is Idle" page appears but nothing after that. No further error what I could find, so probably something very early on. Perhaps something with socket.io, or unsupported ES6+ syntax or method in client changes?

devoto13:
So the symptom is that after Karma page is loaded in IE9 the tests don't run and Karma is stuck at the "starting" step. Reloading the page once or twice makes IE9 connect to the server and tests pass successfully.

After an "exciting" debugging session I figured out that the problem is that on the first load (and sometimes after the page reload) [`io` variable](https://github.com/karma-runner/karma/blob/026fff870913fb6cd2858dd962935dc74c92b725/client/main.js#L14), which should be set by the `socket.io.js` script is undefined. It looks like IE9 silently stops executing `socket.io.js` script somewhere midway and switches to the next script causing the `io` variable to be undefined. I assume it hits some limit or threshold or whatnot. Couldn't find much information on the Internet, besides people experiencing similar issues when loading e.g. unminified jQuery and their own code firing before the `$` variable is available. No good solutions found.

Checking the `socket.io.js` contents, I have noticed that v2 served the minified client script on that path, while v3 now serves the unminified code. Presumably the huge script size is what causing IE9 to choke. Switching back to the minified socket.io client bundle allows tests to pass in IE9.

With the fix, tests passed 5 times in a row, while without it they fail every time, so I assume it does resolve the problem.

Fixes #3665

### DIFF
--- a/static/client.html
+++ b/static/client.html
@@ -109,7 +109,7 @@ It contains socket.io and all the communication logic.
 
   <iframe id="context" src="about:blank" width="100%" height="100%"></iframe>
 
-  <script src="socket.io/socket.io.js"></script>
+  <script src="socket.io/socket.io.min.js"></script>
   <script src="karma.js"></script>
 </body>
 </html>
--- a/static/client_with_context.html
+++ b/static/client_with_context.html
@@ -104,7 +104,7 @@ Is used for single-shot tests.
   <div id="banner" class="offline">
     <h1 id="title">Karma - starting</h1>
   </div>
-  <script src="socket.io/socket.io.js"></script>
+  <script src="socket.io/socket.io.min.js"></script>
   <script src="karma.js"></script>
   <script src="context.js"></script>
   <!-- The scripts need to be at the end of body, so that some test running frameworks
--- a/test/client/karma.conf.js
+++ b/test/client/karma.conf.js
@@ -27,6 +27,13 @@ const launchers = {
     browser_version: '11.0',
     os: 'Windows',
     os_version: '10'
+  },
+  bs_ie9: {
+    base: 'BrowserStack',
+    browser: 'IE',
+    browser_version: '9.0',
+    os: 'Windows',
+    os_version: '7'
   }
 }
 


johnjbarton (approved):
Oh! Based on the fails we saw I thought this was some use of an ES feature by socket-io that IE9 did not support.  Good work!

jginsburgn:
How did you debug IE?

devoto13:
@jginsburgn I logged in to the https://browserstack.com used by Karma, got my user name and access key under ACCESS KEY. Then I added something like below at the top of the `test/client/karma.conf.js` file and commented out all launchers except the one I want to debug:

```
process.env.TRAVIS = '1'
process.env.BROWSER_STACK_USERNAME = 'foo'
process.env.BROWSER_STACK_ACCESS_KEY = 'bar'
```

Then I ran `npm run test:client` as usual, went back to https://browserstack.com, and found my running job. There is a Start interactive session or something button there, which allows you to tap into a VM, open IE's developer tools and look into what is going on.

jginsburgn:
Thanks for that! I need to get up to speed on this in order to be better prepared for debugging IE issues that might happen.

PR #3671, opened 2021-04-06
Assorted link fixes

XhmikosR:
There are plenty more redirected links, but I thought I'd make a first PR with these fixes. :)

I could probably squash the first and the third patch into one if you prefer.

PS1. I'd appreciate if someone could update the build assets for me before merging this. I thought this was automated and there are too many npm vulnerabilities AFAICT to even try to do `npm i`.
PS2. I'm not familiar with the commitlint conventions, so bear with me here
PS3. I personally prefer using `Number.POSITIVE_INFINITY` for concurrency, but I'm not 100% sure on which Node.js versions is supported. It doesn't change the functionality but it's more explicit.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,4 +7,4 @@ assignees: ''
 
 ---
 
-Please read http://karma-runner.github.io/4.0/intro/troubleshooting.html first
+Please read https://karma-runner.github.io/4.0/intro/troubleshooting.html first
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,7 +120,7 @@ server.start();
 
 `dart` file type has been removed without a replacement.
 
-`customFileHandlers` DI token has been removed. Use [`middleware`](http://karma-runner.github.io/5.2/config/configuration-file.html#middleware) to achieve similar functionality.
+`customFileHandlers` DI token has been removed. Use [`middleware`](https://karma-runner.github.io/5.2/config/configuration-file.html#middleware) to achieve similar functionality.
 
 `customScriptTypes` DI token has been removed. It had no effect, so no replacement is provided.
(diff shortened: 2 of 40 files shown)


devoto13:
1. It is automated, the generated website will be updated once the next release is published. So you don't need to worry about it.
2. You can do something like the below to make CI happy:

   docs: fix redirected npmjs links
   docs: change npm to lowercase
   docs: use https when possible

XhmikosR:
Thanks for the hints! That being said, the changes don't affect only the docs. How can I combine more than one component? Maybe `docs,core`?

devoto13:
AFAICT only comments are updated, so it should be fine with just docs scope.

devoto13:
AFAICT only comments are updated, so it should be fine with just docs. Or you can use “docs(core): message” if changes are specific to core.

XhmikosR:
There's one link in code too but I don't think it matters a lot.

devoto13:
> There's one link in code too but I don't think it matters a lot.

Yes.

PS I've amended the last commit to make CI happy.

XhmikosR:
I'd like to keep those links, though. Can't you just update the dist files?

devoto13:
> I'd like to keep those links, though. Can't you just update the dist files?

Unfortunately, we can't do it. Otherwise, we'll have to give up (unnecessarily complicate) a check that bundled client code is up-to-date, which is IMO more important than not having HTTP links.

The right fix would be to change the link in https://github.com/webmodules/dom-serialize, get a release, and update, but the package looks abandoned. Not sure if it worth the trouble.

XhmikosR:
Ah, OK, that makes sense. I wasn't aware that was 3rd-party code. All good then :)

XhmikosR:
BTW the bots are too chatty 😛

devoto13:
> BTW the bots are too chatty 😛

You can block them to avoid any notifications. At least that's what I did 😄

johnjbarton (approved):
Thanks!

PR #3674, opened 2021-04-11
chore: replace `init` scripts with a dependency on itself

devoto13:
This has the same result (symlink to the `karma` in the node_modules), but has two benefits:

- karma developers don't need to run extra command
- no peer dependency warnings when `npm install` is run


johnjbarton (approved):
awesome

Issue #3677, opened 2021-04-26
Support for ESM config files (with "type": "module")


tschaub:
If a project's `package.json` has `"type": "module"`, Karma fails to load the config file.  I'm wondering if there would be interest in supporting ESM config files in addition to CJS (and the other currently supported flavors).

While this might not be complete or the desired solution, it looks like something like this could add ESM config support: https://github.com/tschaub/karma/commit/99f1a92ec85a1dd2efbcd0faf1ad653e4b29cf8e

devoto13:
Yes, we would be happy to accept a PR adding support for the ESM config file.

https://github.com/tschaub/karma/commit/99f1a92ec85a1dd2efbcd0faf1ad653e4b29cf8e is a good start, but it will need some more work. In particular, we'll need to mention this feature in the [documentation](https://github.com/karma-runner/karma/blob/master/docs/config/01-configuration-file.md#overview) and add [e2e test](https://github.com/karma-runner/karma/tree/master/test/e2e) to prevent regressions (I think we can use `.mjs` instead of `type: module` for the purpose of testing for simplicity).

npetruzzelli:
@devoto13 

If I can free up some time, I will try to tackle this. I started a new job so it has kept me very busy! This seems like a natural extension of my previous 2 contributions.

Notes:

-   The Dynamic `import()` API was introduced in Node 12.x
    -   [v11 docs](https://nodejs.org/docs/latest-v11.x/api/esm.html)
    -   [v12 docs](https://nodejs.org/docs/latest-v12.x/api/esm.html#esm_import_expressions)

Both static `import` statements and dynamic `import()` calls both support CommonJS AND ECMA Modules.

The trick is, we would have to use `import()` to replace `require` calls when retrieving the config file path that is passed to us, which makes the use of `Promise` a requirement.

On top of that, I don't know if `import()` would work the same way as `require` with the `coffeescript`, `Livescript`, or `ts-node` packages. See: https://github.com/karma-runner/karma/blob/v6.3.2/lib/config.js#L428-L446

The last time I checked, it didn't work because Node's API for intercepting modules isn't yet stable.

I wouldn't hold your breath waiting for me until you see at least a draft PR. It may be a while before I can start.

@tschaub 

If you have a more immediate need, you could consume the JavaScript API immediately.

1.  let `myKarmaConfig` be the result of passing `null` as the first argument to `karma.config.parseConfig()`
    - (set the 2nd and 3rd arguments as needed)
[...]

PR #3679, opened 2021-05-30
ESM Config File Support

npetruzzelli:
#3677

- Description to be updated
- Tests to be updated

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -354,6 +354,12 @@ const CONFIG_SYNTAX_HELP = '  module.exports = function(config) {\n' +
   '    });\n' +
   '  };\n'
 
+const ESM_CONFIG_SYNTAX_HELP = '  export default function(config) {\n' +
+'    config.set({\n' +
+'      // your config\n' +
+'    });\n' +
+'  };\n'
+
 /**
  * Retrieve a parsed and finalized Karma `Config` instance. This `karmaConfig`
  * object may be used to configure public API methods such a `Server`,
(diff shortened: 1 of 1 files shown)


npetruzzelli:
@devoto13 - Does Karma currently need to support any versions of node that doesn't support `import()`? (versions 11 and below).

devoto13:
Yes, Karma still supports Node 10, but it will be dropped in the next major release as Node 10 has reached the end of life in April. I would suggest to let this PR hang for a bit longer, until we start a branch for the next major release, so we need less cleanup to do.

npetruzzelli:
Sounds like a good plan to me.

loynoir:
Might I ask about when, will this go into master? 😂

---
In case, someone search issue and come here like me, here's my quick workaround using require hook.

https://github.com/karma-runner/karma/issues/3677#issuecomment-984527542

npetruzzelli:
At the very least I would have to merge in the latest `master` branch to bring this current. I think this repo also had a period of downtime when continuous integration was being updated.

Realistically, this isn't going to be the PR that gets merged, for a few reasons:

1.  Karma 6.x _MUST_ support Node.js version 10
    -    to change this is a breaking change, which requires a new major version (Karma 7.x)
    -    a Node feature required to support this did not appear until Node 11.x
2.  Karma 7.x is expected to rewrite significant chunks of config code (deprecating/dropping support for some things to make the code more maintainable)
    -   as a result, how `import()` is supported may change a bit as well
    
    
@devoto13 - is there a milestone / tag / some kind of tracker that lists what needs to be done before a pre-release branch for Karma 7.x is made available for contributions? I might have some time this month/next month to make a contribution

devoto13:
@npetruzzelli CI is fixed now. I think it's mainly some maintainers' time to implement/review/test/land breaking changes, so we don't end up being stuck in pre-release mode for a long time and probably @jginsburgn's approval to start working on the next major 😄

There is https://github.com/karma-runner/karma/issues/3503 tracking breaking changes I would like to make (not necessarily all of them) + some of the open PRs.

It is pretty hot time at work now, so I'm not sure when I'll have to time to look into it.

Note that it's possible to install karma from GitHub as well. So if you rebase this PR, people can install it using similar approach as https://github.com/karma-runner/karma/pull/3726#issuecomment-968708216.

npetruzzelli:
Got it, if I have a chance, I will at least try to update this branch. Thank you.

In the mean time, @loynoir may be interested in: https://github.com/karma-runner/karma/issues/3677#issuecomment-996972105

PR #3686, opened 2021-06-28
chore: update tooling packages to the latest versions

devoto13:
Configurations are updated to accommodate for breaking changes.


jginsburgn:
@devoto13 is this still a work in progress?

devoto13:
@jginsburgn It is ready, but CI is not running because travis-ci.org was [shut down](https://blog.travis-ci.com/2021-05-07-orgshutdown) recently. I wonder if we should migrate to travis-ci.com? Or maybe migrate from both Travis and AppVeyour to GitHub Actions to have a single CI system for all platforms. What do you think?

XhmikosR:
My suggestion would be to use GitHub Actions for both Windows and Linux.

jginsburgn:
I think that the effort to migrate to circle-ci.com would be much less than to do it to another platform. I will take a look at this soon :)

PR #3697, opened 2021-09-07
Return exitCode from onExit exitCode setters

DavidANeil:
This allows a listener to get the exitCode as well.


jginsburgn:
Why did you closed this one?

DavidANeil:
I created it because we were patching the framework to do exactly this, but that ended up leading us to do a not-very-good thing, so I figured closing this would help other people to not make the same mistake.

jginsburgn:
Thanks for your clarification :)

PR #3701, opened 2021-10-01
chore: migrate to GitHub Actions for CI

devoto13:
- Use headless browsers instead of xfvb where possible.
- Replace commitlint-travis with usage of the base commitlint as there is no official CLI for GitHub Actions.
- Remove CI badges as GitHub displays the CI status natively for each commit.

Remaining work (for which I don't have permissions):

- [ ] Land https://github.com/karma-runner/integration-tests/pull/15 and update this PR to use upstream for integration tests.
- [ ] Copy all relevant environment variables from Travis settings to repository secrets: https://github.com/karma-runner/karma/settings/secrets/actions. Note that secrets are available to branches in the repo, but [not available to forks](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow), so GitHub secrets provide the same functionality and security as Travis used to.
  - [ ] BROWSER_STACK_USERNAME
  - [ ] BROWSER_STACK_ACCESS_KEY
  - [ ] SAUCE_USERNAME
  - [ ] SAUCE_ACCESS_KEY
  - [ ] NPM_TOKEN
  - [ ] GH_TOKEN (not 100% if this one is needed with GitHub Actions, I guess we can try without it and see)


jginsburgn:
Why do we want to use headless browsers instead of xvfb?

jginsburgn:
>  Remove/disable Travis/AppVeyor?

I have deleted the Travis webhooks. Can you clarify what is missing for Travis?

devoto13:
> I have deleted the Travis webhooks. Can you clarify what is missing for Travis?

I'll open a new PR to confirm that Travis and AppVeyor are not triggered to see if that's enough or we need to do something else.

> Shouldn't we use BROWSER_STACK_ACCESS_KEY instead?

I don't understand. What exactly do you mean by this?

> Where and how are these used?

SuaceLabs tokens are used in the [integration tests](https://github.com/karma-runner/integration-tests), which are run by `npm run test:integration`.

> Just to confirm, this is only used in the "Release" workflow by the npm run semantic-release step, right?

That's correct.

> I think this is not needed :)

That's actually used by [update-docs.js](https://github.com/karma-runner/karma/blob/master/tools/update-docs.js) to update [documenation](https://github.com/karma-runner/karma-runner.github.com) when release is cut. I would suggest putting the same token as we have in Travis to secrets as it should have push access to the documentation website repo. Or alternatively, we can generate a new token specifically for that repo and store it under a different name (`DOCS_GITHUB_TOKEN` or something). Let me know if you would prefer the second solution?

PR #3708, opened 2021-10-16
Fix GH Actions Parameters and Improve Integration with BrowserStack

jginsburgn:
For details, please see the commit messages.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,6 +16,7 @@ jobs:
       SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
       NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      DOCS_GITHUB_TOKEN: ${{ secrets.DOCS_GITHUB_TOKEN }}
       KARMA_TEST_NO_FALLBACK: 1
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


devoto13:
Actually, one thing. Can you change the type of your commits to `chore` as `fix` will make semantic-release cut a new release, which we probably don't want to do?

I've submitted https://github.com/jginsburgn/karma/pull/1. I'm not particularly sure how did John generate the token for Travis, but as far as I understand it should be [PAT](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) with the write access to https://github.com/karma-runner/karma-runner.github.com. I think that the default GitHub token can only write to this repo and will not work for the script.

jginsburgn:
> @googlebot I consent.

Thanks. One last thing before I merge, I would prefer that the commit messages, which contain valuable information, do not get lost. So instead of squashing, I would fast forward for merging. Sounds good?

jginsburgn:
> > And, how are we going to know that the docs fix worked?
> 
> I guess we'll know for sure once the next release is made, but for now, you can try to do a similar thing to what the script does locally to confirm that the token actually allows to push to the repo
> 
> ```shell
> $ export DOCS_GITHUB_TOKEN=<token>
> $ git clone https://$DOCS_GITHUB_TOKEN@github.com/karma-runner/karma-runner.github.com.git
> $ cd karma-runner.github.com
> $ git checkout -b tmp
> $ touch test.txt
> $ git add text.txt
> $ git commit -m test
> $ git push origin tmp
> $ git push -d origin tmp
> ```
> 
> or something :)

It works!!!!!

PR #3718, opened 2021-10-28
fix(middleware): replace %X_UA_COMPATIBLE% marker anywhere in the file

### LINKED ISSUE #3711: `%X_UA_COMPATIBLE%` displayed in the runner web page in debug mode
The runner web page in debug mode displays an `%X_UA_COMPATIBLE%` text.

image

devoto13:
Previously %X_UA_COMPATIBLE% marker was only replaced if it was located at the start of the line. The limitation looks pretty arbitrary and caused the marker not to be replaced in the custom debug.html file used by Angular CLI as the marker was not located at the start of the line (probably because the file was re-formatted). This commit changes the behavior to replace the marker anywhere within the file, not just at the start of the line and thus fixes the problem for Angular CLI and potentially other people using custom files.

Fixes #3711

### DIFF
--- a/lib/middleware/karma.js
+++ b/lib/middleware/karma.js
@@ -45,7 +45,7 @@ function getQuery (urlStr) {
 function getXUACompatibleMetaElement (url) {
   const query = getQuery(url)
   if (query['x-ua-compatible']) {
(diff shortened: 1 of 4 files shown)


jginsburgn (approved):

sroucheray:
Thanks for the fix @devoto13 & @alan-agius4 !

PR #3719, opened 2021-10-30
build: use matching commitlint config

devoto13:
The commit conventions enforced by commitlint (conventional) were different than the conventions used by the semantic-release (angular). In practice, is was never a problem as they are pretty similar, but it's better to be consistent anyways. Update the commit conventions documentation according to the change and also extend it with the information about which kind of release is triggered by which commit type. Add explicit configuration for release rules for semantic-release to the configuration file - rules are the same as before, but now it is easier to see what they actually are.

### DIFF
--- a/docs/dev/06-git-commit-msg.md
+++ b/docs/dev/06-git-commit-msg.md
@@ -2,6 +2,8 @@
 showInMenu: false
 ---
 
+In the repository we use and enforce the commit message conventions. The conventions are verified using [commitlint] with [Angular config](https://www.npmjs.com/package/@commitlint/config-angular).
+
 ## The reasons for these conventions:
 - automatic generating of the changelog
 - simple navigation through git history (e.g. ignoring style changes)
@@ -9,9 +11,9 @@ showInMenu: false
 ## Format of the commit message:
 ```bash
 <type>(<scope>): <subject>
-
+<BLANK LINE>
 <body>
-
+<BLANK LINE>
 <footer>
 ```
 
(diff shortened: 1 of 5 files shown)


docs/dev/06-git-commit-msg.md
```diff
@@ -55,16 +56,9 @@ The `<scope>` can be empty (e.g. if the change is a global or difficult
...
 ## Message body
-* uses the imperative, present tense: “change” not “changed” nor “changes”
-* includes motivation for the change and contrasts with previous behavior
-
-For more info about message body, see:
-
-* https://365git.tumblr.com/post/3308646748/writing-git-commit-messages
-* https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
 
+Just as in the `<subject>`, use the imperative, present tense: "change" not "changed" nor "changes". The body should include the motivation for the change and contrast this with previous behavior. Message body should include motivation for the change and contrasts with previous behavior.
```
jginsburgn, line 61: Did we copy this and the other paragraphs from Angular's repo?

devoto13: I guess the original prose was copied/inspired by the AngularJS commit message conventions. This particular paragraph I've copied from https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit, yes.

PS Just noticed that I accidentally added the same sentence twice. I'll push a fix shortly.

jginsburgn: SG.

package.json
```diff
@@ -445,7 +445,7 @@
   },
   "devDependencies": {
     "@commitlint/cli": "^12.1.4",
-    "@commitlint/config-conventional": "^12.1.4",
+    "@commitlint/config-angular": "^12.1.4",
```
jginsburgn, line 448: Why are we using an old version?

devoto13: v13 [has dropped Node 10 support](https://github.com/conventional-changelog/commitlint/compare/v12.1.4...v13.0.0), while Karma still supports Node 10. I guess it's less important for commitlint given that it only runs on Node 14 on CI, but thought it would be better to be consistent and upgrade all dependencies once we drop Node 10 support.

jginsburgn: OMG you know everything.

jginsburgn (approved):

Issue #3715, opened 2021-10-26
Replace ua-parser-js or pin current version


davidje13:
It seems ua-parser-js has been the source of several security issues:

1. https://github.com/karma-runner/karma/issues/3562 (regex catastrophic backtracking)
2. https://github.com/karma-runner/karma/issues/3583 (regex catastrophic backtracking)
3. https://github.com/karma-runner/karma/issues/3680 (regex catastrophic backtracking)
4. https://github.com/karma-runner/karma/issues/3713 (malware compromise)

The catastrophic backtracking issues of course aren't much of a problem for Karma, but the recent malware compromise certainly is! Karma represents a large chunk of the library's usage (according to NPM, Karma represents ~25% of ua-parser-js's downloads), and since a version range import is being used, it means that Karma may have unwittingly been a big contributor to spreading the latest malware.

Yet the library is only used in a single location for a very minor purpose (constructing a "friendly" name for browsers to show in the logs; as far as I can tell, this is an undocumented internal function with no particular promises about its API): https://github.com/karma-runner/karma/blob/94cf15e8fa4420c8716998873b77f0c4f59b9e94/lib/helper.js#L9

Maybe it's time to swap that function out for a home-grown simplified version? I think there are a few options:

- handle just the UAs for browsers with supported launchers, and the rest can use the fallback full version string
- don't try to create friendly names for any browsers; just use the full version string for everything
- continue using ua-parser-js, but pin a specific version (`0.7.30`) instead of a range (`^0.7.30`) to at least avoid any future compromises being auto-fetched (and perhaps worth doing a quick review of the current code for other issues, if going with this option)
- any of the above, plus make the function configurable (so that users can provide their own full-user-agent -> nice-name function, which could of course use any version of ua-parser-js if the user wants it). If doing this, my preferred choice would be to have the default just pass through the raw UA, removing the need for the dependency and keeping the code simple.

Do the maintainers have any preferences here?

davidje13:
Created a PR

LucasLopesr:
any news?

PR #3720, opened 2021-10-30
Replace ua-parser-js with a reduced in-house version

davidje13:
As discussed in #3715

This replaces the external library `ua-parser-js` with an in-house version which has a significantly reduced scope.

The parser has been written from scratch with a focus on passing the existing tests, but should be easy to extend with more browsers and OSes if required (though I would recommend keeping it minimal and maybe even _removing_ some entries). The method used is a bit more robust than the wildcard searches used in the other project, and should be less prone to accidental catastrophic backtracking issues, but the main point of this is to reduce the breadth of dependencies to help avoid upstream attacks in the future.

Currently it is not possible for users to plug in their own function; the recommendation would be for people who need "friendly" names for a wide selection of browsers to use the `fullName` property instead and do their own parsing (e.g. via `ua-parser-js` if they wish).


davidje13:
@googlebot I signed it

jginsburgn:
Nice CL!

lib/helper.js
```diff
@@ -3,15 +3,132 @@
...
-    return fullName
+  const parts = extractUAParts(fullName)
+  const [browserName, browserVersion] = extractFromParts(parts, UA_BROWSERS)
+  const [osName, osVersion] = extractFromParts(parts, UA_SYSTEMS)
+  if (browserName || osName) {
+    return `${browserName || 'unknown'} ${browserVersion || '0.0.0'} (${osName || 'unknown'} ${osVersion || '0.0.0'})`
   }
-  return `${ua.browser.name} ${ua.browser.version || '0.0.0'} (${ua.os.name} ${ua.os.version || '0.0.0'})`
+  return fullName || 'unknown'
 }
```
jginsburgn, line 132: I wonder if we should move this logic to its own file. @devoto13 WDYT?

lib/helper.js
```diff
@@ -3,15 +3,132 @@
 const fs = require('graceful-fs')
 const path = require('path')
 const _ = require('lodash')
-const useragent = require('ua-parser-js')
 const mm = require('minimatch')
 
+const extractUAParts = (ua) => {
```
jginsburgn, line 8: Can we add some unit tests for this? @davidje13

lib/helper.js
```diff
@@ -3,15 +3,132 @@
...
+  ({ firstEntry }) => firstEntry.hasDetail(/^linux/i) &&
+    ['Linux', firstEntry.getDetail(/^linux (.+)/i)[1]],
+
+  ({ firstEntry }) => firstEntry.hasDetail(/mac os/i) &&
+    ['Mac OS', firstEntry.getDetail(/mac os(?: x)? ([\d._]+)/i, extractMacVersion)],
+
+  ({ firstEntry }) => firstEntry.hasDetail(/^windows/i) &&
+    ['Windows', firstEntry.getDetail(/windows nt ([\d.]+)/i, extractWindowsVersion)]
+]
+
```
jginsburgn, line 123: Where did you get all this info? Probably from the implementation of `ua-parser-js`? Can we add a reference?

jginsburgn: Also, can you please add comments to the functions?

jginsburgn (changes requested):

jginsburgn:
@davidje13 also please take a look at the checks as there are some failures :)

lib/helper.js
davidje13: source of this is actually the test user-agent strings which were already in this project. The code and regular expressions here have no relation to ua-parser-js's code.

davidje13: As for documentation: if the team is happy with this approach, I can extract the code into its own file, add finer-grained tests, and document. Wanted to get a quick first pass submitted without all that so that I didn't waste too much time if the team was opposed to the whole idea.

davidje13:
> @davidje13 also please take a look at the checks as there are some failures :)

seems Node 10 doesn't have `matchAll`. Easy to replace. I'll update it when I get time.

lib/helper.js
jginsburgn: I am happy with your approach. It is very elegant! :)
I'd like to hear from @devoto13 as well.

jginsburgn:
> > @davidje13 also please take a look at the checks as there are some failures :)
> 
> seems Node 10 doesn't have `matchAll`. Easy to replace. I'll update it when I get time.

Oh yeah! That is minor. :)

lib/helper.js
devoto13: Approach overall LGTM!
I'll try to look more into details in the coming days.

devoto13:
> seems Node 10 doesn't have matchAll. Easy to replace. I'll update it when I get time.

I would be more comfortable landing this change in a major release, so it's probably okay to keep using `matchAll` as we're going to drop Node 10 in the next major.

jginsburgn:
> > seems Node 10 doesn't have matchAll. Easy to replace. I'll update it when I get time.
> 
> I would be more comfortable landing this change in a major release, so it's probably okay to keep using `matchAll` as we're going to drop Node 10 in the next major.

Or, can we add a TODO where appropriate to replace the Node < 10 equivalent with `matchAll` when we release the next major?

Maybe something in the lines of:

```
// TODO(#3503): Replace the following logic with `matchAll`.
...
```

PR #3726, opened 2021-11-12
Significantly improved addFile/changeFile/removeFile performance.

jfstephe:


jfstephe:
@googlebot - signed it!

devoto13:
Thanks for the PR @jfstephe! The change looks very good 👍  
While the return value is not used by karma core, it may still be used by some third-party plugins, so this PR will probably have to wait until we're ready to open master brach for breaking changes before the next major release.

As for the CLA, most likely it is failing because the commit and the signed CLA have different emails.

jfstephe:
@devoto13 - CLA corrected, along with the commit message format (I think!) :-).

**Question:** Can you give me an idea of typical time-scales for this being released? If it's more than a couple of weeks I'll publish this to our private npm locally, if not I'll wait for a while. Thanks!

devoto13:
@jfstephe Thank you for fixing the issues!

I'm afraid it may take more than a couple of weeks before this gets released, so private package is probably the way to go. You should also be able to install your branch directly from GitHub using:

```
$ npm i jfstephe/karma#fix-file-change-performance
```

jfstephe:
Awesome, thanks!

lib/file-list.js
```diff
@@ -172,18 +173,18 @@ class FileList {
     const excluded = this._findExcluded(path)
     if (excluded) {
       log.debug(`Add file "${path}" ignored. Excluded by "${excluded}".`)
-      return this.files
+      return
```
jginsburgn, line 176: @devoto13 
I think not returning the file list is a breaking change. Should we not hold on to this change for release v7+ (https://github.com/karma-runner/karma/issues/3503)?

lib/file-list.js
```diff
@@ -114,6 +114,7 @@ class FileList {
     return lastCompletedRefresh
   }
 
+  // TODO - Change this so that it's not a getter. Getters shouldn't do lots of processing!
```
jginsburgn, line 117: Nit: `TODO:` instead of `TODO -` :)

lib/file-list.js
devoto13: @jginsburgn Yes, I think it should be better to land it in a major release.

jfstephe: sure

lib/file-list.js
jfstephe: Hi - I really don't want to spend the 15 minutes on context switching to remove a '-'. Can this be accepted as is (assuming no other issues)?

jfstephe:
Is this waiting on me for anything?

devoto13:
@jfstephe No, it's waiting until we have setup the branch for beta releases and then we'll merge. I'll fix the `-` before merging, so you don't need to bother with it.

jfstephe:
@devoto13 - Hi, it's been almost a year on this. Any eta on this being merged?

devoto13:
@jfstephe I don't have permissions to merge, so without somebody from Google, there is nothing I can do. And given the recent [karma deprecation announcement](https://github.com/karma-runner/karma#karma-is-deprecated-and-is-not-accepting-new-features-or-general-bug-fixes), I'm afraid that the most likely eta is never, sorry 😞

PR #3733, opened 2021-12-17
Document Asynchronous Configurations in the Config File's Docs

npetruzzelli:
See: https://github.com/karma-runner/karma/issues/3677#issuecomment-997002670


npetruzzelli:
I've been working so much in Azure lately, that I've forgotten how to mark my own PR as a Draft.

The default behavior of the server doesn't support promises, so some additional work will need to be done on the users side to get it working.

I need additional time to add that to this PR.

npetruzzelli:
I looked, but didn't see a way to compile/serve the docs locally (maybe I missed something obvious?) so I'm not sure if I linked to from the config file page to the public api page correctly.

Otherwise this should be ready for review!

devoto13:
@npetruzzelli The docs infra is located in the https://github.com/karma-runner/karma-runner.github.com repo. It's probably the easiest to copy your changes over to that repo and run `grunt` to spin up the local server. Note that you don't need to send a PR to that repo as we automatically sync changes from this repo as part of each release.

docs/config/01-configuration-file.md
```diff
@@ -67,6 +67,69 @@ require('ts-node').register({
...
+Just as with a synchronous configuration's return value not being used, an
+asynchronous configuration's resolved value will not be used. With both
+configuration types, `config.set()` MUST BE called to assign the configuration
+object.
+
+When using the command line interface, returning a promise is all that is
+needed.
+
+
+#### Asynchronous Configuration Using The Public API in Karma 6.x
```
devoto13, line 85: I don't think it is worth having this section here for two reasons:

- It will work out of the box in the upcoming v7.
- This file is primarily intended for the non-programmatic Karma users and programmatic Karma users will use the Public API page anyway where the concept is described in much detail already.

docs/config/01-configuration-file.md
```diff
@@ -67,6 +67,69 @@ require('ts-node').register({
...
+final call to `config.set()`.
+
+Just as with a synchronous configuration's return value not being used, an
+asynchronous configuration's resolved value will not be used. With both
+configuration types, `config.set()` MUST BE called to assign the configuration
+object.
+
+When using the command line interface, returning a promise is all that is
+needed.
+
```
devoto13, line 83: Maybe add an example to this section? Something like below to demonstrate the use case.

```javascript
// karma.conf.js
module.exports = async (config) => {
  const freePort = await findFreePort();
  config.set({
    port: freePort,
    basePath: '../..',
    frameworks: ['jasmine'],
    //...
  });
}
```

devoto13 (commented):
Thanks for adding the docs! Left two comments, but I don't have a strong opinion about them.

docs/config/01-configuration-file.md
npetruzzelli: I'd still like to point newer users in the right direction (just in case they don't think to check the API docs.)

How does this compromise look?

```diff
 ### Asynchronous Configuration

 The function exported by the configuration file may return a promise whose
 resolution will tell Karma when all configuration is complete, including the
 final call to `config.set()`.

 Just as with a synchronous configuration's return value not being used, an
 asynchronous configuration's resolved value will not be used. With both
 configuration types, `config.set()` MUST BE called to assign the configuration
 object.

 When using the command line interface, returning a promise is all that is
 needed.
-
-
-#### Asynchronous Configuration Using The Public API in Karma 6.x
-
-The ability to support promises was added as a feature in Karma `6.3.0`, but
-supporting it by default in the public API would be a breaking change. While
-this may change may in a future major version, users of Karma 6's public API
-have a few extra steps to go through in order to use promises. This way, the 
-behavior of the API does not change unexpectedly for existing users of versions
-`^6.3.0`.
-
-```js
-const path = require('path');
-
-const karma = require('karma');
-
-const cfg = karma.config;
-const Server = karma.Server;
-
-// Read karma.conf.js and start the Karma server
-cfg.parseConfig(
-  // **configFilePath** the file system path to the config file.
-  path.resolve('./karma.conf.js'),
-
-  // **cliOptions** takes priority over anything set by the config file.
-  {},
-
-  // **parseOptions** here we tell parseConfig that we want to use promises.
-  // Note that `throwErrors` MUST also be true to get promise rejections in
-  // place of immediately exiting the process.
-  { promiseConfig: true, throwErrors: true }
-).then(
-  (karmaConfig) => {
-    // use the config with the public API
-    const server = new Server(karmaConfig, function doneCallback(exitCode) {
-      console.log('Karma has exited with ' + exitCode);
-      process.exit(exitCode);
-    })
-  },
-  (rejectReason) => { 
-    // respond to the rejection reason error
-    console.log(rejectReason);
-    process.exit(1);
-  }
-);
-```
-
-For more information on using the public API, please see the
-[Public API Documentation][dev/public-api]]
+
+ If you are using the public API, then please see the 
+[Public API Documentation][dev/public-api]] for details, usage, and addtional
+information.
```

Hopefully it won't automatically date itself in v7+

devoto13: Sounds good, thanks!

npetruzzelli:
Updated. I intentionally left out arrow functions and async/await. My thinking being that I want the focus to be on the use of promises. An experienced developer will already know to use those things to simplify code, while a newer developer may be more anxious when seeing more unfamiliar things when they aren't required.

devoto13 (approved):
LGTM
Thanks!

devoto13:
@npetruzzelli Can you just squash into a single commit and remove the trailing "." from the commit message to make CI happy?

npetruzzelli:
I don't often manually squash or rebase, so I'm not sure I did it correctly.

devoto13:
@npetruzzelli No, does not look right, but no worries, I think @jginsburgn can squash when merging.

I usually can fix such issues myself, thanks to "Allow edits by maintainers" checkbox, but it does not seem to work in this case, because your fork is in the organisation. Or maybe I do something wrong. I think we can let it be for now.

jginsburgn (commented):
@npetruzzelli please excuse my long delay in getting to this.

Can you please rebase and squash? If you don't remember how to squash, maybe you can give me push permissions to your fork?

docs/config/01-configuration-file.md
```diff
@@ -67,6 +67,45 @@ require('ts-node').register({
...
+      config.set({
+        frameworks: ['jasmine'],
+        port: freePort,
+        // ... other Karma configuration
+      });
+    });
+}
+```
+
+If you are using the `parseConfig`method from the public API, then please see
```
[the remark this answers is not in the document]
npetruzzelli: @jginsburgn - I will try to fix the typo and rebase/squash by this weekend.

docs/config/01-configuration-file.md
```diff
@@ -67,6 +67,45 @@ require('ts-node').register({
...
+an async method to find an available port if the preferred port is already in
+use.
+
+```js
+// karma.conf.js
+const findAvailablePort = require('./findAvailablePort.js');
+
+module.exports = function configKarma(config) {
+  const myPreferredPort = 9876;
+  return findAvailablePort(myPreferredPort)
```
nicojs, line 94: This is a bad example IMHO, because it leaves the door wide open for race conditions to occur.

npetruzzelli: The most common description of a race that I've had experience with is that multiple operations are being executed in parallel with no way to control the order of execution, which can lead to an unpredictable/unstable state.

I don't see the potential for race conditions here. The hypothetical `findAvailablePort` returns a promise. That promise must resolve before the onFulfilled parameter executes (which then execute **`config.set`**). 

As long as it is a `Promise` instance, or an object that is a conforming implementation of Promises/A+, then the order of operations is controlled and there is no race.

Are you concerned about promise rejections not getting handled in the example?

npetruzzelli:
I managed to make a mistake and accidentally closed this PR as a side effect of hard resetting my branch.

The new PR can be found here: #3762

docs/config/01-configuration-file.md
nicojs: I mean that getting a free port and then using that port. The port can be claimed by another process in the mean time. 

I've experienced this twice on the past.
1. On an old fashion buildserver that ran parallel builds.
2. During mutation testing with StrykerJS, where we also need to spin up multiple test runner processes.

The right way to choose a free port is by leaving it `undefined` and letting karma itself choose it.

npetruzzelli: That is reasonable. In the time it takes between getting a port number and starting the server, something else could claim the port, especially if other async tasks happen before then.

The example can easily be changed to another process that is often asynchronous, "getThingFromFileSystem" or "getThingFromNetwork". I would rather the example draw attention to "hey, this is asynchronous" instead of "this is a way to solve this specific problem/task" even if the intent was for illustration. I'd even be ok with the super generic "doAsyncThing", as long as it gets the point across that it returns a promise.

nicojs: Exactly. I didn't see this PR at first and opened #3761 where I add documentation like that.

Issue #3738, opened 2022-01-09
pin colors lib to 1.4.0 as it been hacked and version 1.4.1 is broken


arelg:
Please read https://karma-runner.github.io/4.0/intro/troubleshooting.html first
https://github.com/Marak/colors.js/issues/285#issuecomment-1008168237

donmahallem:
No hack... just the author breaking his package on purpose. he did the same with faker.js

arelg:
didnt know ty @YoniSegal for patching

bouhady:
the problem also exists on 1.4.0 suddenly! 
is there any workaround for libraries such as gulp-nodemon which is using this version as sub-dependencies?

YoniSegal:
> the problem also exists on 1.4.0 suddenly! is there any workaround for libraries such as gulp-nodemon which is using this version as sub-dependencies?

Are you sure you're using 1.4.0?
I'm not having any issues.

YoniSegal:
> didnt know ty @YoniSegal for patching

No problem.
I closed my PR because I've never contributed to an open source project and am not sure i'm doing it right.

The PR is trying to merge into master, but I want it to fix an issue at the v4.4.1 tag.

XhmikosR:
@jginsburgn this should be addressed sooner than later since it seems npm hasn't removed the offending colors versions.

esskar:
Please provide a quick fix with an update to 1.4.0

RupprechJo:
Yes, this is really urgent, breaks all builds with karma on the CI Server

donmahallem:
Can't you just add colors as dev dependency with pinned version to 1.4.0 as a temporary fix? 

E.g.

`npm install --save-dev --save-exact colors@1.4.0`

sharif-elshobkshy:
Hello everyone.

Currently Magicscroll is using the newest/latest "Colors" version.

https://github.com/karma-runner/karma/blob/master/package.json#L427
`"colors": "^1.4.0"`

If a new tag is created on "Colors", Karma will eventually update "Colors" to newest (broken) version.

Shouldn't Karma fix the package on the latest working version (before the corrupted commits)?
`"colors": "1.4.0"`

Thanks.

falsandtru:
npm fixed colors and karma works well now.

reduckted:
Looks like this was fixed in #3763 and releasd in v6.3.17.

devoto13:
Closing per above.

PR #3739, opened 2022-01-09
fix(dependency): Use older version of colors package.

YoniSegal:
Use colors version 1.4.0. Malicious bug has been introduced to colors version 1.4.1

Fixes https://github.com/karma-runner/karma/issues/3738#issue-1097234601


YoniSegal:
Karma v4.4.1 currently points to colors ^1.4.0 as a dependency.
There is a bug in 1.4.1. This PR specifies 1.4.0 as the release to be used.

However, because I'm attempting to merge a change into a lower release, I'm not sure how this Pull Request should be made.

It says I'm attempting merge into karma-runner:master, but I want to fix a lower version than the master HEAD and then have you deploy it into a new v4.4 patch. How do I make such a change?

devoto13:
@YoniSegal We generally only support the latest version of Karma with fixes because of the limited maintenance time, so I don't think we'll be merging and releasing a 4.x branch. The fix will be made on the latest version, so I would suggest to update to it. If you can't updated for some reason, consider using [NPM overrides](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides) for [Yarn resolutions](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/) to set colors dependency to 1.4.0 in your project.

YoniSegal:
> @YoniSegal We generally only support the latest version of Karma with fixes because of the limited maintenance time, so I don't think we'll be merging and releasing a 4.x branch. The fix will be made on the latest version, so I would suggest to update to it. If you can't updated for some reason, consider using [NPM overrides](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides) for [Yarn resolutions](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/) to set colors dependency to 1.4.0 in your project.

I understand that as a general rule.
But I think this should be an exception as the fix is quite simple.
You just need to go to 4.4.1, pin the dependency and release a patch 4.4.2. Will that really cost so much maintenance time?

devoto13:
Well, I don't have permissions to release anything myself, so I would let @jginsburgn decide. 

But note that it's not that somebody manually publishes to NPM, it's a CI job which does it and getting a CI green on a pretty old branch may actually take quite some effort.

SerkanSipahi:
@devoto13 what about karma v.5.x.x and v6.x.x ?

devoto13:
@SerkanSipahi The malicious releases were taken off the NPM registry, so we'll not be releasing 5.x, but we'll pin `colors` package in the 6.x branch and likely replace it in 7.x line once it is started.

SerkanSipahi:
@devoto13 sounds good 👌

jginsburgn:
> Well, I don't have permissions to release anything myself, so I would let @jginsburgn decide.
> 
> But note that it's not that somebody manually publishes to NPM, it's a CI job which does it and getting a CI green on a pretty old branch may actually take quite some effort.

I agree with @devoto13 in that it could be significant to make CI/CD green in an older release. However, I can accept a PR that passes all CI tests to pin `colors` in older versions. Otherwise, we do not have the time bandwidth for that.

PR #3740, opened 2022-01-10
pin colors to 1.4.0

mannyluvstacos:
A Security Vuln was identified in the Colors package for >1.4.0, offending packages being `1.4.1`, `1.4.44-liberty`
- [source1](https://twitter.com/snyksec/status/1480286811482206216?ref_src=twsrc%5Egoogle%7Ctwcamp%5Eserp%7Ctwgr%5Etweet)
- [source2](https://twitter.com/snyksec/status/1480286811482206216?ref_src=twsrc%5Egoogle%7Ctwcamp%5Eserp%7Ctwgr%5Etweet)
- [source3](https://security.snyk.io/vuln/SNYK-JS-COLORS-2331906)

This PR pins the color package to `1.4.0` as advised on the [snyk page](https://snyk.io/blog/open-source-maintainer-pulls-the-plug-on-npm-packages-colors-and-faker-now-what/)

PR #3741, opened 2022-01-10
fix(deps): pin colors package to 1.4.0 due to security vulnerability

sergei-startsev:
Another one, hopefully contains all required changes :weary:

Pin `colors` package to `1.4.0` version due to security vulnerability. See details:
- [1.4.0...1.4.2 diff](https://diff.intrinsic.com/colors/1.4.0/1.4.2)
- https://github.com/Marak/colors.js/issues/285
- https://www.theverge.com/2022/1/9/22874949/developer-corrupts-open-source-libraries-projects-affected
- https://www.bleepingcomputer.com/news/security/dev-corrupts-npm-libs-colors-and-faker-breaking-thousands-of-apps/

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5211,7 +5211,7 @@
         "body-parser": "^1.19.0",
         "braces": "^3.0.2",
         "chokidar": "^3.5.1",
-        "colors": "^1.4.0",
+        "colors": "1.4.0",
         "connect": "^3.7.0",
         "di": "^0.0.1",
         "dom-serialize": "^2.2.1",
--- a/package.json
+++ b/package.json
@@ -420,11 +420,14 @@
     "chalkerx@gmail.com>",
     "weiran.zsd@outlook.com>"
   ],
+  "overrides": {
+    "colors": "1.4.0"
+  },
   "dependencies": {
     "body-parser": "^1.19.0",
     "braces": "^3.0.2",
     "chokidar": "^3.5.1",
-    "colors": "^1.4.0",
+    "colors": "1.4.0",
     "connect": "^3.7.0",
     "di": "^0.0.1",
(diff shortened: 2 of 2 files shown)


ellyxc:
is it possible if we set in config `color:false` don't load/require the color to avoid similar issue in the future?

devoto13 (approved):
Thanks for the PR!

sergei-startsev:
> is it possible if we set in config `color:false` don't load/require the color to avoid similar issue in the future?

I'd rather replace `colors` with another package (e.g. [chalk](https://www.npmjs.com/package/chalk)) in the long run.

simonl65:
Would be great if you would merge this - if only as a temporary workaround - as it's likely affecting MANY users worldwide ;-)

Happy for you to use an alternative, but delays cost lots of hair pulling.

SerkanSipahi:
Here is an interim solution https://github.com/nrwl/nx/issues/8450#issuecomment-1008759313 !

nicojs:
Update: NPM has pulled versions 1.4.2 and 1.4.1. Installing the `latest` colors version will give you `1.4.0`. This means that the problem is solved for karma users today, but this PR should still be merged IMO, for the karma users of tomorrow. (and later be replaced with a package that has a responsible maintainer)

jginsburgn:
@sergei-startsev please fix the CI tests :)

jginsburgn:
> @sergei-startsev please fix the CI tests :)

Sorry, I just realized that CI will turn green after https://github.com/karma-runner/karma/pull/3742 is merged. Let's just wait for that.

MannyPamintuanAtCare:
closes #3740

jginsburgn (approved):

simonl65:
Thank you

PR #3743, opened 2022-01-10
fix: replace `colors` with `ansi-colors`

### LINKED ISSUE #3738: pin colors lib to 1.4.0 as it been hacked and version 1.4.1 is broken

alan-agius4:
The author of the colors package purpose broke this package.

See: https://www.bleepingcomputer.com/news/security/dev-corrupts-npm-libs-colors-and-faker-breaking-thousands-of-apps/ and https://github.com/Marak/colors.js/issues/285

Closes #3738

### DIFF
--- a/lib/reporters/base_color.js
+++ b/lib/reporters/base_color.js
@@ -1,4 +1,4 @@
-const { red, yellow, green, cyan } = require('colors/safe')
+const { red, yellow, green, cyan } = require('ansi-colors')
 
 function BaseColorReporter () {
   this.USE_COLORS = true
--- a/package.json
+++ b/package.json
@@ -421,10 +421,10 @@
     "weiran.zsd@outlook.com>"
   ],
   "dependencies": {
+    "ansi-colors": "^4.1.1",
     "body-parser": "^1.19.0",
     "braces": "^3.0.2",
     "chokidar": "^3.5.1",
-    "colors": "^1.4.0",
     "connect": "^3.7.0",
     "di": "^0.0.1",
     "dom-serialize": "^2.2.1",


devoto13:
Unfortunately, this may be breaking to the plugins relying on the modified string prototype. I remember looking into switching to `colors/safe` some time ago and had to stop because of this. I don't remember any details though 😓 

IMO We should probably stick with pinning for now and replace the library as part of the major release.

alan-agius4:
Would be great if you could remember the problem that it caused.

I am good in replacing in package in a major if you think it’s breaking.

alan-agius4:
Great 🚀

djphan:
Just curious if this should be pinned to an explicit version instead of using the `^` or `~`. 

Just doing some dives into the colors issue and being pinned to 1.4.0 before the breaking changes would have added time. But don't want to delay any fixes

rumbcam:
The safest solution would be to just set colors to version 1.4.0 and drop the ^
"colors": "1.4.0",

ellyxc:
i guess few reporters like spec reporter also use colors

jginsburgn:
> 

https://github.com/karma-runner/karma/issues/3503

jginsburgn:
Due to the potential breakages I will mark this PR a WIP.

DABH:
This PR can be closed in favor of https://github.com/karma-runner/karma/pull/3763 .  `colors` has been migrated to `@colors/colors`, so we can do the simple renaming without making any further (and in particular, no breaking) changes.

DABH:
(As an update -- looks like #3763 was successfully merged, so I think this can be closed.)

jginsburgn:
> (As an update -- looks like #3763 was successfully merged, so I think this can be closed.)

Agreed!

PR #3750, opened 2022-01-23
fix(helper): make mkdirIfNotExists helper resilient to concurrent calls

devoto13:
The main motivation for this change is https://github.com/karma-runner/karma-coverage/issues/434#issuecomment-1017939333 where concurrent calls to the helper fail because of the race between the check and directory creation. This is a temporary solution and should be replaced with the native [`mkdir`](https://nodejs.org/api/fs.html#fsmkdirpath-options-callback) with the `recursive` option once minimum supported Node is bumped to 12.


jginsburgn:
@devoto13 do you consider this PR ready to be merged?

devoto13:
@jginsburgn Yes, it is ready. It is a companion PR for the https://github.com/karma-runner/karma-coverage/pull/463, so both should be merged to resolve the coverage generation race.

jginsburgn (approved):

PR #3756, opened 2022-02-01
fix: warn when `singleRun` and `autoWatch` are `false`

jginsburgn:
Setting `singleRun` and `autoWatch` to `false` will make the server remain idle
even when capturing a browser. It is a no-op combination. If it is detected,
fail and explain.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -96,6 +96,10 @@ class Server extends KarmaEventEmitter {
 
     this.log.debug('Final config', util.inspect(config, false, /** depth **/ null))
 
+    if (!config.autoWatch && !config.singleRun) {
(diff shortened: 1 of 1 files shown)


devoto13:
Hmm, isn't it possible to trigger tests using `karma run` command in this case? I think the Karma plugin in IntelliJ IDEA/WebStrom uses this combination to keep the server ready, but only run test when user hits the Run button.

devoto13 (dismissed):
Makes sense! Let's see if anybody complains about noisy logs :)

jginsburgn:
I rebased and I need your approval again @devoto13

devoto13 (approved):

PR #3759, opened 2022-02-09
fix(security): mitigate the "Open Redirect Vulnerability"

jginsburgn:


client/karma.js
```diff
@@ -239,7 +239,16 @@ function Karma (updater, socket, iframe, opener, navigator, location, document)
       self.updater.updateTestStatus('complete')
     }
     if (returnUrl) {
-      if (!/^https?:\/\//.test(returnUrl)) {
+      // TODO(https://github.com/karma-runner/karma/issues/3503): replace the for-loop below with the `Array.prototype.includes()` method, when dropping support for IE.
+      var isReturnUrlInAllowlist = false
+      for (var i = 0; i < this.config.allowedReturnUrls.length; i++) {
+        var allowedReturnUrl = this.config.allowedReturnUrls[i]
+        if (allowedReturnUrl === returnUrl) {
+          isReturnUrlInAllowlist = true
```
DrJKL, line 247: Want to break?

docs/config/01-configuration-file.md
```diff
@@ -277,6 +277,15 @@ upon the completion of running the tests. Setting this to false is useful when e
...
 If true, Karma does not display the banner and browser list. Useful when using karma on component tests with screenshots.
 
+## client.allowedReturnUrls
+**Type:** Array
+
+**Default:** `[]`
+
+**Description:** Define the allowed values for the `return_url` query parameter.
+
+If the value of the `return_url` query parameter is not in this list, navigation to it will be blocked. This mitigates the "Open Redirect Vulnerability".
```
DrJKL, line 287: I don't see justification in the description for other client options.

DrJKL (dismissed):

client/karma.js
jginsburgn: Done.

docs/config/01-configuration-file.md
jginsburgn: What do you mean?

test/client/karma.spec.js
```diff
@@ -462,6 +465,26 @@ describe('Karma', function () {
       clock.tick(10)
     })
 
+    it.only('should not navigate the client to return_url if not benign', function () {
```
devoto13, line 468: nit: remove `.only`

docs/config/01-configuration-file.md
```diff
@@ -277,6 +277,15 @@ upon the completion of running the tests. Setting this to false is useful when e
 
 If true, Karma does not display the banner and browser list. Useful when using karma on component tests with screenshots.
 
+## client.allowedReturnUrls
```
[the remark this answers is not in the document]
jginsburgn: > Hmm, wouldn't the option in its current state disallow any existing usage of the `return_url` unless the user updates the config? Thus resulting in a breaking change.

I had that concern myself. I thought it might not be under the breaking change (major update) category, because the API is not changing *per se*. Users should only need to change their configuration file. However, I agree that a change even in the configuration is a breaking change.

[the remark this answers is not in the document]
jginsburgn: Follow up comments to your other points are coming soon :) I need to be afk.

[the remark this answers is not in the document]
jginsburgn: > IMO it makes sense to only allow http/https URLs by default and if the user wants extra security, they can modify the property. So maybe set default as ['^https?://'] and allow users to override it if they want to disallow all URLs or only allow some specific ones. Then we can swap the default to [] in the next major if at all. My thinking is that Open Redirect is not very applicable for Karma anyways - users would not normally click on Karma links.

>Or maybe we should even take a bigger action and get rid of the redirect_url feature altogether. It was added in https://github.com/karma-runner/karma/commit/6af2c897f3b35060a146efdef7da597ba53d8cdd 9 years ago and there is no issue outlining the motivation for this feature and no other mentions of it, so I'm not sure if anybody is using it at all. So maybe removing the feature would be the best mitigation here 😅

I will fix this CL to allow `https?://` by default. That way it will not be a breaking change. And in the next major release we can definitely get rid of `return_url`; I agree with your line of thinking.

test/client/karma.spec.js
jginsburgn: You're right in that I need to remove `.only` but wrong in that it is a nit. I would say it is mandatory :) 👍

docs/config/01-configuration-file.md
jginsburgn: Done.

test/client/karma.spec.js
jginsburgn: Done.

devoto13 (approved):
Nice!

docs/config/01-configuration-file.md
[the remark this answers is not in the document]
jginsburgn: Done.

PR #3761, opened 2022-02-11
docs(configuration-file): document async function

nicojs:
See https://github.com/karma-runner/karma/pull/3660

### DIFF
--- a/docs/config/01-configuration-file.md
+++ b/docs/config/01-configuration-file.md
@@ -53,6 +53,20 @@ module.exports = (config) => {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


devoto13:
There is already a [PR](https://github.com/karma-runner/karma/pull/3733/files) in flight to document async config. I'm not sure whether it is best to present an example with old-school promise, async function, or both. I'm leaning towards the async function example as more modern and readable. So I think we can rebase it once the other PR is merged as it provides more info about the feature.

jginsburgn (approved):

PR #3779, opened 2022-03-23
Ensure testing on BrowserStack when in CI

jginsburgn:


devoto13:
Did you allow access to GitHub secrets from forks somehow? Because otherwise, any PRs from forks will fail because they don't have access to the secrets.

jginsburgn:
> Did you allow access to GitHub secrets from forks somehow? Because otherwise, any PRs from forks will fail because they don't have access to the secrets.

You are absolutely right! I think we should, however, should have a mechanism for PRs from forks to test in SauceLabs and BrowserStack. I am going to move this discussion to a bug!

PR #3780, opened 2022-03-27
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated. While [String.prototype.substring()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) isn't deprecated I replaced it with `slice()` as well.  `slice()` is generally a bit faster (and uses less bites as the name is shorter) and this way we don't have 2 functions which nearly do the same thing in the code.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/client/main.js
+++ b/client/main.js
@@ -15,7 +15,7 @@ var socket = io(location.host, {
   reconnectionDelay: 500,
   reconnectionDelayMax: Infinity,
   timeout: BROWSER_SOCKET_TIMEOUT,
-  path: KARMA_PROXY_PATH + KARMA_URL_ROOT.substr(1) + 'socket.io',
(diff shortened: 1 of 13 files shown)


devoto13 (approved):
LGTM, thanks!

jginsburgn (approved):

PR #3789, opened 2022-04-19
fix(client): error out when opening a new tab fails

jginsburgn:
Safari in iOS fails opening a new tab with the `window.open` API. This makes it explicit that that is the culprit, in the right case. Also, it suggests a workaround: enabling pop ups.


devoto13 (approved):
Makes sense!

PR #3797, opened 2022-05-16
Support subresource integrity verification

falsandtru:
https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

### DIFF
--- a/docs/config/02-files.md
+++ b/docs/config/02-files.md
@@ -54,6 +54,11 @@ The `files` array determines which files are included in the browser, watched, a
 * **Default.** `false`
 * **Description.** Should the files be served from disk on each request by Karma's webserver?
 
+### `integrity`
+* **Type.** String
+* **Default.** `undefined`
+* **Description.** Set the `integrity` HTML attribute value to the `<script>` or the `<link>` tag to load the resource that matches the given pattern if the pattern is an absolute URL.
+
 ## Pattern matching and `basePath`
 - All of the relative patterns will get resolved using the `basePath` first.
 - If the `basePath` is a relative path, it gets resolved to the
--- a/lib/config.js
+++ b/lib/config.js
@@ -32,7 +32,7 @@ try {
 } catch {}
 
 class Pattern {
-  constructor (pattern, served, included, watched, nocache, type, isBinary) {
+  constructor (pattern, served, included, watched, nocache, type, isBinary, integrity) {
     this.pattern = pattern
     this.served = helper.isDefined(served) ? served : true
     this.included = helper.isDefined(included) ? included : true
@@ -41,6 +41,7 @@ class Pattern {
     this.weight = helper.mmPatternWeight(pattern)
     this.type = type
     this.isBinary = isBinary
+    this.integrity = integrity
   }
 
   compare (other) {
@@ -49,8 +50,8 @@ class Pattern {
 }
 
 class UrlPattern extends Pattern {
-  constructor (url, type) {
-    super(url, false, true, false, false, type)
+  constructor (url, type, integrity) {
+    super(url, false, true, false, false, type, undefined, integrity)
   }
 }
 
--- a/lib/file-list.js
+++ b/lib/file-list.js
@@ -62,9 +62,9 @@ class FileList {
 
     let lastCompletedRefresh = this._refreshing
     lastCompletedRefresh = Promise.all(
-      this._patterns.map(async ({ pattern, type, nocache, isBinary }) => {
+      this._patterns.map(async ({ pattern, type, nocache, isBinary, integrity }) => {
         if (helper.isUrlAbsolute(pattern)) {
-          this.buckets.set(pattern, [new Url(pattern, type)])
+          this.buckets.set(pattern, [new Url(pattern, type, integrity)])
           return
         }
 
--- a/lib/middleware/karma.js
+++ b/lib/middleware/karma.js
@@ -182,16 +182,16 @@ function createKarmaMiddleware (
               }
             }
 
+            const integrityAttribute = file.integrity ? ` integrity="${file.integrity}"` : ''
+            const crossOriginAttribute = includeCrossOriginAttribute ? ' crossorigin="anonymous"' : ''
             if (fileType === 'css') {
-              scriptTags.push(`<link type="text/css" href="${filePath}" rel="stylesheet">`)
+              scriptTags.push(`<link type="text/css" href="${filePath}" rel="stylesheet"${integrityAttribute}${crossOriginAttribute}>`)
             } else if (fileType === 'dom') {
               scriptTags.push(file.content)
             } else if (fileType === 'html') {
-              scriptTags.push(`<link href="${filePath}" rel="import">`)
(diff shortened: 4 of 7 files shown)


falsandtru:
@jginsburgn @devoto13 Can you approve the CI tests?

falsandtru:
@jginsburgn @devoto13 Without integrity checking, your local machines and CI servers are at risk of being compromised by falsification of remote resources. Can you review it and support integrity checking?

falsandtru:
OK

lib/middleware/karma.js
```diff
@@ -182,19 +182,20 @@ function createKarmaMiddleware (
               }
             }
 
+            const integrityAttribute = file.integrity ? ` integrity="${file.integrity}"` : ''
+            const crossOriginAttribute = includeCrossOriginAttribute ? ' crossorigin="anonymous"' : ''
             if (fileType === 'css') {
-              scriptTags.push(`<link type="text/css" href="${filePath}" rel="stylesheet">`)
+              scriptTags.push(`<link type="text/css" href="${filePath}" rel="stylesheet"${integrityAttribute}${crossOriginAttribute}>`)
```
devoto13, line 188: I wonder if adding `crossorigin` attribute here may break tests if somebody is loading a stylesheet from the CDN or something because of failed CORS check 🤔

It probably shouldn't because the it will send a [simple request](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests) where CORS check is not ran, but would be good to know for sure. @falsandtru Do you know? Or can you check it otherwise?

lib/config.js
```diff
@@ -32,7 +32,7 @@ try {
 } catch {}
 
 class Pattern {
-  constructor (pattern, served, included, watched, nocache, type, isBinary) {
+  constructor (pattern, served, included, watched, nocache, type, isBinary, integrity) {
```
devoto13, line 35: Can you also document this new property [here](https://github.com/karma-runner/karma/blob/master/docs/config/02-files.md)?

lib/config.js
```diff
@@ -49,8 +50,8 @@ class Pattern {
 }
 
 class UrlPattern extends Pattern {
-  constructor (url, type) {
-    super(url, false, true, false, false, type)
+  constructor (url, type, integrity) {
+    super(url, false, true, false, false, type, false, integrity)
```
devoto13, line 54: nit: Should it be `undefined` instead of `false` for the `isBinary`?

devoto13 (commented):
Two small comments and one question. 
Otherwise LGTM

lib/config.js
falsandtru: I set false instead of undefined because all the other parameters are passed boolean as the default value. Should only the isBinary be passed undefined?

lib/config.js
falsandtru: I'll try it.

lib/middleware/karma.js
falsandtru: > I wonder if adding crossorigin attribute here may break tests if somebody is loading a stylesheet from the CDN or something because of failed CORS check 

I feel such fallback requests should also be checked by CORS. Otherwise that check should be disabled by configuration.

> It probably shouldn't because it will send a [simple request](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests) where CORS check is not ran, but would be good to know for sure.

I'm also not sure but this should be an appropriate change. So I intend to revert this CORS constraint of link tags if any reasonable use case to disable the constraint without using the configuration is found.

devoto13: Okay, sounds like a reasonable plan to me! I guess the risk of breaking somebody is extremely low here.

lib/config.js
devoto13: In practice is is [never used](https://github.com/karma-runner/karma/blob/5e71cf591e9006ad2621dd88d09739bfbffd463c/lib/file-list.js#L66) for the absolute URL patterns, but IMO it is better to set it to `undefined` because it used to be `undefined` before this PR and also to avoid confusion for the future reader.

Having said that, I don't think it should block this PR.

devoto13 (approved):
LGTM
Thanks for the PR!

lib/config.js
falsandtru: OK, I'll fix it.

falsandtru: done.

devoto13 (dismissed):
LGTM

falsandtru:
Fixed the too long commit headers.

falsandtru:
@jginsburgn Can you review?

jginsburgn:
I am sorry I had not been able to review this PR! I am unfamiliar with Integrity Checking. Let me get familiarized and review this merge request promptly.

lib/file-list.js
```diff
@@ -62,9 +62,9 @@ class FileList {
 
     let lastCompletedRefresh = this._refreshing
     lastCompletedRefresh = Promise.all(
-      this._patterns.map(async ({ pattern, type, nocache, isBinary }) => {
+      this._patterns.map(async ({ pattern, type, nocache, isBinary, integrity }) => {
```
jginsburgn, line 65: What will the behavior be if a pattern matches multiple files? Will all the matched files have the same hash? If so, that makes no sense, right?

jginsburgn: Maybe we could validate that when integrity is used, a pattern cannot match more than a single file or validate that the value passed to pattern has no wildcards, which would yield the same effect.

docs/config/02-files.md
```diff
@@ -54,6 +54,11 @@ The `files` array determines which files are included in the browser, watched, a
 * **Default.** `false`
 * **Description.** Should the files be served from disk on each request by Karma's webserver?
 
+### `integrity`
+* **Type.** String
+* **Default.** `undefined`
+* **Description.** Set the `integrity` HTML attribute value to the `<script>` or the `<link>` tag to load the resource that matches the given pattern if the pattern is an absolute URL.
```
jginsburgn, line 60: > if the pattern is an absolute URL

QQ: Where is this assertion in the PR?

lib/file-list.js
falsandtru: Integrity is verified only if a pattern is an absolute URL.

docs/config/02-files.md
falsandtru: Here: https://github.com/karma-runner/karma/pull/3797/files#diff-bbc76ee4392cb9602e2ce4d6081c2985bad45532441cec4da57b4a2d8a57b2b8R66

falsandtru: @jginsburgn OK?

falsandtru: @jginsburgn Can you approve?

lib/file-list.js
jginsburgn: You mean in the browser?

falsandtru: That code doesn't seem to depend on environments.

jginsburgn: Right. Sorry!

jginsburgn (approved):

jginsburgn:
I'm sorry I took so long.

PR #3816, opened 2022-09-13
fix: pass integrity value

falsandtru:
SRI was not enabled.


falsandtru:
@jginsburgn @devoto13 Can you review?

jginsburgn (approved):

jginsburgn:
> @jginsburgn @devoto13 Can you review?

Sorry it took so long, and btw I am no longer an active owner of the project. I may chime in here and then to contribute but it is better to rely on @karma-runner/google-web-test-team for active maintenance and guidance!

PR #3838, opened 2023-03-17
docs: alternative for local `karma-cli` install

kevireilly:
While installing `karma-cli` globally may commonly make some sense, another alternative is keeping it encapsulated within each individual repository which is also easy enough.


anthony-redFox:
Hi, the project `karma-cli` is closed, I don't think that need to push people to use old project.

PR #3846, opened 2023-04-27
docs: Add deprecation notice to Karma README

juliemr:


keithamus:
Kudos to the Karma team for this decision! It shows a great deal of maturity and respect for the community to step aside when the time is right.

PR #3849, opened 2023-05-16
chore: use last version of ua-parser-js and drop http2

anthony-redFox:


dotnetCarpenter:
LGTM! Unfortunately, I'm not a maintainer

PR #3851, opened 2023-06-05
new stopper

asiermartinez:


lib/stopper-2.js
```diff
@@ -0,0 +1,69 @@
+const http = require('http')
+const cfg = require('./config')
+const logger = require('./logger')
+const helper = require('./helper')
+const { lookup } = require('./utils/dns-utils')
+
```
asiermartinez, line 6:
```suggestion
let serverListening = false

```

lib/stopper-2.js
```diff
@@ -0,0 +1,69 @@
...
+    } else {
+      log.error(`Server returned status code: ${response.statusCode}`)
+      done(1)
+    }
+  })
+
+  request.on('error', function (e) {
+    if (e.code === 'ECONNREFUSED') {
+      log.error(`There is no server listening on port ${config.port}`)
+      done(1, e.code)
```
asiermartinez, line 63:
```suggestion
    if (!serverListening) {
```

asiermartinez: And this other change https://github.com/karma-runner/karma/pull/3851/files#r1217940249

